PR #13, opened 2015-01-20
Fixes to setup.py

tbekolay:
- Added `ez_setup.py`, which is necessary if the user doesn't have setuptools installed yet
- Using a string for version
  - To support this, made a `version` module in nengo_viz, which `setup.py` imports (as is done in Nengo)
- Correct `setup` arg is `install_requires` rather than `requires`

### DIFF
--- a/ez_setup.py
+++ b/ez_setup.py
@@ -1,0 +1,332 @@
+#!/usr/bin/env python
+"""Bootstrap setuptools installation
+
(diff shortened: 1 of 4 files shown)


tcstewar:
Thank you!  I had no idea and had just been copying setup,py blindly from old versions...  :)

tbekolay:
No problemo :rainbow:

PR #25, opened 2015-01-28
Improved slider

### LINKED ISSUE #5: Make sliders look pretty
- Add a line showing where they can move
- Show the numerical value at the top
- Shortcut for zeroing sliders
- Move sliders automatically in accordance with their default value
- Jump time forward if you move the sliders (you can't change values in the past)
  - in Nengo 1.4 this was implemented by having the sliders be deactivated when you're looking at past data, but I think this time just jumping forward in time to the present would be clearer.

### LINKED ISSUE #12: Dragging sliders need max/min bounds
When dragging past the bounds of a slider, the cursor position and slider get out of synch.

### LINKED ISSUE #23: Moving elements to top
Currently moving elements to top removes and re-adds the element making any other click events on the element unusable.

cchan987:
viz.js 41-50 (Using zindex to move elements to front, allows click events on other elements) fixes #23

viz_sliders.js
30-33 (add a displayed value to the sliders) 
43-53 (sliders can only be used on mouseover) Fixes #12 
57-86 (sliders jump to middle/zero when middle clicked) Partial Fix #5 
92-131 (sliders can only be used on mouseover) Fixes #12

### DIFF
--- a/nengo_viz/static/viz_slider.js
+++ b/nengo_viz/static/viz_slider.js
@@ -139,7 +139,7 @@ VIZ.Slider = function(args) {
         //Good for positioning regardless of # of sliders
         var guide_x = args.width / (2 * args.n_sliders) + 
             (args.width / args.n_sliders) * i - this.guideline_width / 2;
-        VIZ.set_transform(guideline,guide_x,0);
+        VIZ.set_transform(guideline, guide_x, 0);
         this.div.appendChild(guideline);
         }
 


cchan987:
I grouped these changes together as they were all connected, let me know if it would be better to split them into different pull requests/ branches

tcstewar:
Hmmm...  a few quick bug reports....

 - the z-index approach seems to mostly work, but I've seen a few situations where the slider graph isn't jumping to the front when I drag it around.  I haven't been able to quite reliably replicate the problem, though
 - Limiting slider movement on mouseover seems to have two undesirable consequences:
    - if you move quickly to the extreme limits of the slider, it ignores you
    - touch control doesn't work any more
 - Middle-click to reset-to-zero seems to only work about 10% of the time for me.

(for testing purposes, I was using Chrome under Windows)

cchan987:
hmm, for the middle clicking, it should only works when clicking the darker grey slider bar.

cchan987:
Added functionality for touch controls (needs testing on touchscreen which I don't have, @tcstewar may be able to help out XD)
and made the sliders more friendly for rapid dragging.

cchan987:
zindex for putting elements on top updated, simpler, crisper, better.

cchan987:
viz.css - changed background from white to transparent
viz.js - rewrote move element to top function using z-index

viz_slider.css - added slider guideline styling

viz_slider.js 
- Added the value to be displayed on the slider
- added zindex for the slider
-added , on slider drag end, if slider beyond limits, return drag-y value to boundary value
-added guidelines to the sliders
-modified resize handler to accommodate guideline

Point of concern:
-Guideline for sliders is a very long vertical line (999999) drawn in d3, when slider is resized, the svg holding the line is resized to show more/less of the line. I was struggling with redrawing the guideline without throwing the layout of the other elements off and this was a quick fix that presented itself.(thoughts?)

cchan987:
guideline issue resolved

cchan987:
Also note: the 2d.py example is in this branch, should it be removed?

tcstewar:
Looking pretty good...  :)

If you think the 2d.py example would be useful for future testing, then sure, leave it in.

The main thing I think still needs to get cleaned up here is that I'd like to keep with style of having spaces on either side of operators.  So ```i * width / N``` rather than ```i*width/N```.  Personally, I'm fine either way, but we should be consistent and all the existing code is with spaces.  :)

cchan987:
Changes made as per discussion!

nengo_viz/static/viz_slider.js
```diff
@@ -55,16 +97,56 @@ VIZ.Slider = function(args) {
...
+        var guideline = document.createElement('div');
+        this.sliders[i].guideline = guideline;
+        guideline.classList.add('guideline');
+        guideline.style.position = "fixed";
+        guideline.style.height = args.height;
+        guideline.style.width = this.guideline_width;
+        //Good for positioning regardless of # of sliders
+        var guide_x = args.width / (2 * args.n_sliders) + 
+            (args.width / args.n_sliders) * i - this.guideline_width / 2;
+        VIZ.set_transform(guideline,guide_x,0);
```
tcstewar: Oops, forgot about this guideline:  spaces after commas

cchan987:
whoops! fixed

PR #28, opened 2015-01-29
Switched plots to white background, border on hover

tcstewar:
This is a closer match to the original visualizer, and makes for a cleaner image

### DIFF
--- a/nengo_viz/static/viz.css
+++ b/nengo_viz/static/viz.css
@@ -1,3 +1,8 @@
 .graph {
-    background: #eeeeee;
+    background: white;
+    border: 1px solid white;
+}
+
+.graph:hover {
+    border: 1px solid #888888;
 }


nengo_viz/static/viz.css
```diff
@@ -1,3 +1,8 @@
 .graph {
-    background: #eeeeee;
-}
\ No newline at end of file
+    background: white;
+    border: 1px solid white;
```
cchan987: I'm thinking that if the backgrounds are changed to white, the border should be visible(non-white, perhaps a grey shade), this would give a better sense of where the components are.

tcstewar: That's what the :hover is for.  You only need to know where the components are when you're manipulating them.  The intent is to reduce visual clutter.  (i.e. the visual clutter from seeing the outlines of every component all the time.  notice that this is true whether we have a grey background or a border -- they both introduce visual clutter)

cchan987:
hmm, I see what your getting at. For the sliders it may be useful to know their relative positions to the top/bottom even when they are not being manipulated though. (eg. the user can see the slider is at its max value already without having to hover over it)

tcstewar:
> hmm, I see what your getting at. For the sliders it may be useful to know their relative positions to the top/bottom even when they are not being manipulated though. (eg. the user can see the slider is at its max value already without having to hover over it)

Yes, which is why one of the issues listed for sliders is to put a guide line underneath them so you can see the range it can go to (which is what sliders in every other user interface use, so we should probably follow that standard.

cchan987:
I'll add the guideline in as part of the improved_slider branch

PR #37, opened 2015-02-04
Iss4

cchan987:
Added Title place-holder (using component id)
Added x-axis labels at fixed positions
Used same fonts as plot ticks
modified min-width and min-height for components to be slightly bigger (plots turn to a mess if they are smaller)
Added suppression of x-axis label when graph is shrunk to small size
Added a call to resize on plot creation (In case label should be suppressed)

### DIFF
--- a/nengo_viz/static/viz_value.js
+++ b/nengo_viz/static/viz_value.js
@@ -24,12 +24,6 @@ VIZ.Value = function(args) {
         .attr('width', '100%')
         .attr('height', '100%');
 
-    //Add the title 
-    this.svg.append("text")
-        .text(args.id)
-        .attr('class', 'graph_text')
-        .attr('y', 15);
-
     /** scales for mapping x and y values to pixels */
     this.scale_x = d3.scale.linear();
     this.scale_y = d3.scale.linear();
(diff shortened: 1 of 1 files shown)


nengo_viz/static/viz_value.js
```diff
@@ -22,28 +23,50 @@ VIZ.Value = function(args) {
     this.svg = d3.select(this.div).append('svg')
         .attr('width', '100%')
         .attr('height', '100%');
-        
+
+    //Add the title 
+    this.svg.append("text")
+        .text(args.id)
```
tcstewar: This should actually be the ensemble's ```.label``` attribute.

nengo_viz/static/viz_value.js
```diff
@@ -75,6 +98,8 @@ VIZ.Value = function(args) {
     this.path.enter().append('path')
              .attr('class', 'line')
              .style('stroke', function(d, i) {return colors[i];});
+
+    this.on_resize(100, 100);
```
tcstewar: Why is this being called with a fixed 100, 100?  Shouldn't it be ```on_resize(args.width, args.height)```?

nengo_viz/static/viz_value.js
tcstewar: (note: to get at this value, it'll have to be passed in from the server.  I think the easiest way is to do that on the ```Value.javascript()``` call.)

tcstewar: (and the same thing should happen for all Components, for that matter, so it might be best if this was handled in ```viz.js``` on the client side (like id and width and height and x and y are right now)

cchan987:
Updated
-soft coded initial resize call
-removed title, to be added in viz.js
-Left a space for title with margin_top unchanged

PR #39, opened 2015-02-04
Iss38

cchan987:


nengo_viz/static/viz_slider.js
```diff
@@ -151,6 +128,32 @@ VIZ.Slider = function(args) {
...
+                                    // Set sliders attributed position to the middle
+                                    target.setAttribute('drag-y', midpoint);
+
+                                    //Move the slider to the middle, subtract half slider height due to pixel offset
+                                    VIZ.set_transform(target, x_pos, midpoint - height / 2);
+
+                                    //Send update to the server
+                                    self.ws.send(slider_index + ',' + new_value);
+                                };
+
```
tcstewar: Indentation!  :)

nengo_viz/static/viz_slider.js
```diff
@@ -151,6 +128,32 @@ VIZ.Slider = function(args) {
 VIZ.Slider.prototype = Object.create(VIZ.Component.prototype);
 VIZ.Slider.prototype.constructor = VIZ.Slider;
 
+VIZ.Slider.prototype.zero_out = function(graph,slider) {
```
tcstewar: Spacing after comma

tcstewar:
Instead of this being a ```zero_out``` function, how about making it a ```set_value``` function?  That seems a lot more generally useful to me....

tcstewar:
(For example, it could be used to actually automatically move the slider based on the input value, as in the Java visualizer)

cchan987:
Good idea, I'll make it so that there is a param input_value that can be from [-1,1] which will move the slider to that point automatically.

tcstewar:
> Good idea, I'll make it so that there is a param input_value that can be from [-1,1] which will move the slider to that point automatically.

I wouldn't put a limit on that range -- I'd assume we'll want to be able to change that range in the future (like in that java visualizer)

cchan987:
updated

nengo_viz/static/viz_slider.js
```diff
@@ -151,6 +127,41 @@ VIZ.Slider = function(args) {
 VIZ.Slider.prototype = Object.create(VIZ.Component.prototype);
 VIZ.Slider.prototype.constructor = VIZ.Slider;
 
+VIZ.Slider.prototype.set_value = function(graph, slider, min_in, max_in, value) {
+    /** check if click was the middle mouse button */
```
tcstewar: I'm not sure about min_in and max_in.  Those really seem like things the slider itself should know about, not things that have to be passed in to a call to ```set_value```.  Instead, I'd recommend adding ```max_value``` and ```min_value``` to the slider object.

I'd do this up around lines 26-28:
```
        slider.index = i;
        slider.div = document.createElement('div');
        slider.value = 0;
        slider.min_value = -1;    // add these lines
        slider.max_value = 1;    // add these lines
```

tcstewar: Actually, wait a second, no, ignore all of that.  You already have a min_value and max_value as part of the ```slider.scale``` object.  Just use that, and all this extra code goes away!

nengo_viz/static/viz_slider.js
```diff
@@ -151,6 +127,41 @@ VIZ.Slider = function(args) {
...
+
+    //Math for mapping one range to another
+    var left_min = min_in;
+    var left_max = max_in;
+    var left_span = left_max - left_min;
+    var right_min = self.scale.range()[1];
+    var right_max = self.scale.range()[0];
+    var right_span = right_max - right_min;
+    var scale = (value - left_min) / left_span;
+    var point = right_min + (scale * right_span);
```
tcstewar: Use the slider.scale object to do this mapping, rather than writing it all yourself.  :)

cchan987:
updated
-Got rid of unneeded math

nengo_viz/static/viz_slider.js
```diff
@@ -151,6 +127,34 @@ VIZ.Slider = function(args) {
 VIZ.Slider.prototype = Object.create(VIZ.Component.prototype);
 VIZ.Slider.prototype.constructor = VIZ.Slider;
 
+VIZ.Slider.prototype.set_value = function(graph, slider, value) {
```
tcstewar: Why is ```graph``` being passed in?  That's just ```this```, after all.  It should be

```
VIZ.Slider.prototype.set_value = function(slider, value) {
```

and called with

```
                   self.set_value(this, 0);
```

nengo_viz/static/viz_slider.js
```diff
@@ -151,6 +127,34 @@ VIZ.Slider = function(args) {
...
+    /** check if click was the middle mouse button */
+    var target = slider;
+    var self = graph;
+    var slider_index = target.slider.index; // Get index (For 1D > sliders)
```
tcstewar: Hmm, this feels weird to me.  I'd pass in the index and get the slider with ```this.sliders[index]```.  It just seems odd to me to pass in the object and figure out its index, rather than passing in the index and figuring out the object.

cchan987:
Ah right, it is probably more efficient to pass an index int rather than an object. I'll fix that up

cchan987:
updated!

nengo_viz/static/viz_slider.js
```diff
@@ -151,6 +128,35 @@ VIZ.Slider = function(args) {
...
+    var target = this.sliders[slider_index].div;
+
+    //important for 2d sliders
+    var x_pos = target.getAttribute('fixed-x'); 
+    
+    //Get the scaled value
+    var point = this.scale(value);
+
+    //Get the slider height
+    var height = parseInt(target.style.height);
```
tcstewar: Hmm, I hadn't realized we aren't storing ```args.height``` anywhere, so we have to get it through something ugly like this.  I'm pretty sure we should store ```args.height``` and ```args.width``` so that we can just access them via ```this.width``` and ```this.height```.  But that can be a separate PR.

nengo_viz/static/viz_slider.js
```diff
@@ -151,6 +128,35 @@ VIZ.Slider = function(args) {
...
+
+    //Change shown text value to 0
```
tcstewar: Keep comments in sync with the code!

nengo_viz/static/viz_slider.js
```diff
@@ -151,6 +128,35 @@ VIZ.Slider = function(args) {
...
+    target.firstChild.textContent = value;
+
+    //Change sliders value to 0
+    target.slider.value = value;
```
tcstewar: Same thing here with the comment....  :)

cchan987:
-comments updated
-#40 opened for slider height variable

cchan987:
The height variable was already there, I was taking the long way around before.

Fixed #40 in latest commit.

PR #45, opened 2015-02-05
Include bootstrap, jquery, and restyled button

amaziying:


nengo_viz/static/viz_sim_control.js
```diff
@@ -37,15 +37,19 @@ VIZ.SimControl = function(div, args) {
     
     /** Create the pause button */
     this.pause_button = document.createElement('button');
-    this.pause_button.innerHTML='Pause';
+    this.pause_button.className = "btn btn-default play-pause-button";
+    this.pause_button.innerHTML='<i class="glyphicon glyphicon-pause"></i>';
```
tcstewar: Why is this an ```<i>``` tag?

nengo_viz/templates/page.html
```diff
@@ -10,7 +10,10 @@
     <script src="static/viz_slider.js"></script>
     <link href="static/viz_slider.css" rel="stylesheet" type="text/css" />
     <script src="static/viz_sim_control.js"></script>
-    <link href="static/viz_sim_control.css" rel="stylesheet" type="text/css" />    
+    <link href="static/viz_sim_control.css" rel="stylesheet" type="text/css" /> 
+    <script src="static/lib/js/jquery-2.1.3.min.js"></script>
+    <script src="static/lib/js/bootstrap.min.js"></script>
+    <link href="static//lib/css/bootstrap.min.css" rel="stylesheet" type="text/css">
```
tcstewar: We should have some consistency in the locations of these files.  If you think there should be this separate ```lib``` directory, then interact-1.2.2.js and d3.v3.min.js should go in there too.  

(note: to move files correctly with git, do something like ```git mv interact-1.2.2.js lib/js/interact-1.2.2.js```)

tcstewar:
I like the overall look of bootstrap.  :)  Thanks for bringing it in!

When I try this in Chrome under Windows, if I click on the play/pause button then all of a sudden my window gains a scroll bar on the right.  I haven't tried it on other situations yet.

Also, the play/pause button should be larger, and on the right side (as in the previous java visualizer).  I like the look of the speed and time display.  :)

cchan987:
After pulling master to this branch the slider value jumps to the top left, originally in the middle left.

Is this desired behaviour?
screenshot-2

nengo_viz/static/viz_sim_control.js
amaziying: The <i> tag is kind of an industry standard for using icons (used by facebook and twitter). The alternative would be to use <span> tags. I don't really have a preference of which way to do it, but here's a good comparison can be found here: http://stackoverflow.com/questions/11135261/should-i-use-i-tag-for-icons-instead-of-span

nengo_viz/templates/page.html
amaziying: Right, so I will move the D3 and interact library.
I'm also thinking of organizing the nengo_viz JS and CSS files as well, because of the organization issues when we add more and more files.

amaziying:
Terry, I really like bootstrap too! No problem :)
I can't reproduce that issue on my version of chrome + my window size
and will do! I'll create an issue for moving the button and metrics to the right of the timeline + make button bigger.

Chris,
I did not intend on that behaviour, but with the bootstrap library comes with some default styles. You would need to adjust your css a bit if you want to move it back to the middle.

nengo_viz/static/viz_sim_control.js
tcstewar: Weird...  although according to that page, Bootstrap 3 recommends using ```<span>``` instead of ```<i>```, which feels better to me....

amaziying: Sure, let's stick to using span tags for icons then :)
committed the change.

cchan987:
I'm still having the screen size issue when pressing pause/play unfortunately

cchan987:
I think the metrics container is responsible for the sidebar appearing

cchan987:
changing the margin of the metrics container to 20 in the sim_control.css solved the problem for me

tcstewar:
I think we're pretty much agreed on using bootstrap, so I'm going to merge this in if there are no objections.  I've made the CSS change suggested by @cchan987 (although, in the future, it's usually up to the person who makes the pull request to do those small changes; otherwise we end up with multiple people making the same changes and it gets all confusing).

PR #51, opened 2015-02-10
Zero out

cchan987:
A green zero-out button for enabling zeroing on touch interfaces.

Still to be decided if this is the option we want to go with or if the config menus (to be implemented) will suffice.


tcstewar:
Hmm...  interesting.  I still want to compare to how this feels to just having it on the config menu.

My worries are:
 - this isn't particularly discoverable behaviour
 - the target area is very small (so it's hard to hit with a finger, or a mouse for that matter)

That said, it's less visually distracting than I thought it would be like this.

cchan987:
the config_menu branch now has a rough mockup of dropdown set_value functionality

PR #54, opened 2015-02-12
SimControl times

tcstewar:
Allows configuring the amount of time kept and the amount of time shown by the simulator control


tcstewar:
(Note: this should be merged *after* spike_raster)

PR #56, opened 2015-02-12
Added options for range on sliders

tcstewar:
Specifying the minimum and maximum value on a Slider

### DIFF
--- a/nengo_viz/components/slider.py
+++ b/nengo_viz/components/slider.py
@@ -3,13 +3,18 @@ import numpy as np
 from nengo_viz.components.component import Component
 
 class Slider(Component):
-    def __init__(self, viz, node, **kwargs):
+    def __init__(self, viz, node, min_value=-1, max_value=1, **kwargs):
         super(Slider, self).__init__(viz, **kwargs)
         self.node = node
         self.base_output = node.output
         node.output = self.override_output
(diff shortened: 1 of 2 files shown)


nengo_viz/components/slider.py
```diff
@@ -3,13 +3,18 @@
...
+    def __init__(self, viz, node, min_value=-1, max_value=1, **kwargs):
         super(Slider, self).__init__(viz, **kwargs)
         self.node = node
         self.base_output = node.output
         node.output = self.override_output
         self.override = [None] * node.size_out
         self.value = np.zeros(node.size_out)
+        self.min_value = min_value
+        self.max_value = max_value
+        self.start_value = np.zeros(node.size_out)
```
cchan987: start value depends on np.zeros(node.size_out), what exactly does this do?

tcstewar: Good question.  Right now, the code is only handling start_value correctly if node.output is not callable.  If it is callable, then it just defaults back to np.zeros(node.size_out), which is what it was doing before (before this change the sliders would just always start at zero)

The next stage of this would be to handle a callable node.output correctly, which involves actually having the sliders move up and down on their own, reflecting the callable input (which is what the old visualizer does)

PR #57, opened 2015-02-12
Raster display time

tcstewar:
First attempt at making the raster plot have an option to display the time or not.  

There is a lot of duplicate code now between the raster and the value plots, so this should get cleaned up (in a separate PR).

(note: this PR should not be merged until after spike_raster (and it should get rebased to not have the duplicate commit))


cchan987:
I am not familiar with rebasing. I can confirm it does not conflict with my changes. I will leave this pull request in the knowledgeable hands of @tcstewar.

tcstewar:
Hmm, I'm actually not even sure this option should be there...  seems a weird thing to have in the one special case of raster plots.  So I think I'll just close it and we'll worry about it if/when graphs get reworked so this option can just happen for all graphs that have time-based x-axes.

PR #61, opened 2015-02-13
Tile view

cchan987:
This is a quick little component organizer that I put together to help with organizing the components for testing. 
-Prevents overlap on build

I found that it was useful for testing and can be easily modified later if we want to include it with the master branch.


cchan987:
I put the tiling of the components in the Javascript instead of python since we may want to introduce a button along the lines of 'click to organize components'

tcstewar:
Looks like there's some merge conflicts on this one, if it's still desired.  There's a very simple one on the server side in master already, if that's sufficient.

PR #64, opened 2015-03-02
Add jquery

cchan987:


cchan987:
A branch simply to add jquery

tcstewar:
Isn't jquery already in the include_bootstrap branch?  I think we're ready to merge that on into master....

PR #66, opened 2015-03-04
Labels

tcstewar:
This adds labels to every plot and adds tools for automatically setting labels based on the `locals()` dictionary (i.e. guessing a useful label for unlabelled items).


tcstewar:
Oops, I left ```tile_component``` in there....  if people want I can separate it out to its own PR -- it's a simple tiling option on the server side.

PR #68, opened 2015-03-04
Semantic pointer graph

tcstewar:
Added a Semantic Pointer graph showing the decoded value from a Vocabulary


cchan987:
This is pretty neat!

cchan987:
spa.py runs VERY slow on my computer. Around 0.04x

tcstewar:
Yup.  It's a big model (it's the giant circular convolution network that does it).  It's also a good indication that there should be a "building" indicator somewhere...  :)

PR #70, opened 2015-03-11
Server tile

cchan987:
Nicer view for user


tcstewar:
I'm fine with this for now, but I'd much happier if tile_components was called manually rather than automatically.  For example, right now nengo_gui automatically calls tile_components when it pops up nengo_viz.  And if people want to pop up nengo_viz themselves with a viz = nengo_viz.Viz(model) call, then they can call viz.tilecomponents

Still, if it's useful for Chris' class right now, go head.

PR #73, opened 2015-03-20
Pan cords

cchan987:


nengo_viz/static/viz_pan.js
```diff
@@ -0,0 +1,57 @@
...
+
+/*A Posn is an object with:
+ posn.x - int 
+ posn.y - int
+*/
+
+//Used for storing the cumulative x and y panning of the model
+VIZ.pan.cposn = {x:0, y:0};
+
+function init_main_events() {
```
amaziying: @cchan987 
There's something in javascript called self-invoking functions, or anonymous functions. Essentially, as soon as you define the function, it'll call itself right afterwards. Might make things a bit cleaner here.
http://www.w3schools.com/js/js_function_definition.asp

nengo_viz/static/viz_pan.js
```diff
@@ -0,0 +1,57 @@
...
+		});
+	}
+
+/*Pass this function the amount of change you want to apply to the screen*/
+VIZ.pan.shift = function(dx,dy) {
+	VIZ.pan.cposn.x += dx;
+	VIZ.pan.cposn.y += dy;
+	$('.graph').each(function(i, element){ // Get all the graph elements
+		var holde = $(element).css('transform').match(/(-?[0-9\.]+)/g); //Ugly method of finding the current transform of the element
+		VIZ.set_transform(element, Number(holde[4]) + dx, Number(holde[5]) + dy); // Do the transformation
```
amaziying: @cchan987 
I think data-x and data-y attributes need to be updated along with the transform.

PR #74, opened 2015-03-26
Network diagram

tcstewar:
A new component that shows the network diagram

Basic features:
- networks are expandable and collapsable (by clicking on them)
- automatic layer-based layouts
- automatically saves layout
- scrollwheel zooming (and resizing)
- drag to move items (and pan background)
- constrains items to stay inside their parent network

### DIFF
--- a/nengo_viz/components/netgraph.py
+++ b/nengo_viz/components/netgraph.py
@@ -88,8 +88,13 @@ class NetGraph(Component):
             config[self.viz.model].size = 1.0, 1.0
             try:
                 with open(self.viz.viz.filename + '.cfg') as f:
-                    config_code = f.read()
-                exec config_code in locals
+                    config_code = f.readlines()
+                for line in config_code:
+                    try:
+                        exec line in locals
+                    except Exception as e:
+                        print('error parsing config', line)
+                        print(e)
             except IOError:
                 pass
         else:


nengo_viz/static/viz.js
```diff
@@ -91,6 +95,14 @@ VIZ.Component = function(args) {
             target.style.width  = newWidth + 'px';
             target.style.height = newHeight + 'px';
             self.on_resize(newWidth, newHeight);
+            
+            var target = event.target;
```
amaziying: @tcstewar you've already defined target at line 84?

nengo_viz/static/viz_netgraph.js
```diff
@@ -0,0 +1,243 @@
...
+    /** Since connections may go to items that do not exist yet (since they
+     *  are inside a collapsed network), this dictionary keeps a list of
+     *  connections to be notified when a particular item appears.  The
+     *  key in the dictionary is the uid of the nonexistent item, and the
+     *  value is a list of VIZ.NetGraphConnections that should be notified
+     *  when that item appears. */
+    this.collapsed_conns = {}; 
+    
+    /** dragging the background pans the full area by changing offsetX,Y */
+    var self = this;
```
amaziying: Usually declared at the very top of the function as a standard, but doesn't really matter here

nengo_viz/static/viz.js
tcstewar: Oops....   okay, I've removed that now...  :)

nengo_viz/static/viz_netgraph_conn.js
```diff
@@ -0,0 +1,155 @@
...
+        }
+    }
+    console.log('could not find post');
+    console.log(this.posts);
+}
+
+
+/** remove this connection */
+VIZ.NetGraphConnection.prototype.remove = function() {
+    if (this.parent != null) {
```
amaziying: if (this.parent)
is sufficient :)

nengo_viz/static/viz_netgraph.js
tcstewar: Fair enough -- it'd be a good standard to stick with to try to define variables like that as close to the top of the constructor as possible...  lemme rearrange it...

nengo_viz/static/viz_netgraph_item.js
```diff
@@ -0,0 +1,452 @@
...
+            
+            item.constrain_position();                
+            ng.notify({act:"pos_size", uid:uid, 
+                       x:item.pos[0], y:item.pos[1],
+                       width:item.size[0], height:item.size[1]});
+                       
+            event.stopPropagation();           
+            });
+            
+    if (info.type == 'net') {
```
amaziying: When you can, use '===' instead of '==', because it's always better to have strict-er conditions

nengo_viz/static/viz_netgraph_item.js
```diff
@@ -0,0 +1,452 @@
...
+
+/** collapse an expanded network */
+VIZ.NetGraphItem.prototype.collapse = function(report_to_server) {
+    this.g.classList.remove('expanded');
+
+    /** move the label back to the middle */
+    this.label.setAttribute('transform', '');
+    
+    /** remove child NetGraphItems and NetGraphConnections */
+    while (this.child_connections.length > 0) {
```
amaziying: while (this.child_connections.length){}
is sufficient

nengo_viz/static/viz_netgraph_item.js
```diff
@@ -0,0 +1,452 @@
...
+        this.child_connections[0].remove();
+    }
+    while (this.children.length > 0) {
```
amaziying: while (this.child_connections.length){}
is sufficient

nengo_viz/static/viz_netgraph_conn.js
tcstewar: Fair enough.  I'm a bit cautious with the javascript boolean conversion rules, but you're right this one's pretty straightforward.

nengo_viz/static/viz_netgraph_item.js
amaziying: maybe it's better javascript convention to go through the array and just remove each. your way works perfectly fine though.
i.e.

this.child_connections.forEach(function (child) {
 child.remove();
});

nengo_viz/static/viz_netgraph_item.js
```diff
@@ -0,0 +1,452 @@
...
+/** remove the item from the graph */
+VIZ.NetGraphItem.prototype.remove = function() {
+    if (this.expanded) {
+        /** collapse the item, but don't tell the server since that would
+         *  update the server's config */
+        this.collapse(false);
+    }
+
+    /** remove the item from the parent's children list */
+    if (this.parent != null) {
```
amaziying: if (this.parent)
is sufficient

nengo_viz/static/viz_netgraph_conn.js
tcstewar: Hmm... on second thought, I'm very tempted to stick with a standard that all if conditions should evaluate to booleans, rather than relying on automatic boolean conversion.  As I was going though looking at places where I could replace ```if (x != null)``` with ```if (x)```, I started getting confused as to which ones were actually boolean variables and which ones were doing this auto-type conversion.

That said, I'm definitely open to discussion on this one, but for now I like the explicit approach.

nengo_viz/static/viz_netgraph_item.js
```diff
@@ -0,0 +1,452 @@
...
+        conn.redraw();
+    }
+
+    /** remove from the SVG */
+    this.ng.g_items.removeChild(this.g);    
+}
+
+VIZ.NetGraphItem.prototype.constrain_position = function() {
+    var changed = false;
+    if (this.parent != null) {
```
amaziying: if (this.parent)

nengo_viz/static/viz_netgraph_item.js
```diff
@@ -0,0 +1,452 @@
...
+    var screen_h = this.get_nested_height() * h * this.ng.scale;
+        
+    if (screen_w < this.minWidth) {
+        screen_w = this.minWidth;
+    }
+    if (screen_h < this.minHeight) {
+        screen_h = this.minHeight;
+    }
+    
+    if (this.type == 'ens') {
```
amaziying: again, use '===' when possible

nengo_viz/static/viz_netgraph_item.js
tcstewar: Ooo...  good catch.  In fact, I should also be using ```!==``` in a few places as well.

nengo_viz/static/viz_netgraph_item.js
tcstewar: The problem with the forEach approach here is that the ```remove``` function is deleting things from the array, so the forEach gets confused and misses half the items.  I could duplicate the array first and then do a forEach on the duplicate, but that seemed messy.

PR #76, opened 2015-03-29
Network menu

tcstewar:
This is a reworking of the config_menu branch (which has gotten a bit out of date).  It gives a simple menu for both the network graph items and the VIZ.Component plots as well.

### DIFF
--- a/nengo_viz/static/viz_menu.js
+++ b/nengo_viz/static/viz_menu.js
@@ -1,13 +1,26 @@
 
-VIZ.Menu = function(div, item) {
-    this.visible = false;
-    this.div = div;
-    this.menu = null;
-    this.actions = null;
+/**
+ * Create a menu that will appear inside the given div
+ */
+VIZ.Menu = function(div) {
+    this.visible = false;   // whether it's currently visible
+    this.div = div;         // the parent div
+    this.menu = null;       // the div for the menu itself
+    this.actions = null;    // the current action list for the menu
 };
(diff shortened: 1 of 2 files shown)


nengo_viz/static/viz.js
```diff
@@ -127,6 +139,18 @@ VIZ.Component.prototype.on_resize = function(width, height) {};
...
+
+VIZ.Component.prototype.generate_menu = function() {
+    var self = this;
+    var items = [];
+    if (this.label_visible) {
+        items.push(['hide label', function() {self.hide_label();}]);
+    } else {
+        items.push(['show label', function() {self.show_label();}]);    
+    }
+    return items;
```
amaziying: items becomes a single 1 item array with an array inside i.e.
[['show label', function()...]]

Not sure if that is your intended effect...

amaziying: I think by javascript convention, the item in items should be objects as opposed to arrays, then you can refer to them later by accessing the keys of the object as opposed to the indices of the array.

so for example,

items.push({
 title: 'hide label',
 action: function () {
    self.hide_label();
 }
});

tcstewar: Yeah, I could see doing each item as a dictionary instead of as a list.  that sort of change will probably start to be necessary as we allow more complex menu stuff (like disabled items and checkboxes and whatever else we end up wanting).

PR #83, opened 2015-04-02
updated listener for pan

cchan987:
Using the interact event so that panning listeners are consistent for network diagram and graphs.


tcstewar:
Cool!  It's always a nice sign when code gets deleted.   (It took me the longest time to learn that as a programmer)   :)

PR #85, opened 2015-04-06
deleted mouse out of browser detection

cchan987:


tcstewar:
Looks good, but you should also remove the line in templates/page.html that tries to open this file you just deleted.

PR #88, opened 2015-04-08
Pan zoom cleanup

cchan987:
functionized implementation of scroll zooming and added save layout after resize

### DIFF
--- a/nengo_viz/static/viz.js
+++ b/nengo_viz/static/viz.js
@@ -143,7 +143,6 @@ VIZ.Component = function(parent, args) {
 
     /** flag whether there is a scheduled update that hasn't happened yet */
     this.pending_update = false;
-    //VIZ.pan.events();
     
     this.menu = new VIZ.Menu(self.parent);
     interact(this.div)


examples/basic.py.cfg
```diff
@@ -0,0 +1,30 @@
+_viz_0 = nengo_viz.Value(ens)
```
tcstewar: Oops, this file probably shouldn't be in the commit....  :)

nengo_viz/static/viz_scale.js
```diff
@@ -23,8 +18,8 @@ VIZ.scale.zoom = function (wheel, event){
 		var offsetY = VIZ.pan.cposn.ul.y;
 	}
 	else{
-		var offsetX = event.offsetX;
-		var offsetY = event.offsetY;
+		var offsetX = event.offsetX || event.clientX;
+		var offsetY = event.offsetY || event.clientY;
```
tcstewar: Hmm...  not sure what this is doing.  Why is this code looking in two different places for the x,y values?  In what situations are they in one place or in the other?

Also, note that if ```event.offsetX``` is 0 and ```event.clientX``` is undefined, then this will set ```offsetX``` to ```undefined``` (since ```0 || undefined``` evaluates to ```undefined```).  Is this the intended behaviour?

tcstewar:
I just tested this in firefox and the scrollwheel zooming is broken for the plots (they always get bigger and never get smaller)

cchan987:
I tested this branch and it's looking ready.

tcstewar:
Cool, that looks good to me (and fixes the Firefox bug!)

PR #92, opened 2015-04-09
Slider set value

cchan987:
Merge pan_zoom_cleanup first

### DIFF
--- a/nengo_viz/static/viz.js
+++ b/nengo_viz/static/viz.js
@@ -51,7 +51,6 @@ VIZ.Component = function(parent, args) {
     this.width = args.width;
     this.height = args.height;
     
-    console.log(VIZ.pan.cposn.ul.x);
     var transform_val = cord_map(VIZ.pan.cposn, {x:args.x, y:args.y});
 	VIZ.set_transform(this.div, transform_val.x, transform_val.y);
     
@@ -156,7 +155,6 @@ VIZ.Component = function(parent, args) {
     this.menu = new VIZ.Menu(self.parent);
     interact(this.div)
         .on('tap', function(event) {
-            console.log("this is launching the menu");
             if (event.button == 0) {
                 if (self.menu.visible_any()) {
                     self.menu.hide_any();
--- a/nengo_viz/static/viz_slider.js
+++ b/nengo_viz/static/viz_slider.js
@@ -39,8 +39,8 @@ VIZ.Slider = function(parent, args) {
         valueDisplay.classList.add('unselectable')
         valueDisplay.innerHTML = slider.value;
         slider.div.appendChild(valueDisplay);
-        this.value_display = valueDisplay
-        $(valueDisplay).on('keypress', my_foo);
+        slider.value_display = valueDisplay
+        
 
         /** put the slider in the container */
         slider.div.style.position = 'fixed';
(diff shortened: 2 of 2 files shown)


nengo_viz/static/viz.js
```diff
@@ -454,3 +454,31 @@ VIZ.make_colors = function(N) {
...
     return c;
 }
+
+//Check if value is a number
+VIZ.is_num = function(value){
+    if (!(isNaN(value)) && !(value.trim() == '') ) {
+        return true;
+    }
+    else{
+        alert('Not a valid number! Please try again.');
```
tcstewar: Yikes!  I would really not expect a helper function like ```is_num``` to pop up an alert on a failure.  That should be the responsibility of whoever calls ```is_num```.

nengo_viz/static/viz.js
```diff
@@ -454,3 +454,31 @@ VIZ.make_colors = function(N) {
...
+//boundary value if input is above/below max/min
+//Otherwise outputs the input
+VIZ.max_min = function(value, min, max) {
+    if (value < min) {
+        alert('value below range limits, using value: ' + min + ' instead.')
+        return min;
+    }
+    else if (value > max) {
+        alert('value above range limits, using value: ' + max + ' instead.')
+        return max;
```
tcstewar: As with the above comment, these alerts should be removed.

nengo_viz/static/viz_scale.js
```diff
@@ -80,5 +74,6 @@ VIZ.scale.redraw_size = function(scale_x, scale_y) {
 		var w = VIZ.Component.components[i].width;
 		var h = VIZ.Component.components[i].height;
 		VIZ.Component.components[i].on_resize(w * scale_x, h * scale_y);
+		VIZ.Component.components[i].save_layout();
```
tcstewar: The rest of the code avoids calling save_layout() inside a utility function like redraw_size.  Instead, it's always calling save_layout() from within the event handler (or at least, that was the intent -- if there are places I missed, let me know).  That way we avoid having a whole bunch of save_layout() calls happening as a side effect of a single action (which just overwhelms the communication between the client and server).  So if I put a save_layout() call inside VIZ.scale.redraw_size, then that means that any event that ends up calling redraw_size and doing something else is in danger of doing redundant saving.

nengo_viz/static/viz_slider.js
```diff
@@ -212,14 +212,44 @@ VIZ.Slider.prototype.generate_menu = function() {
...
 
+VIZ.Slider.prototype.user_value = function () {
+    var slider_index = 0
+    if (this.sliders.length > 1) {
+        do{
+            slider_index = prompt('Enter the index of the slider to Change');
+            if (new_value == null) {
+                return;
+            }
+        } while (!(VIZ.is_num(slider_index)));
```
tcstewar: If the user inputs an invalid number, it should not keep asking the user until they give a valid answer.

nengo_viz/static/viz_slider.js
```diff
@@ -212,14 +212,44 @@ VIZ.Slider.prototype.generate_menu = function() {
...
     // add the parent's menu items to this
     // TODO: is this really the best way to call the parent's generate_menu()?
     return $.merge(items, VIZ.Component.prototype.generate_menu.call(this));
 };
 
+VIZ.Slider.prototype.user_value = function () {
+    var slider_index = 0
+    if (this.sliders.length > 1) {
+        do{
+            slider_index = prompt('Enter the index of the slider to Change');
```
tcstewar: I really don't like this way of setting the sliders by first typing in an index.  Why not do it the way the java visualizer does it?

PR #103, opened 2015-04-17
keep simcontrol on top

cchan987:


tcstewar:
looks good to me -- nice simple fix that seems a lot simpler than the dynamic zindex approach.  It's still a bit odd, but not sure what to do about that.

PR #104, opened 2015-04-21
Rename screen

cchan987:


nengo_viz/static/viz_pan.js
```diff
@@ -34,12 +34,12 @@ VIZ.pan.shift = function(dx,dy) {
 /*snap_to pans the screen to the specified posn cords quickly. 
 Effectively the same as changing the cposn cords to the posn points, and panning the screen accordingly*/
```
tcstewar: This comment refers to the old cposn name and should be updated to refer to VIZ.Screen

Issue #97, opened 2015-04-14
Minimalistic set value option for sliders


cchan987:
When setting value, user shouldn't have to type in a modal.
Instead, change the text value field on the slider by typing in it

Issue #86, opened 2015-04-06
Prevent shown_time bar from extending beyond the bounds of time_slider div


cchan987:

tcstewar:
Can you describe this bug?  I thought I had it checking for this.  that's what this code is for:

```javascript
if (new_time > self.last_time - self.shown_time) {
                new_time = self.last_time - self.shown_time;
            }
            if (new_time < self.last_time - self.kept_time) {
                new_time = self.last_time - self.kept_time;
            }
```

cchan987:
Here's a screenshot:
![image](https://cloud.githubusercontent.com/assets/10402911/7070169/35e1579e-deab-11e4-9b8d-1388cc5663e9.png)

First extend the shown_time bar so that it fills the whole time_slider area, then when you continue to drag outwards, it will go off the right hand side

tcstewar:
Ah, I see -- the left edge is being checked, but not the right edge.

tcstewar:
Fixed in #155

PR #110, opened 2015-04-27
Slider set value

cchan987:
merge 1st

### DIFF
--- a/nengo_viz/static/viz_slider.js
+++ b/nengo_viz/static/viz_slider.js
@@ -58,13 +58,14 @@ VIZ.Slider = function(parent, args) {
             function(event) {
                 /** check if click was the middle mouse button */
                 if (event.which == 2){
-                    self.set_value(this.slider.index, 0);
+                    self.set_value(this.slider.index, 0, true);
                 }
             }
         );
 
         interact(slider.div)
             .on('tap', function(event) {
+                console.log('crazy:', event);
                 var ind = event.currentTarget.slider.index;
                 self.input_set_value(ind);
                 event.stopPropagation();


cchan987:
solves #97

cchan987:
solves #86

tcstewar:
Bug report: I get very strange behaviour if I have multiple sliders and I use the click-to-set-value approach.  The slider values end up blank sometimes.  To replicate:

1) start up basic.py
2) make sliders for both Nodes
3) Click on one slider's value to start editting it
4) before typing anything, click on the other slider's value to start editting that.

The second slider's value will disappear, but you are still editting the first slider's value.  You also now cannot edit the second slider's value until you drag the slider to move it.

nengo_viz/static/viz_netgraph.js
```diff
@@ -119,7 +120,7 @@ VIZ.NetGraph = function(parent, args) {
                     self.menu.show(event.clientX, event.clientY, 
                                    self.generate_menu());
                 }
-                event.stopPropagation();  
+                //event.stopPropagation();
```
tcstewar: Why was this removed?

nengo_viz/static/viz_slider.js
```diff
@@ -9,6 +9,11 @@ VIZ.Slider = function(parent, args) {
     VIZ.Component.call(this, parent, args);
     var self = this;
 
+    //Check if user is filling in a number into a slider
+    self.filling_slider_value = false;
+
+    self.notify_msgs = [];
+
```
tcstewar: In all of our other code, this is done with ```this``` rather than ```self```.  We've been using ```self``` only when needed (inside functions that are defined inside one of this class methods).

nengo_viz/static/viz_slider.js
```diff
@@ -35,10 +40,12 @@ VIZ.Slider = function(parent, args) {
         slider.value = args.start_value[i];
 
         /** Show the slider Value */
-        var valueDisplay = document.createElement('p');
-        valueDisplay.classList.add('unselectable')
+        var valueDisplay = document.createElement('div');
+        valueDisplay.classList.add('value_display')
+        valueDisplay.id = 'value_display';
```
tcstewar: Hmm...  I'm pretty sure HTML ```id``` values have to be unique.  I'm not sure what problems this might cause, but it might be related to the wrong slider being modified sometimes.

nengo_viz/static/viz_slider.js
```diff
@@ -212,12 +226,105 @@ VIZ.Slider.prototype.generate_menu = function() {
...
     return $.merge(items, VIZ.Component.prototype.generate_menu.call(this));
 };
 
+VIZ.Slider.prototype.input_set_value = function(ind) {
+    var self = this;
+    if (self.filling_slider_value){
+        return;
+    }
+    else{
+        VIZ.netgraph.menu.hide_any();
```
tcstewar: Why not ```this.menu.hide_any()```?  That should be the same...  that was my intent with the ```hide_any``` function so you didn't have to do things like this.

nengo_viz/static/viz_slider.js
```diff
@@ -212,12 +226,105 @@ VIZ.Slider.prototype.generate_menu = function() {
...
+        var msg = document.getElementById('value_in_field').value;
+        $(text_div).off('keypress');
+        if (VIZ.is_num(msg)) {
+            self.set_value(ind, VIZ.max_min(Number(msg), slider_range[1], slider_range[0]));
+            return;
+        }
+        else {
+            alert('failed to set value');
+            text_div.innerHTML = 0;
+            self.set_value(ind, VIZ.max_min(0, slider_range[1], slider_range[0]));
```
tcstewar: Why does it reset to 0 on an invalid value, rather than staying at whatever value it was at?

nengo_viz/static/viz_slider.js
```diff
@@ -212,12 +226,105 @@ VIZ.Slider.prototype.generate_menu = function() {
...
+    msg = this.notify_msgs[0];
+    this.ws.send(msg);
+    if (this.notify_msgs.length > 1) {
+        var self = this;
+        window.setTimeout(function() {
+            self.send_notify_msg();
+        }, 1);
+    }
+    this.notify_msgs.splice(0, 1);
+}
```
tcstewar: Why was this needed?  All this code is mostly copied from VIZ.NetGraph, so if we need this sort of functionality, then we should make it a useful class rather than copying code.

nengo_viz/static/viz_slider.js
```diff
@@ -212,12 +226,105 @@ VIZ.Slider.prototype.generate_menu = function() {
...
+
+VIZ.Slider.prototype.user_value = function () {
+    var self = this;
+    var new_value = prompt('Set value\nExample: 1.3, 4.2');
```
tcstewar: In the Nengo 1.4 visualizer, the example would have as many items as there were sliders.  This example gets confusing if you don't have exactly 2 sliders.

nengo_viz/static/viz_slider.js
```diff
@@ -212,12 +226,105 @@ VIZ.Slider.prototype.generate_menu = function() {
...
+    
+    if (new_value == null) {
+        return;
+    };
+    new_value = new_value.split(',');
+    var slider_range = this.scale.domain();
+
+    for (var i = 0; i < this.sliders.length; i++){
+        if (!(VIZ.is_num(new_value[i]))) {
+            alert("invalid input :" + new_value[i] + "\nFor the slider in position " + (i +1) );
```
tcstewar: I'm very much not fond of popping up a whole bunch of alert messages. In the old visualizer, we just failed silently (i.e. no alert message at all).

nengo_viz/static/viz_slider.js
```diff
@@ -212,12 +226,105 @@ VIZ.Slider.prototype.generate_menu = function() {
...
+            break;
+        }
+        insert_value = VIZ.max_min(new_value[i], slider_range[1], slider_range[0]);
+
+        self.set_value(i, insert_value);
+    }
+    //this.save_layout();
```
tcstewar: commented out code should be removed

nengo_viz/static/viz_menu.js
```diff
@@ -38,7 +38,8 @@ VIZ.Menu.prototype.show = function (x, y, items) {
      * clicked on, we need the menu's zIndex to be very large
      * TODO: change this to be one more than the highest existing zIndex
      */
-    this.menu.style.zIndex = "999999999";
+    VIZ.max_zindex++;
+    this.menu.style.zIndex = VIZ.max_zindex;
```
tcstewar: Does this have to be part of this PR, or did it just get left in here?

nengo_viz/static/viz_slider.js
tcstewar: never mind; I see you removed it in a separate PR.  Note that this is why we try to keep PRs self-contained :)

tcstewar:
Another bug report:  clicking to edit the slider value does not cause any displayed popup menu to go away

tcstewar:
Using the arrow keys while editting a slider value doesn't seem to work in Firefox (except for the very first arrow key press).

tcstewar:
Also, is there any way to click to move the cursor?  That doesn't seem to be supported.

nengo_viz/static/viz_slider.js
cchan987: > Why does it reset to 0 on an invalid value, rather than staying at whatever value it was at?

It was my thinking that a reset to zero would serve also serve as an alert that the value set had failed. but now considering that 0 may be out of the range of the slider, this does seem to be flawed.

tcstewar: :) Wow, I hadn't even thought of the possibility that 0 is outside the range...  good point...

nengo_viz/static/viz_netgraph.js
cchan987: > Why was this removed?

I was testing some different Tap events for the sliders and this was interferring. 

However I didn't end up using those tap events so I'll turn this back on.

PR #111, opened 2015-04-27
Slider text function

cchan987:
Merge 2nd
Merge the slider_set_value branch first!

### DIFF
--- a/nengo_viz/static/viz_slider.js
+++ b/nengo_viz/static/viz_slider.js
@@ -388,9 +388,9 @@ VIZ.Slider.prototype.layout_info = function () {
     return info;
 };
 
-//slider_index: Nat, new_text: Num
+//slider_index: Nat, new_shown_value: Num
 //Rounds to 2 decimal places
-VIZ.Slider.prototype.update_value_text = function (slider_index, new_text) {
+VIZ.Slider.prototype.update_value_text = function (slider_index, new_shown_value) {
     var target = this.sliders[slider_index].value_display;
-    target.innerHTML = new_text.toFixed(2);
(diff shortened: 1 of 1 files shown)


cchan987:
solves #105

nengo_viz/static/viz_slider.js
```diff
@@ -239,3 +345,9 @@ VIZ.Slider.prototype.layout_info = function () {
     info.max_value = this.scale.domain()[0];
     return info;
 };
+
+VIZ.Slider.prototype.update_value_text = function (slider_index, new_text) {
+    var self = this;
```
tcstewar: For style consistency, this should be removed and just use ```this``` instead.

If we want to, we could go through all the code and use ```self``` everywhere (although I'd like to see the various reasons for and against that).  However, right now that's not the approach we've been doing, so we should keep the code style consistent.

tcstewar:
This PR breaks the set value capability when there is more than one slider.

To replicate, open up 2d.py and try using the "set value" option in the menu for the slider.

cchan987:
This one is looking good for testing/merging

nengo_viz/static/viz_slider.js
```diff
@@ -386,3 +387,10 @@ VIZ.Slider.prototype.layout_info = function () {
     info.max_value = this.scale.domain()[0];
     return info;
 };
+
+//slider_index: Nat, new_text: Num
+//Rounds to 2 decimal places
+VIZ.Slider.prototype.update_value_text = function (slider_index, new_text) {
```
tcstewar: ```new_text``` isn't text, so this variable name seems a bit weird.  ```new_value``` would be clearer.

PR #114, opened 2015-04-29
Set val range

cchan987:
merge 4th

### DIFF
--- a/nengo_viz/static/viz_slider.js
+++ b/nengo_viz/static/viz_slider.js
@@ -329,9 +329,17 @@ VIZ.Slider.prototype.set_range = function() {
     var new_range = prompt('Set range', '' + range[1] + ',' + range[0]);
     if (new_range !== null) {
         new_range = new_range.split(',');
-        var min = parseFloat(new_range[1]);
-        var max = parseFloat(new_range[0]);
-        this.scale.domain([min, max]);
+        var max = new_range[1];
+        var min = new_range[0];
+        if (!(VIZ.is_num(min)) || !(VIZ.is_num(max)) ) {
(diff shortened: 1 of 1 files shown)


nengo_viz/static/viz_slider.js
```diff
@@ -212,24 +226,125 @@ VIZ.Slider.prototype.generate_menu = function() {
...
-        var max = parseFloat(new_range[0]);
-        this.scale.domain([min, max]);
+        var max = new_range[1];
+        var min = new_range[0];
+        if (!(VIZ.is_num(min)) || !(VIZ.is_num(max)) ) {
+            alert('Invalid input');
+            return;
+        }
+        if (max < min){
+            alert('First value must be less than the second value.');
```
tcstewar: Why?  What if I want the numbers the other way around?

nengo_viz/static/viz_slider.js
```diff
@@ -212,24 +226,125 @@ VIZ.Slider.prototype.generate_menu = function() {
...
     var new_range = prompt('Set range', '' + range[1] + ',' + range[0]);
     if (new_range !== null) {
         new_range = new_range.split(',');
-        var min = parseFloat(new_range[1]);
-        var max = parseFloat(new_range[0]);
-        this.scale.domain([min, max]);
+        var max = new_range[1];
+        var min = new_range[0];
+        if (!(VIZ.is_num(min)) || !(VIZ.is_num(max)) ) {
+            alert('Invalid input');
```
tcstewar: When giving an error message like that, include what caused the error (either ```max``` or ```min```)

PR #118, opened 2015-04-30
Disappearing menu

### LINKED ISSUE #115: Menu does not disappear when dragging
Right now the popup menu disappears when you click elsewhere.  Should it also disappear when you drag elsewhere?  In particular, it seems weird that it doesn't go away when I start moving sliders up and down, or moving objects in the netgraph.

cchan987:
Fixes #115


tcstewar:
Nice... looks good to me.

PR #120, opened 2015-05-01
Re-implemented adjustable time on the SimControl

tcstewar:
This is a re-implementation of the adjust_shown_time branch which was accidentally merged and then all of its changes were manually deleted (rather than removing the commits).  

The feature added here is that the SimControl slider bar can now be resized by dragging on its edges.  This changes the amount of time shown on all the graphs in the system.


cchan987:
image

There is an issue where when lengthening the shown time slider when it already takes up the whole time_display box makes the shown_time slider extend beyond the bounds. (see pic above)

tcstewar:
#155 is better than this one, so I'm closing this version

PR #136, opened 2015-05-11
Immobile sim times

cchan987:
Prevents times from racing across the sim_control bar.

### DIFF
--- a/nengo_gui/static/sim_control.js
+++ b/nengo_gui/static/sim_control.js
@@ -261,11 +261,32 @@ Nengo.TimeSlider = function(args) {
     this.axis = d3.svg.axis()
         .scale(this.kept_scale)
         .orient("bottom")
-        .ticks(10);
+        .ticks(0);
     this.axis_g = this.svg.append("g")
         .attr("class", "axis")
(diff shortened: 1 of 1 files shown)


cchan987:
addresses #3

cchan987:
should we cut the time: XX:xx field from the sim control's metric container?

tbekolay:
I was curious what this looked like, so I rebased it to master. Here it is!

screen shot 2015-06-09 at 3 52 52 pm

I think with some tweaks this could be an improvement. I'm not sure where the numbers should be, but they look odd where they are.

Also, it would be super helpful if this also showed the times at the left and right of what is actually being shown (depending on the blue window).

tbekolay:
Oh, and also, a bit of bug: if you resize the window, the numbers don't move, and therefore disappear. Probably something needs to be added to the function called on resize.

Seanny123:
@tbekolay what if the number were overlayed on top of the line like so:
better bar

Seanny123:
On hold until #534 is resolved.

tcstewar:
This PR is rather out of date, and the current display seems to be working well, so I'm closing this one.

PR #148, opened 2015-05-19
Optimize update_client, lock safety

arvoelke:
Popping the first item repeatedly is O(n^2). (http://stackoverflow.com/questions/195625/what-is-the-time-complexity-of-popping-elements-from-list-in-python) This change makes `update_client` O(n).

Also made the lock safer with a context-manager. Previously, an exception in between would have caused a deadlock.

### DIFF
--- a/nengo_viz/components/value.py
+++ b/nengo_viz/components/value.py
@@ -28,9 +28,9 @@ class Value(Component):
         self.data.append(self.struct.pack(t, *x))
 
     def update_client(self, client):
-        while len(self.data) > 0:
-            data = self.data.pop(0)
-            client.write(data, binary=True)
+        for item in self.data:
+            client.write(item, binary=True)
+        self.data = []
 
     def javascript(self):
         info = dict(uid=self.uid, n_lines=self.n_lines, label=self.label)
--- a/nengo_viz/components/value.py
+++ b/nengo_viz/components/value.py
@@ -1,6 +1,7 @@
 import nengo
 import numpy as np
 import struct
+import collections
 
 from nengo_viz.components.component import Component, Template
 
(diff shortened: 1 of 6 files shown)


tcstewar:
This all looks good to me, once the deque change is made.  :)

tcstewar:
I just rebased this and added the deque change.  If someone gives me a +1 I'll merge...  :)

hunse:
There are a lot of other components that still use a list. We should switch them over as well.

nengo_viz/components/value.py
```diff
@@ -1,6 +1,7 @@
 import nengo
 import numpy as np
 import struct
+import collections
```
hunse: I know this is anal, but this import should go first.

tcstewar: <grins> I was just looking at exactly that and thinking I should fix it, but then thinking "meh, no one's going to be that anal about it...."  ;)

I'll move it (and the struct import) above and add the blank line.  :)

tcstewar:
> There are a lot of other components that still use a list. We should switch them over as well.

Good point...  I'll do those as well...

nengo_viz/components/value.py
hunse: I looked at it, and thought "I really shouldn't comment on that", then I realized if I saw that later in the code I would be annoyed, and fixing it now would save time for me and all the people who would be annoyed by it later.

tcstewar:
All switched over to deques.  And fixed the import sequence...  ;)

tcstewar:
This is all rebased to master -- everyone happy with it?

tbekolay:
LGTM!

PR #152, opened 2015-05-19
WIP: Sparkle plots

hunse:
This is a first working version of sparkle plots (neuron spiking on a grid).

### DIFF
--- a/nengo_gui/components/__init__.py
+++ b/nengo_gui/components/__init__.py
@@ -3,6 +3,7 @@ from .value import Value, ValueTemplate
 from .xyvalue import XYValue, XYValueTemplate
 from .sim_control import SimControl, SimControlTemplate
 from .raster import Raster, RasterTemplate
+from .spike_grid import SpikeGrid, SpikeGridTemplate
 from .voltage import Voltage, VoltageTemplate
 from .pointer import Pointer, PointerTemplate
 from .netgraph import NetGraph, NetGraphTemplate
--- a/nengo_gui/components/spike_grid.py
+++ b/nengo_gui/components/spike_grid.py
@@ -1,0 +1,57 @@
+import nengo
+import numpy as np
+import struct
+
+from nengo_gui.components.component import Component, Template
+
+
+class SpikeGrid(Component):
+    def __init__(self, viz, config, uid, obj, n_neurons=None):
+        super(SpikeGrid, self).__init__(viz, config, uid)
+        self.obj = obj.neurons
+        self.data = []
+        self.label = viz.viz.get_label(obj)
+        self.max_neurons = self.obj.size_out
+        if n_neurons is None:
+            n_neurons = self.max_neurons
+        self.n_neurons = n_neurons
+        self.pixels_x = np.ceil(np.sqrt(self.n_neurons))
+        self.pixels_y = np.ceil(float(self.n_neurons) / self.pixels_x)
+        self.n_pixels = self.pixels_x * self.pixels_y
+        self.struct = struct.Struct('<%df' % (1 + self.n_pixels))
+
+    def add_nengo_objects(self, viz):
+        with viz.model:
+            self.node = nengo.Node(self.gather_data, size_in=self.obj.size_out)
+            self.conn = nengo.Connection(self.obj, self.node, synapse=None)
+
+    def remove_nengo_objects(self, viz):
+        viz.model.connections.remove(self.conn)
+        viz.model.nodes.remove(self.node)
+
+    def gather_data(self, t, x):
+        # TODO: pass only spiking neurons, using subclass of Nengo.Image?
+        #   Considerations include how to filter if we're only passing spike
+        #   times, i.e. need to write a new DataStore, or filter on server
+        #   side, but filtering server-side means we no longer have spikes
+        #   and have to send a lot more data.
+        y = np.zeros(self.n_pixels, dtype=x.dtype)
+        y[:x.size] = (x > 0.0) * 255.
+        data = self.struct.pack(t, *y)
+        self.data.append(data)
+
+    def update_client(self, client):
+        while len(self.data) > 0:
+            data = self.data.pop(0)
+            client.write(data, binary=True)
+
+    def javascript(self):
+        info = dict(uid=self.uid, label=self.label,
+                    pixels_x=self.pixels_x, pixels_y=self.pixels_y)
+        json = self.javascript_config(info)
+        return 'new Nengo.Image(main, sim, %s);' % json
+
+
+class SpikeGridTemplate(Template):
+    cls = SpikeGrid
+    config_params = dict(**Template.default_params)
--- a/nengo_gui/static/image.js
+++ b/nengo_gui/static/image.js
@@ -1,0 +1,127 @@
+/**
+ * Shows an image or pixel grid over time
+ * @constructor
+ *
+ * @param {dict} args - A set of constructor arguments (see Nengo.Component)
+ * @param {int} args.n_lines - number of decoded values
+ * @param {float} args.miny - minimum value on y-axis
+ * @param {float} args.maxy - maximum value on y-axis
+ * @param {Nengo.SimControl} args.sim - the simulation controller
+ */
+
+Nengo.Image = function(parent, sim, args) {
+    var self = this;
+
+    Nengo.Component.call(self, parent, args);
+    self.sim = sim;
+    self.display_time = args.display_time;
+    self.pixels_x = args.pixels_x;
+    self.pixels_y = args.pixels_y;
+    self.n_pixels = self.pixels_x * self.pixels_y;
+
+    /** for storing the accumulated data */
+    self.data_store = new Nengo.DataStore(self.n_pixels, self.sim, 0.01);
+
+    /** spacing between the graph and the outside edges (in pixels) */
+    self.margin_top = 30;
+    self.margin_bottom = 0;
+    self.margin_left = 0;
+    self.margin_right = 0;
+
+    /** draw the plot as an SVG */
+    self.svg = d3.select(self.div).append('svg')
+        .attr('width', '100%')
+        .attr('height', '100%')
+        .attr('style', [
+            'margin-top:', self.margin_top, 'px;',
+            'margin-bottom:', self.margin_bottom, 'px;',
+            'margin-left:', self.margin_left, 'px;',
+            'margin-right:', self.margin_right, 'px;'
+        ].join(""));
+
+    /** call schedule_update whenever the time is adjusted in the SimControl */
+    self.sim.div.addEventListener('adjust_time',
+            function(e) {self.schedule_update();}, false);
+
+    /** create the image */
+    self.image = self.svg.append("image")
+        .attr("x", 0)
+        .attr("y", 0)
+        .attr("width", "100%")
+        .attr("height", "100%")
+        .attr("style", [
+            "image-rendering: -webkit-optimize-contrast;",
+            "image-rendering: -moz-crisp-edges;",
+            "image-rendering: -o-crisp-edges;",
+            "image-rendering: crisp-edges;"
+        ].join(""));
+
+    self.canvas = document.createElement('CANVAS');
+    self.canvas.width = self.pixels_x;
+    self.canvas.height = self.pixels_y;
+
+    self.on_resize(args.width, args.height);
+
+};
+Nengo.Image.prototype = Object.create(Nengo.Component.prototype);
+Nengo.Image.prototype.constructor = Nengo.Image;
+
+/**
+ * Receive new line data from the server
+ */
+Nengo.Image.prototype.on_message = function(event) {
+    var self = this;
+    var data = new Float32Array(event.data);
+    self.data_store.push(data);
+    self.schedule_update();
+}
+
+/**
+ * Redraw the lines and axis due to changed data
+ */
+Nengo.Image.prototype.update = function() {
+    var self = this;
+
+    /** let the data store clear out old values */
+    self.data_store.update();
+
+    var data = self.data_store.get_last_data();
+    var ctx = self.canvas.getContext("2d");
+    var imgData = ctx.getImageData(0, 0, self.pixels_x, self.pixels_y);
+    for (var i = 0; i < self.n_pixels; i++) {
+        imgData.data[4*i + 0] = data[i];
+        imgData.data[4*i + 1] = data[i];
+        imgData.data[4*i + 2] = data[i];
+        imgData.data[4*i + 3] = 255;
+    }
+    ctx.putImageData(imgData, 0, 0);
+    var dataURL = self.canvas.toDataURL("image/png");
+
+    self.image.attr("xlink:href", dataURL);
+};
+
+/**
+ * Adjust the graph layout due to changed size
+ */
+Nengo.Image.prototype.on_resize = function(width, height) {
+    var self = this;
+    if (width < self.minWidth) {
+        width = self.minWidth;
+    }
+    if (height < self.minHeight) {
+        height = self.minHeight;
+    };
+
+    self.svg
+        .attr("width", width - self.margin_left - self.margin_right)
+        .attr("height", height - self.margin_top - self.margin_bottom);
+
+    self.update();
+
+    self.label.style.width = width;
+
+    self.width = width;
+    self.height = height;
+    self.div.style.width = width;
+    self.div.style.height = height;
+};
--- a/nengo_gui/static/image.js
+++ b/nengo_gui/static/image.js
@@ -52,8 +52,7 @@ Nengo.Image = function(parent, sim, args) {
         .attr("style", [
             "image-rendering: -webkit-optimize-contrast;",
             "image-rendering: -moz-crisp-edges;",
-            "image-rendering: -o-crisp-edges;",
-            "image-rendering: crisp-edges;"
+            "image-rendering: pixelated;"
         ].join(""));
 
     self.canvas = document.createElement('CANVAS');
--- a/nengo_gui/static/image.js
+++ b/nengo_gui/static/image.js
@@ -50,7 +50,7 @@ Nengo.Image = function(parent, sim, args) {
     self.canvas.width = self.pixels_x;
     self.canvas.height = self.pixels_y;
 
-    self.on_resize(args.width, args.height);
+    self.on_resize(this.get_screen_width(), this.get_screen_height());
 
 };
 Nengo.Image.prototype = Object.create(Nengo.Component.prototype);
--- a/nengo_gui/static/image.js
+++ b/nengo_gui/static/image.js
@@ -22,21 +22,12 @@ Nengo.Image = function(parent, sim, args) {
     /** for storing the accumulated data */
     self.data_store = new Nengo.DataStore(self.n_pixels, self.sim, 0.01);
 
-    /** spacing between the graph and the outside edges (in pixels) */
-    self.margin_top = 30;
-    self.margin_bottom = 0;
-    self.margin_left = 0;
-    self.margin_right = 0;
-
     /** draw the plot as an SVG */
     self.svg = d3.select(self.div).append('svg')
         .attr('width', '100%')
         .attr('height', '100%')
         .attr('style', [
-            'margin-top:', self.margin_top, 'px;',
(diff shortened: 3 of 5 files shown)


hunse:
I find that if I make a bunch of these (> 4) things get kind of slow. I'm not sure if this is more related to storing the data (which could be fixed by just storing spikes, though the DataStore would still have to calculate the smoothed value at each time step) or if the problem is with rendering the data.

tcstewar:
Looks like there's also ```image-rendering: pixelated;``` which seems to work in recent versions of Chrome  https://www.chromestatus.com/feature/5118058116939776

hunse:
This is the [page I was working off](https://developer.mozilla.org/en/docs/Web/CSS/image-rendering), but looking at it again, I'm not sure why I chose `crisp-edges` over `pixelated`. I feel like I tried `pixelated` and it didn't work for some reason, but it does seem like the more applicable choice.

jgosmann:
rebased to master

jgosmann:
Added a commit to use `pixelated` where available.

If I run a simulation for a while with a sparkle plot it crashes with this exception:
```
  File "/Volumes/Home/blubb/Documents/projects/uni/nengo_viz/nengo_viz/swi.py", line 227, in handle_ws_request
    getattr(self, command)(*[client] + args[1:], **db)
  File "/Volumes/Home/blubb/Documents/projects/uni/nengo_viz/nengo_viz/server.py", line 136, in ws_viz_component
    component.update_client(client)
  File "/Volumes/Home/blubb/Documents/projects/uni/nengo_viz/nengo_viz/components/spike_grid.py", line 46, in update_client
    client.write(data, binary=True)
  File "/Volumes/Home/blubb/Documents/projects/uni/nengo_viz/nengo_viz/swi.py", line 632, in write
    self.socket.send(header)
error: [Errno 35] Resource temporarily unavailable
```

jgosmann:
Also, the margins are defined in pixels with should be changed to something relative to the `scale` (or change to a relative unit like `em`).

hunse:
> Also, the margins are defined in pixels

I took that code from another object, so maybe someone can figure out where I got it from and apply the appropriate change to ems?

> f I run a simulation for a while with a sparkle plot it crashes

How long did you have to run it for? I ran for a minute or two without any signs of crashing (which is arguably not very long). I made two sparkle plots, though, and that made everything grid to a halt. Long-term, I think we're going to need a better way to do sparkle plots. (I'm not sure if the bigger problem is creating the images on the client side, filtering the data on the client side, or just the amount of data sent from the server to the client.)

jgosmann:
Yeah, I can to the em change when I find some more free time.

I didn't have to run it for very long. Only a seconds and it's reproducible.

hunse:
> I didn't have to run it for very long. Only seconds and it's reproducible.

Is that on your laptop? Is it the same on your desktop? I wouldn't be surprised if this is related to processor speed (as I said, I don't see it on my machine). How many neurons in your plot?

My only idea for a quick fix is to limit the maximum number of neurons in the plot. Reducing the amount of data sent from the server to the client (by only sending spikes) could help too, but that's more involved.

hunse:
One other idea I just had for this is to do the sparkle plots like @tcstewar did plotting for the neuromorphic cameras: We just send spike times (and the index of the spiking neuron) from server to client, and then we set the pixel corresponding to that neuron in the image to 1. Each time step, we scale all pixels in the image by a fixed ratio (e.g. 0.9), which performs some filtering. Ideally we'd do this in something like WebGL I think, but it might help in the current SVG setup, too.

hunse:
But that's all probably stuff for later, unless someone's really keen on doing it. For now, let's just add the very buggy sparkle plots.

tcstewar:
I get the same problem as @jgosmann but on window:

```
  File "c:\users\terry\documents\github\nengo_viz\nengo_viz\swi.py", line 227, in handle_ws_request
    getattr(self, command)(*[client] + args[1:], **db)
  File "c:\users\terry\documents\github\nengo_viz\nengo_viz\server.py", line 136, in ws_viz_component
    component.update_client(client)
  File "c:\users\terry\documents\github\nengo_viz\nengo_viz\components\spike_grid.py", line 46, in update_client
    client.write(data, binary=True)
  File "c:\users\terry\documents\github\nengo_viz\nengo_viz\swi.py", line 632, in write
    self.socket.send(header)
error: [Errno 10035] A non-blocking socket operation could not be completed immediately
```

My guess is that we're filling up the websocket buffer.  I'm curious what happens if we just handle that exception well....

tcstewar:
Putting this in ```spike_grid.py``` fixes the problem for me:

```python
        while len(self.data) > 0:
            data = self.data.pop(0)
            try:
                client.write(data, binary=True)
            except:
                pass
```

jgosmann:
> Is that on your laptop? Is it the same on your desktop? I wouldn't be surprised if this is related to processor speed (as I said, I don't see it on my machine). How many neurons in your plot?

Only laptop (which is faster than my desktop). Neurons whatever the number in `ens` in `net.py` is.

@tcstewar Maybe we should be a bit more specific in that except? Or at least log the exception somewhere (though that might be something for later because IMO we have to refactor how output is done anyways)

tcstewar:
> Maybe we should be a bit more specific in that except?

Yup, a socket.error seems to make the most sense to me, as I'm pretty sure that's what it is

> (though that might be something for later because IMO we have to refactor how output is done anyways)

I definitely agree.  But later, I think.  :(

jgosmann:
@tcstewar Your fix is not working for me. I get a bunch of `EAGAIN` errors and at some point `ECONNRESET` (connection reset by peer)`. I suppose, if we scoop to much data into the socket and we keep doing this at some point the system (or browser?) just disconnects?

tcstewar:
> Your fix is not working for me. I get a bunch of EAGAIN errors and at some point ECONNRESET (connection reset by peer)`. I suppose, if we scoop to much data into the socket and we keep doing this at some point the system (or browser?) just disconnects?

:(  Grr.  What happens if you ignore the EAGAIN errors?  

Also, what happens for you if you throttle the communication by doing something like only sending every 10th message?

jgosmann:
> :( Grr. What happens if you ignore the EAGAIN errors?

That's what I've been doing.

> Also, what happens for you if you throttle the communication by doing something like only sending every 10th message?

That works.

tbekolay:
This seems like it might not be an easy fix. Should we punt this one to after the release?

jgosmann:
I think, I commented on this just a minute ago, but somehow I didn't?

Anyways, we should try polling the socket with `select` whether it is ready for writing and if not just drop the data.

jgosmann:
hm, polling with select doesn't work ...
so yeah, it doesn't seem to be an easy fix

tbekolay:
I'm going to take this off the release milestone then. If we figure it out during the summer school, we can do another release.

jgosmann:
With #306 the only remaining issue for me seems to be that the plot opens with a size of (0, 0).

jgosmann:
Rebased to current master with (most of) the renames.

Seanny123:
Needs to be rebased again, but after that it's ready to merge.

tbekolay:
I thought there were issues with this being too slow to be used for some browsers / machines?

hunse:
I'm not sure how much it depends on browser or machine, but it can get pretty slow. I'm not sure if it's worth merging it as-is, just to have something, or waiting until we have a better version.

Seanny123:
I was talking about code quality, based off the assumption that the actual performance had been resolved somewhere in the history. This was a silly assumption.

The code seems to not work after I move the slider (it turns completely black) and the (grey) spikes only re-appear when I hit pause.

tcstewar:
> I'm not sure how much it depends on browser or machine, but it can get pretty slow. I'm not sure if it's worth merging it as-is, just to have something, or waiting until we have a better version.

I've got a quick fix that should make this a little more stable so that it'll at least be there until we work on a better version.  

The fix has three parts:  First, handle the communication error by just ignoring it (so that message gets dropped).  Second, do the filtering server-side (so that the dropped messages don't lead to weird visual artifacts).  With this version, there's no cost to doing the filtering server-side, since we're sending data for each pixel anyway.  Third, send data as uint8 rather than as float.

Together, those seem to make it a lot more stable.

Should I commit those changes to this branch?  Here's the diff, as an alternative:

```
diff --git a/nengo_gui/components/spike_grid.py b/nengo_gui/components/spike_grid.py
index 989d642..2bc178c 100644
--- a/nengo_gui/components/spike_grid.py
+++ b/nengo_gui/components/spike_grid.py
@@ -18,32 +18,38 @@ class SpikeGrid(Component):
         self.pixels_x = np.ceil(np.sqrt(self.n_neurons))
         self.pixels_y = np.ceil(float(self.n_neurons) / self.pixels_x)
         self.n_pixels = self.pixels_x * self.pixels_y
-        self.struct = struct.Struct('<%df' % (1 + self.n_pixels))
+        self.struct = struct.Struct('<%dB' % (1 + self.n_pixels))
+        self.max_value = 1.0

     def add_nengo_objects(self, viz):
         with viz.model:
             self.node = nengo.Node(self.gather_data, size_in=self.obj.size_out)
-            self.conn = nengo.Connection(self.obj, self.node, synapse=None)
+            self.conn = nengo.Connection(self.obj, self.node, synapse=0.01)

     def remove_nengo_objects(self, viz):
         viz.model.connections.remove(self.conn)
         viz.model.nodes.remove(self.node)

     def gather_data(self, t, x):
+        self.max_value = max(self.max_value, np.max(x))
+
         # TODO: pass only spiking neurons, using subclass of Nengo.Image?
         #   Considerations include how to filter if we're only passing spike
-        #   times, i.e. need to write a new DataStore, or filter on server
-        #   side, but filtering server-side means we no longer have spikes
-        #   and have to send a lot more data.
-        y = np.zeros(self.n_pixels, dtype=x.dtype)
-        y[:x.size] = (x > 0.0) * 255.
+        #   times, i.e. need to write a new DataStore.
+        y = np.zeros(self.n_pixels, dtype=np.uint8)
+        y[:x.size] = x * 255 / self.max_value
         data = self.struct.pack(t, *y)
         self.data.append(data)

         while len(self.data) > 0:
             data = self.data.pop(0)
-            client.write(data, binary=True)
+            try:
+                client.write(data, binary=True)
+            except:
+                # if there is a communication problem, just drop the frame
+                # (this usually happens when there is too much data to send)
+                pass

     def javascript(self):
         info = dict(uid=self.uid, label=self.label,
diff --git a/nengo_gui/static/image.js b/nengo_gui/static/image.js
index cc38f75..1e4395e 100644
--- a/nengo_gui/static/image.js
+++ b/nengo_gui/static/image.js
@@ -20,7 +20,7 @@ Nengo.Image = function(parent, sim, args) {
     self.n_pixels = self.pixels_x * self.pixels_y;

     /** for storing the accumulated data */
-    self.data_store = new Nengo.DataStore(self.n_pixels, self.sim, 0.01);
+    self.data_store = new Nengo.DataStore(self.n_pixels, self.sim, 0);

     /** draw the plot as an SVG */
     self.svg = d3.select(self.div).append('svg')
@@ -61,7 +61,7 @@ Nengo.Image.prototype.constructor = Nengo.Image;
  */
 Nengo.Image.prototype.on_message = function(event) {
     var self = this;
-    var data = new Float32Array(event.data);
+    var data = new Uint8Array(event.data);
     self.data_store.push(data);
     self.schedule_update();
 }
```

tcstewar:
If people want to try out the change I made above, there's a branch ```sparkle2``` with it.  @Seanny123  if you like the change, you can include it here just by cherry-picking 4b13f1820e1345877cdb94ff09cccb4ebaec823a

Could people try out the sparkle2 branch on their browsers and see if it works well for them?  :)

celiasmith:
Works nicely for me (mac, chrome), as far as speed/visualization go.  It acts a bit odd when resizing because it allows the drag frame to be any aspect ratio, but the graph is fixed 1:1.  

On Safari it looks really weird (like it's been visually filtered by something).

tcstewar:
> It acts a bit odd when resizing because it allows the drag frame to be any aspect ratio, but the graph is fixed 1:1.

True, that is a little weird and inconsistent with our other graphs...  I'm fine with that behaviour for now and then maybe tweak it in a future PR (it'd be kinda neat to have it keep the aspect ratio of the pixels but change the #rows and #cols as you resize it....  :)

> On Safari it looks really weird (like it's been visually filtered by something).

That's probably the resizing filter.  There are about five different competing standards for telling the image to just do a nearest-pixel resize, rather than linear smoothing, and I wouldn't be surprised if we haven't included the Safari-specific flag....  but for now I'm only worried about Chrome and Firefox.  :)

Seanny123:
Both `sparkle` and `sparkle2` work for me on Windows 8 Chrome. I'll let you know when I regain access to my problematic linux laptop to see which one works better.

tbekolay:
Just checked both of these out. I used the default model, but pumped up the number of neurons in A to 1000. A few things that I noticed:

- Neither `sparkle` nor `sparkle2` noticed when I changed the number of neurons in the population.
- `sparkle` failed after a while with the following error:
```
Exception happened during processing of request from ('127.0.0.1', 42604)
  File "/usr/lib/python2.7/SocketServer.py", line 599, in process_request_thread
    self.finish_request(request, client_address)
  File "/usr/lib/python2.7/SocketServer.py", line 334, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/usr/lib/python2.7/SocketServer.py", line 655, in __init__
    self.handle()
  File "/usr/lib/python2.7/BaseHTTPServer.py", line 340, in handle
    self.handle_one_request()
  File "/usr/lib/python2.7/BaseHTTPServer.py", line 328, in handle_one_request
    method()
  File "/home/tbekolay/Code/nengo_gui/nengo_gui/swi.py", line 210, in do_GET
    self.handle_request(args, db)
  File "/home/tbekolay/Code/nengo_gui/nengo_gui/swi.py", line 252, in handle_request
    self.handle_ws_request(args, db)
  File "/home/tbekolay/Code/nengo_gui/nengo_gui/swi.py", line 245, in handle_ws_request
    client.close()
  File "/home/tbekolay/Code/nengo_gui/nengo_gui/swi.py", line 707, in close
    self._sendall(packet)
  File "/home/tbekolay/Code/nengo_gui/nengo_gui/swi.py", line 721, in _sendall
    bytes_sent += self.socket.send(data[bytes_sent:])
error: [Errno 11] Resource temporarily unavailable
```
I didn't get any crashes with `sparkle2`.
- `sparkle` was noticeably more sluggish than `sparkle2` when moving a slider.
- `sparkle` had darker greys than `sparkle2`, for the same network (seed set). `sparkle2` could get some squares that were essentially white, while for `sparkle` they were more like a medium gray.
- Both were slower on Firefox than on Chrome.

In conclusion, `sparkle2` seems stable enough to me.

One thing that I'm not super sure about is the label that we use in the right click menu. Right now, it's "Spikes (grid)". I don't really know what that's going to be before I open it, and to be honest, I don't really know what it is once it's open. Is there some way we can make it more clear what people are seeing with this?

tcstewar:
> Neither sparkle nor sparkle2 noticed when I changed the number of neurons in the population.

Yup, that's a variant of #425 and is true about just about every graph in the system.  I'll be fixing that shortly...

> One thing that I'm not super sure about is the label that we use in the right click menu. Right now, it's "Spikes (grid)". I don't really know what that's going to be before I open it, and to be honest, I don't really know what it is once it's open. Is there some way we can make it more clear what people are seeing with this?

Hmm...  good question.  In Nengo 1.4 this was a "firing rate" plot, which might be slightly clearer...  any other suggestions for names?  I don't like "Sparkle plot" -- that's probably even less clear.

tcstewar:
How do people feel about "Firing pattern" as a name for this plot?  I think it gets the idea across, and it is the pattern aspect that's kinda fun about it....

Seanny123:
+1 for firing pattern

cchan987:
> How do people feel about "Firing pattern" as a name for this plot?

I also like firing pattern!

jgosmann:
> didn't get any crashes with sparkle2.

I did by changing the number of neurons. When going from high to low it seems that the plot stays at the same number, but shows black squares for the now removed neurons. When going from low to high I get this exception (on the console, not in the GUI):
```
Exception in thread Thread-3:
  File "/home/jgosmann/.pyenv/versions/2.7.9/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "/home/jgosmann/.pyenv/versions/2.7.9/lib/python2.7/threading.py", line 763, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/home/jgosmann/Documents/projects/nengo_gui/nengo_gui/page.py", line 463, in runner
    self.build()
  File "/home/jgosmann/Documents/projects/nengo_gui/nengo_gui/page.py", line 408, in build
    c.add_nengo_objects(self)
  File "/home/jgosmann/Documents/projects/nengo_gui/nengo_gui/components/spike_grid.py", line 29, in add_nengo_objects
    size_in=self.obj.neurons.size_out)
  File "/home/jgosmann/.pyenv/versions/2.7.9/lib/python2.7/site-packages/nengo-2.0.1-py2.7.egg/nengo/base.py", line 24, in __call__
    inst.__init__(*args, **kwargs)
  File "/home/jgosmann/.pyenv/versions/2.7.9/lib/python2.7/site-packages/nengo-2.0.1-py2.7.egg/nengo/node.py", line 123, in __init__
    self.output = output  # Must be set after size_out; may modify size_out
  File "/home/jgosmann/.pyenv/versions/2.7.9/lib/python2.7/site-packages/nengo-2.0.1-py2.7.egg/nengo/base.py", line 63, in __setattr__
    super(NengoObject, self).__setattr__(name, val)
  File "/home/jgosmann/.pyenv/versions/2.7.9/lib/python2.7/site-packages/nengo-2.0.1-py2.7.egg/nengo/node.py", line 30, in __set__
    result = self.validate_callable(node, output)
  File "/home/jgosmann/.pyenv/versions/2.7.9/lib/python2.7/site-packages/nengo-2.0.1-py2.7.egg/nengo/node.py", line 45, in validate_callable
    result, invoked = checked_call(output, *args)
  File "/home/jgosmann/.pyenv/versions/2.7.9/lib/python2.7/site-packages/nengo-2.0.1-py2.7.egg/nengo/utils/stdlib.py", line 31, in checked_call
    return CheckedCall(func(*args, **kwargs), True)
  File "/home/jgosmann/Documents/projects/nengo_gui/nengo_gui/components/spike_grid.py", line 44, in gather_data
    y[:x.size] = x * 255 / self.max_value
ValueError: Validation error when setting 'Node.output': could not broadcast input array from shape (500) into shape (56)
```

Also, I am not sure if some of the differences between `sparkle` and `sparkle2` are because they branch off master at different points.

tcstewar:
> When going from low to high I get this exception (on the console, not in the GUI):

Ah, good catch.  I think I can fix that fairly readily in sparkle2 now that #514 is in.

> Also, I am not sure if some of the differences between sparkle and sparkle2 are because they branch off master at different points.

Also a good point.  sparkle2 has been rebased to master, but the only actual changes from sparkle are:

https://github.com/nengo/nengo_gui/commit/496a157ef788132ed6e9003b845a72c8bcaeb45a
https://github.com/nengo/nengo_gui/commit/8e59bca44a1277b67a30b24408082fa082e10dd3
https://github.com/nengo/nengo_gui/commit/12dae0f0ee09f584d83c41a46db5ef8bf7b61775

Its sounding like the consensus is that the sparkle2 approach is fine.  Should I bring those changes into this branch, or do a PR on sparkle2?  I originally made the separate branch because @Seanny123 had assigned himself to this one and I didn't want to step on any toes.... :)

Seanny123:
Go ahead and step on my toes. I generally assign my stuff to things I think
I can handle. It's like tagging myself a to-do list. Should I not be using
it that way?

tcstewar:
> Go ahead and step on my toes. I generally assign my stuff to things I think
> I can handle. It's like tagging myself a to-do list. Should I not be using
> it that way?

That seems mostly fair, but it's also a marker to others that they shouldn't do anything on this, because there can only be one assignee at a time.

tcstewar:
I've started a new PR #522 for switching over to ```sparkle2``` and cleaning up the issues raised here.  The crash when changing the number of neurons is now fixed.  I'll retire this PR and discussion should continue in the other one.

PR #154, opened 2015-05-19
Voltage plots

hunse:
Here's a first stab at a voltage plot. I tested it on `basic.py`, and though the output looks a little funny, it's actually correct (I checked it with plotting directly using matplotlib). The funniness is due to the filtering; maybe we want to reduce or turn off the filtering for voltage plots?

### DIFF
--- a/nengo_viz/components/voltage.py
+++ b/nengo_viz/components/voltage.py
@@ -41,7 +41,7 @@ class Voltage(Component):
         trange = sim.trange()[-len(data):]
 
         for t, datum in zip(trange, data):
-            datum = (datum + np.arange(self.n_neurons)) / self.n_neurons
+            datum = (datum + np.arange(self.n_neurons))
             packet = self.struct.pack(t, *datum)
             client.write(packet, binary=True)
 
@@ -55,4 +55,4 @@ class Voltage(Component):
 class VoltageTemplate(Template):
     cls = Voltage
     config_params = dict(
-        max_value=1.0, min_value=0.0, **Template.default_params)
+        max_value=5.0, min_value=0.0, **Template.default_params)
--- a/nengo_viz/components/voltage.py
+++ b/nengo_viz/components/voltage.py
@@ -1,0 +1,54 @@
+import nengo
+import numpy as np
+import struct
+
+from nengo_viz.components.component import Component, Template
+
+
+class Voltage(Component):
+    def __init__(self, viz, config, uid, obj, n_neurons=10):
+        super(Voltage, self).__init__(viz, config, uid)
+        self.viz = viz
+        self.obj = obj.neurons
+        self.data = []
+        self.label = viz.viz.get_label(obj)
+        self.max_neurons = self.obj.size_out
+        self.n_neurons = min(n_neurons, self.max_neurons)
+        self.struct = struct.Struct('<%df' % (1 + self.n_neurons))
+
+    def add_nengo_objects(self, viz):
+        with viz.model:
+            self.probe = nengo.Probe(self.obj[:self.n_neurons], 'voltage')
+
+    def remove_nengo_objects(self, viz):
+        viz.model.probes.remove(self.probe)
+
+    def format_data(self, t, x):
+        data = self.struct.pack(t, *x[:self.n_neurons])
+        self.data.append(data)
+
+    def update_client(self, client):
+        sim = self.viz.sim
+        if sim is None:
+            return
+
+        # TODO: this is hack to delete probe data in Nengo 2.0.1, since we
+        # can't limit the size of probes. Fix this up with Nengo 2.1.
+        data = sim.data.raw[self.probe][:]
+        del sim.data.raw[self.probe][:]  # clear the data
+        trange = sim.trange()[-len(data):]
+
+        for t, datum in zip(trange, data):
+            packet = self.struct.pack(t, *datum)
+            client.write(packet, binary=True)
+
+    def javascript(self):
+        info = dict(uid=self.uid, label=self.label, n_lines=self.n_neurons)
+        json = self.javascript_config(info)
+        return 'new VIZ.Value(main, sim, %s);' % json
+
+
+class VoltageTemplate(Template):
+    cls = Voltage
+    config_params = dict(
+        max_value=1.0, min_value=0.0, **Template.default_params)
--- a/nengo_viz/components/voltage.py
+++ b/nengo_viz/components/voltage.py
@@ -1,3 +1,5 @@
+from __future__ import division
+
 import nengo
 import numpy as np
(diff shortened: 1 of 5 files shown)


tcstewar:
> maybe we want to reduce or turn off the filtering for voltage plots?

Yup, that makes sense to me -- I'm pretty sure we didn't filter it in 1.4.

hunse:
Okay, removed the filtering. I also made the amount of time viewed changeable in #155, which makes these voltage plots a lot easier to view.

nengo_viz/components/voltage.py
```diff
@@ -0,0 +1,54 @@
...
+
+        for t, datum in zip(trange, data):
+            packet = self.struct.pack(t, *datum)
+            client.write(packet, binary=True)
+
+    def javascript(self):
+        info = dict(uid=self.uid, label=self.label,
+                    n_lines=self.n_neurons, synapse=0)
+        json = self.javascript_config(info)
+        return 'new VIZ.Value(main, sim, %s);' % json
```
tcstewar: Hmm..   This is the first Component that doesn't do a one-to-one mapping between the python class and the javascript class.  I'm not sure that's necessarily a bad thing, but it's worth stopping to consider.

I think I'd prefer making a VIZ.Voltage that's a subclass of VIZ.Value.  That way it could be expanded to have other features, such as having the voltage plots separated vertically, as in 1.4.

nengo_viz/static/viz_value.js
```diff
@@ -17,7 +17,8 @@ VIZ.Value = function(parent, sim, args) {
     this.display_time = args.display_time;
 
     /** for storing the accumulated data */
-    this.data_store = new VIZ.DataStore(this.n_lines, this.sim, 0.01);
+    var synapse = (args.synapse !== null) ? args.synapse : 0.01;
```
tcstewar: synapse can't be null here, since it's always being set to either 0 or 0.01 above.  And if it's not set, it'll be undefined, not null.

nengo_viz/static/viz_value.js
```diff
@@ -41,17 +44,17 @@ VIZ.Value = function(parent, sim, args) {
                     .attr('class', 'graph_text unselectable')
                     .attr('y', args.height - (this.margin_bottom-20))
                     .attr('x', args.width - (this.margin_right + 20));
-        
-    this.axis_time_end = axis_time_end[0][0];  
+
+    this.axis_time_end = axis_time_end[0][0];
```
tcstewar: And more spacing fixes to remove.  Sorry about this!  The spacing stuff will all get cleared up before the summer school.....   :)

nengo_viz/components/voltage.py
hunse: In that case I would make a `VIZ.Voltage` that does the voltages separated vertically, and leave this as-is. And if our requirements here diverge from `VIZ.Value`, then we can always make a separate subclass for that later, too. It just doesn't make sense to me to make a subclass, since the only problem with my approach is that someone could make changes to `VIZ.Value` not realizing it's making changes to the voltage plot as well, and that's still going to be a problem if we use a subclass. I would prefer renaming `VIZ.Value` to be more generic.

tcstewar: Ah, that's a good point.  Some refactoring would be needed here.  In fact, this may be similar to some of the refactoring @jgosmann did in his "all is relative" branch.  So it probably makes sense to leave this as it is and deal with it when we start wanting separate features for the value graph and the voltage graph.

nengo_viz/components/voltage.py
```diff
@@ -0,0 +1,54 @@
...
+        self.struct = struct.Struct('<%df' % (1 + self.n_neurons))
+
+    def add_nengo_objects(self, viz):
+        with viz.model:
+            self.probe = nengo.Probe(self.obj[:self.n_neurons], 'voltage')
+
+    def remove_nengo_objects(self, viz):
+        viz.model.probes.remove(self.probe)
+
+    def format_data(self, t, x):
```
jgosmann: This function doesn't seem to be called?

jgosmann:
Added a commit to stack the voltage traces. I think, this is ready for review.
In the future we should decide whether to keep the different colors for the neurons or go single colored like in the spike raster. Also, the y axis tick labels are not really helpful for this plot. But both of this points are not super quick fixes because the voltage plot currently relies on the value plot.

jgosmann:
I also rebased to master.

tbekolay:
Checked this out, looks pretty neat! I was a bit confused because I thought if I rescaled the y-axis, I would be able to see more neurons' voltages. That's not the case, it always just shows 5 neurons. It probably doesn't matter since it's kind of ridiculous to look at even 5 neurons' voltage traces, but I wonder if we can be more clear that we're only showing voltages for the first 5 neurons? I guess we do this in spike rasters too... Well, in any case, this works and looks good to me ;)

hunse:
At some point we'll add something like the voltage plots in old Nengo, which shows each voltage on a different line like the spike rasters, making it easier to read.

As for the number of neurons, I copied the structure from the spike rasters, which have an option for number of neurons on the server side, but there's no way for the user to actually access it. It would be nice to be able to configure this on the fly on the client side (especially to hide some neurons/select the voltage for only one neuron, etc.), though we also don't want to much info from server to client if it's not being plotted. We'll have to talk about how to achieve this. It would be cool if the user could select individual neurons based on their characteristics, for example select five "on" neurons and five "off" neurons. I'm imagining a popup that has a table of all the neurons in the population and can be sorted by encoder, gain, bias, etc.. The same thing could be used to select neurons for spike rasters or just to provide info about the population. But this is all for down the road.

For now, clarifying that we're just showing five neurons is a good idea.

jgosmann:
I changed the range over which the neurons are stacked. Is that better to indicate that only 5 neurons are shown?

Right now we're somewhat limited in option due to the dependence on the value plot.

hunse:
> Right now we're somewhat limited in option due to the dependence on the value plot.

My idea was to have a voltage plot that is just the same as the value plot (i.e. a normal plot where all traces are on the same axes), and then have a whole nother plot for the stacked voltages.

jgosmann:
I think, without stacking the voltages, it is pretty much unusable for more than one neuron.

celiasmith:
I agree

tbekolay:
I also agree, the stacking is necessary. The range change helps a lot I think! And I agree with @hunse's justification for having it be just 5 neurons, for now. I don't think it's worth making them all the same color right now, as that'd be a somewhat big change. So, I think this branch is ready to go.

Anyone have any issues with this in its current state?

tbekolay:
Okay, I'm going to rebase and merge this.

PR #155, opened 2015-05-20
Made time slider resizable to change time window

hunse:
This allows the user to view a wider or narrower range of time,
and can help make some plots much clearer (e.g. voltage plot).

### DIFF
--- a/nengo_viz/static/viz.js
+++ b/nengo_viz/static/viz.js
@@ -4,6 +4,19 @@ var VIZ = {};
 VIZ.max_zindex = 0;
 
 /**
+ * Helper function to clip a number, keeping it between two values.
+ */
+VIZ.clip = function(x, low, high) {
+    if (x < low) {
+        x = low;
+    }
+    if (x > high) {
+        x = high;
+    }
+    return x;
+}
+
+/**
  * Helper function to set the transform of an element.
  */
 VIZ.set_transform = function(element, x, y) {
--- a/nengo_viz/static/viz_sim_control.js
+++ b/nengo_viz/static/viz_sim_control.js
@@ -209,34 +209,26 @@ VIZ.TimeSlider = function(args) {
             edges: {left: true, right: true, bottom: false, top: false}
         })
         .on('resizemove', function (event) {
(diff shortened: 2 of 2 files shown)


nengo_viz/static/viz.js
```diff
@@ -4,10 +4,21 @@ var VIZ = {};
 VIZ.max_zindex = 0;
 
 /**
+ * Helper function to clip a number, keeping it between two values.
+ */
+VIZ.clip = function(x, low, high) {
+    if (x < low)
```
tbekolay: Should always include the `{ }` even if they're not necessary; it's too easy to add another line to this and think that it's in the body of the if when it's actually not.

nengo_viz/static/viz.js
```diff
@@ -4,10 +4,21 @@ var VIZ = {};
...
+        x = high;
+    return x;
+}
+
+/**
  * Helper function to set the transform of an element.
  */
 VIZ.set_transform = function(element, x, y) {
-    element.style.webkitTransform = 
+    element.style.webkitTransform =
```
tbekolay: Probably best (for the time being) to use `git add -p` to only commit the non-whitespace changes...

nengo_viz/static/viz.js
hunse: I guess if you're used to Python. In C, I'll sometimes even put this all on one line `if (x < low) x = low;`. It might actually be better as one line (than what it currently is), since then you're not tempted to add another line. But having the curly braces is certainly the safest.

nengo_viz/static/viz.js
hunse: Yeah, I had done that before we had the discussion yesterday. When we're looking at merging, I (or whoever is merging) can go through and take out all the whitespace stuff.

hunse: Though I guess if I do it now then we don't have to worry about it later...

hunse:
Okay, made the requested changes (using curly braces and removing whitespace changes).

tcstewar:
Looks good to me.  +1

PR #156, opened 2015-05-20
Removed extra whitespace

hunse:
Removed the extra spaces on the ends of lines in JavaScript and CSS
files. My editor does this automatically, so it's annoying when
I'm making a commit and a bunch of whitespace changes get bundled in.

If we can merge this soon, I think that's best, and then everyone can work from this. Additionally, if everyone can set their editor to do this (if possible), that will help avoid it in the future. Also, is there an easy way to test this?

### DIFF
--- a/nengo_viz/static/main.css
+++ b/nengo_viz/static/main.css
@@ -13,11 +13,11 @@ body {
 .graph {
     background: #eeeeee;
 }
-    
+
 .line {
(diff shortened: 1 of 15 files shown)


tcstewar:
We should probably close this and do a new one...  and make it clear that we should be sticking with the same policy as nengo.

PR #157, opened 2015-05-20
Fixed bug where maxy or miny ignored if zero

hunse:
Since we were "or"-ing together values to determine the defaults,
the default value would be used if the inputs were zero, since
zero evaluates as false. Now checking whether they're `null`.

### DIFF
--- a/nengo_viz/static/viz_value.js
+++ b/nengo_viz/static/viz_value.js
@@ -27,9 +27,7 @@ VIZ.Value = function(parent, sim, args) {
     /** scales for mapping x and y values to pixels */
     this.scale_x = d3.scale.linear();
     this.scale_y = d3.scale.linear();
-    var miny = (args.miny !== null) ? args.miny : -1;
-    var maxy = (args.maxy !== null) ? args.maxy : 1;
-    this.scale_y.domain([miny, maxy]);
+    this.scale_y.domain([args.miny, args.maxy]);
 
     /** spacing between the graph and the outside edges (in pixels) */
     this.margin_top = 30;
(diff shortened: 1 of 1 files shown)


nengo_viz/static/viz_value.js
```diff
@@ -27,8 +27,10 @@ VIZ.Value = function(parent, sim, args) {
     /** scales for mapping x and y values to pixels */
     this.scale_x = d3.scale.linear();
     this.scale_y = d3.scale.linear();
-    this.scale_y.domain([args.miny || -1, args.maxy || 1]);
-    
+    var miny = (args.miny !== null) ? args.miny : -1;
+    var maxy = (args.maxy !== null) ? args.maxy : 1;
```
tcstewar: Should that be "undefined" rather than null?

tcstewar: I'm also tempted to wipe this check out completely....  I don't think there are any cases any more where .miny and .maxy aren't set by the caller.

hunse: Yeah, I was wondering about getting rid of it. Python's `None` seems to get mapped to `null` in JavaScript, which is why I used it.

tcstewar: I'd originally intended that check for situations where someone didn't specify args.minx at all, which would map to ```undefined``` in JavaScript.  But now it's always specified, so probably getting rid of the check would be best (and consistent with the lack of checks elsewhere)

hunse:
Changed this to just remove the defaults as per the discussion.

tcstewar:
Looks great to me.  Thanks!

hunse:
Aaaah, you never squashed it!

tcstewar:
Ah, well, squashing hasn't happened in this repository at all anyway.  Probably something I should learn about and do, though.

hunse:
This one looks worse, though, because it says SQUASHME in big letters in the commit message. :smile:

PR #159, opened 2015-05-20
Changes to increase speed and responsiveness

hunse:
The first commit (the only so far), makes the config file save no more than every 2 seconds, and only after the changes have been sent to the UI. This makes everything considerably more responsive, especially for large models like Spaun.

### DIFF
--- a/nengo_viz/viz.py
+++ b/nengo_viz/viz.py
@@ -105,6 +105,9 @@ class Viz(object):
         self.default_labels = self.name_finder.known_name
 
         self.config = self.load_config()
+        self.config_save_needed = False
+        self.config_save_time = None   # time of last config file save
+        self.config_save_period = 2.0  # minimum time between saves
 
         self.lock = threading.Lock()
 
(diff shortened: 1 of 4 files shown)


hunse:
Okay, I've added a cleaner way of closing threads. It's pretty similar to what @tcstewar had originally, but all the threads are daemons now so that if they don't close nicely, they'll just die with the main thread. Firefox seems to exit nicely with no zombie threads, but Chrom(e|ium) spawns a bunch of sockets with no requests when loading a new page (i.e. the shutdown page), and these become zombies since they're effectively stuck waiting for a request. I don't feel bad about closing them abruptly :smile:

hunse:
From what I can tell so far, this works better than #162 and should hopefully fix @arvoelke's problem with not being able to use the same port (but I haven't tried this in the testing framework yet, so we'll see).

hunse:
I think this is ready to go. There are some other speed improvements to the name and parent finding systems that I've been working on, but they're not ready yet so I'll do them as part of another PR.

nengo_viz/viz.py
```diff
@@ -163,9 +166,25 @@ def load_config(self):
...
+
+        self.lock.acquire()
+        self.config_save_time = now_time
+        self.config_save_needed = False
         with open(self.filename + '.cfg', 'w') as f:
             f.write(self.config.dumps(uids=self.default_labels))
+        self.lock.release()
+
+    def modified_config(self):
+        if not self.config_save_needed:
```
tbekolay: Probably better to just set it to true? That takes as long as the check to see if it's already true.

tbekolay:
Aside from my inline comment, this LGTM!

hunse:
Arg, GitHub's doing it's weird commit ordering again.

Anyway, fixed that comment and also use `with` for the lock.

tbekolay:
LGTM! The `with` for the lock is great too.

hunse:
Cool. Squashed those in.

tcstewar:
Looks great to me too!  Thank you!

PR #161, opened 2015-05-20
Detect code changes

tcstewar:
If you open up and edit the ```__file__``` using a separate text editor, the display should update whenever you save it.

### DIFF
--- a/nengo_viz/components/component.py
+++ b/nengo_viz/components/component.py
@@ -4,7 +4,6 @@ class Component(object):
     def __init__(self, viz, config, uid):
         self.config = config
         self.uid = uid
-        self.new_component = None
 
     def update_client(self, client):
         pass
--- a/nengo_viz/components/component.py
+++ b/nengo_viz/components/component.py
@@ -4,6 +4,8 @@ class Component(object):
     def __init__(self, viz, config, uid):
         self.config = config
         self.uid = uid
+        self.new_component = None
+
     def update_client(self, client):
         pass
     def message(self, msg):
--- a/nengo_viz/components/netgraph.py
+++ b/nengo_viz/components/netgraph.py
@@ -1,7 +1,7 @@
 import time
 import struct
-import threading
 import os
+import traceback
 
 import numpy as np
 import nengo
@@ -40,8 +40,13 @@ class NetGraph(Component):
         try:
             exec(code, locals)
         except:
-            pass
-        model = locals['model']
+            traceback.print_exc()
+        try:
+            model = locals['model']
+        except:
+            traceback.print_exc()
+            return
+
         locals['nengo_viz'] = nengo_viz
         name_finder = nengo_viz.NameFinder(locals, model)
 
@@ -112,6 +117,7 @@ class NetGraph(Component):
         self.viz.config = self.viz.viz.config
         self.config = self.viz.viz.config
         self.viz.viz.uid_prefix_counter = {}
+        self.layout = nengo_viz.layout.Layout(model)
 
         components = []
         for c in self.viz.components[:2]:
--- a/nengo_viz/components/netgraph.py
+++ b/nengo_viz/components/netgraph.py
@@ -79,6 +79,15 @@ class NetGraph(Component):
                     old_post = old_item.post_obj
                     new_pre = new_item.pre_obj
                     new_post = new_item.post_obj
+                    if isinstance(old_pre, nengo.ensemble.Neurons):
+                        old_pre = old_pre.ensemble
+                    if isinstance(old_post, nengo.ensemble.Neurons):
+                        old_post = old_post.ensemble
+                    if isinstance(new_pre, nengo.ensemble.Neurons):
+                        new_pre = new_pre.ensemble
+                    if isinstance(new_post, nengo.ensemble.Neurons):
+                        new_post = new_post.ensemble
+
                     old_pre = self.viz.viz.get_uid(old_pre)
                     old_post = self.viz.viz.get_uid(old_post)
                     new_pre = self.viz.viz.get_uid(new_pre,
@@ -108,10 +117,19 @@ class NetGraph(Component):
         self.viz.config = self.viz.viz.config
         self.config = self.viz.viz.config
         self.viz.viz.uid_prefix_counter = {}
-        self.viz.changed = True
 
-        #for c in self.viz.components:
-        #    c.new_component = c.template.create(self.viz)
+        components = []
+        for c in self.viz.components[:2]:
+            components.append(c)
+            locals[c.uid] = c.template
+        self.viz.components = components
+        for template in self.viz.viz.find_templates():
+            if not isinstance(template,
+                              (nengo_viz.components.SimControlTemplate,
+                               nengo_viz.components.NetGraphTemplate)):
+                self.viz.add_template(template)
+
+        self.viz.changed = True
 
 
     def get_parents(self, uid, default_labels=None):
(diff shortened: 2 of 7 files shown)


tcstewar:
@tbekolay or @hunse if you could take a look at this, that'd be great!  (you two have been doing a great job of other code reviews, so I thought I'd ask you first...  :)  )

nengo_viz/components/component.py
```diff
@@ -4,6 +4,8 @@ class Component(object):
     def __init__(self, viz, config, uid):
         self.config = config
         self.uid = uid
+        self.new_component = None
```
tbekolay: What is this for? I grepped and couldn't find any other references to `new_component`...

nengo_viz/components/netgraph.py
```diff
@@ -21,19 +23,128 @@ def __init__(self, viz, config, uid):
...
         self.parents = {}
         self.networks_to_search = [self.viz.model]
+        self.initialized_pan_and_zoom = False
+        self.last_modify_time = os.path.getmtime(self.viz.viz.filename)
 
-    def get_parents(self, uid):
+    def check_for_reload(self):
+        t = os.path.getmtime(self.viz.viz.filename)
+
+        if self.last_modify_time is None or self.last_modify_time < t:
```
tbekolay: Are there situations where `last_modify_time` could be `None`?

nengo_viz/components/netgraph.py
```diff
@@ -21,19 +23,128 @@ def __init__(self, viz, config, uid):
...
+            traceback.print_exc()
+            return
+
+        locals['nengo_viz'] = nengo_viz
+        name_finder = nengo_viz.NameFinder(locals, model)
+
+        self.networks_to_search = [model]
+        self.parents = {}
+
+        for uid, old_item in self.uids.items():
```
tbekolay: Might be a bit faster to use `iteritems` (or, for Python 3 compatibility, `nengo.utils.compat.iteritems`).

nengo_viz/components/netgraph.py
```diff
@@ -21,19 +23,128 @@ def __init__(self, viz, config, uid):
...
+                        new_post = new_post.ensemble
+
+                    old_pre = self.viz.viz.get_uid(old_pre)
+                    old_post = self.viz.viz.get_uid(old_post)
+                    new_pre = self.viz.viz.get_uid(new_pre,
+                            default_labels=name_finder.known_name)
+                    new_post = self.viz.viz.get_uid(new_post,
+                            default_labels=name_finder.known_name)
+
+                    if new_pre != old_pre or new_post != old_post:
```
tbekolay: Are you checking for the items being the same object, or something else? I don't think we've implemented `__eq__` methods, so this is probably equivalent to `new_pre is not old_pre or new_post is not old_post`. It's better to explicitly do that if you're checking for the objects being exactly the same instance, otherwise if we do implement `__eq__` methods this might cause a very difficult to find bug.

nengo_viz/static/viz_netgraph.js
```diff
@@ -162,6 +162,21 @@ VIZ.NetGraph.prototype.on_message = function(event) {
...
+        var item = this.svg_objects[data.uid];
+        item.set_label(data.name);    
+    } else if (data.type === 'remove') {
+        var item = this.svg_objects[data.uid];
+        if (item === undefined) {
+            item = this.svg_conns[data.uid];
+        }
+        item.remove();    
+    } else if (data.type === 'reconnect') {
+        console.log(['reconnect', data]);
```
tbekolay: Should this still be in here?

nengo_viz/components/netgraph.py
```diff
@@ -42,12 +153,16 @@ def get_parents(self, uid):
         return parents
 
     def update_client(self, client):
+        self.check_for_reload()
```
tbekolay: How often is `update_client` called? Going out to the filesystem to determine the `mtime` is fast in I/O terms, but it's still I/O. If this is called, e.g., 50+ times per second we might only want to check for reload sometimes. If it's not that much, then probably it's fine. Long term, we should probably listen to filesystem events instead of polling (made an issue for that; #182).

tbekolay:
Made some comments inline; other than those it looks good to me! Note that this can be very easily rebased (did it no conflicts on my machine), so it should probably be done to avoid the duplicate commits.

nengo_viz/components/netgraph.py
tcstewar: Nope, this is checking whether the objects are the "same" in the sense of they have the same uid.  They won't be the same object, since when the code changes, we revaluate the script, which creates all-new objects.  So we use the uid to figure out object identity

This gets complicated for Connections, since they don't really have a sensible uid.  (something like ```model.connections[3]``` is about as close as we could get to a uid, but then that breaks horribly when someone adds or removes a connection).  So I'm using the uid of the pre and the post of the connection instead.

nengo_viz/components/netgraph.py
tcstewar: Oops, nope, there aren't....  good catch!

nengo_viz/components/netgraph.py
tcstewar: It's probably close to 50 times per second.  I think I'll switch this so it only calls check_for_reload every half-second or so, which should be sufficient.

nengo_viz/static/viz_netgraph.js
tcstewar: oops, nope, it shouldn't...  :)

nengo_viz/components/netgraph.py
studywolf: just flagging this as possibly a good thing to have in the config menu, some users might prefer also just having a button saying 'update' or whatever...by 'just flagging this' i'm not sure what i mean. if there's a way to flag it that'd be sweet.

tcstewar: That's a good point.  In ```nengo_gui``` this behaviour got turned on by a command-line option (and wasn't on by default).  Maybe we should do the same here?

nengo_viz/components/component.py
tcstewar: Oops...  that was from my first attempt at this....  will remove.

nengo_viz/components/netgraph.py
tcstewar: Nice.  I should remind myself of that one....

nengo_viz/components/netgraph.py
tbekolay: Makes sense!

tbekolay:
With the new commits, this LGTM. Do you want me to merge @tcstewar?

tcstewar:
> With the new commits, this LGTM. Do you want me to merge @tcstewar?

I'll do it...  going to try out the ```-f``` I asked you about.  :)

tbekolay:
Nice :rainbow:

nengo_viz/components/netgraph.py
hunse: There could be more than one connection with the same `pre` and `post` though, no?

nengo_viz/viz.py
```diff
@@ -67,7 +67,11 @@ def runner(self):
             if self.sim is None:
                 time.sleep(0.01)
             else:
-                self.sim.step()
+                try:
+                    self.sim.step()
+                except AttributeError:
```
hunse: Why are we expecting this? Maybe a comment is in order here.

nengo_viz/components/netgraph.py
hunse: Also, at some point (probably not now), we might want to consider an algorithm that tries to match all objects simultaneously, since it's always going to be a one-to-one mapping with additional non-matching items in either the old model or the new model or both.

hunse:
Hah, I just realized you guys had merged already...

nengo_viz/components/netgraph.py
tcstewar: Good points.  In fact, we should decide in general what should happen with a Connection with the same pre and post...  at the moment that's two visually identical arrows in the NetGraph.  I'll set up a separate Issue to think about that...

PR #162, opened 2015-05-20
Server shutdown command

arvoelke:
Tests need to be able to shutdown the server in between different models.

Current solution is a little hacky, but the best we could come up with. It requires an additional request after the /shutdown in order to flush out the polling connections. Ports also need to be changed after each shutdown, because resources are not properly cleaned up for some unknown reason.

### DIFF
--- a/nengo_viz/swi.py
+++ b/nengo_viz/swi.py
@@ -74,6 +74,7 @@ class SimpleWebInterface(BaseHTTPServer.BaseHTTPRequestHandler):
     server_version = 'SimpleWebInterface/2.0'
     serve_files = []
     serve_dirs = []
+    shutdown_flag = False
 
     pending_headers = None
     testing_user = None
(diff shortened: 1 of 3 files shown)


hunse:
Aaah, I've already done this in #159.

hunse:
Wait, maybe our things are a bit different. Mine is to allow the user to interrupt the server and shut it down cleanly.

hunse:
I just added a harsh way to do this in #159. I made all the threads daemons, so they have to exit when the main thread does.

tcstewar:
Do we still want any features from this, or is this redundant with the things we've already got merged into master?

nengo_viz/swi.py
```diff
@@ -289,9 +291,9 @@ def handle_request(self, args, db):
             self.send_response(200)
             self.send_header('Content-type', 'text/html')
             self.end_headers()
-            self.wfil.write(("<html><body>Invalid request:"
-                             "<pre>args=%s</pre><pre>db=%s</pre>"
-                             "</body></html>" % (args, db)).encode("utf-8"))
+            self.wfile.write(("<html><body>Invalid request:"
```
hunse: This was a bug? Should make sure this gets fixed in `master`.

tcstewar: Good point...  I just made a separate PR for that: #192

nengo_viz/viz.py
```diff
@@ -180,8 +182,16 @@ def get_uid(self, obj):
...
     def create_sim(self):
         """Create a new Simulator with this configuration"""
-        return VizSim(self)
+        vizsim = VizSim(self)
+        self.vizsims.append(vizsim)
+        return vizsim
+
+    def cleanup(self):
+        for vizsim in self.vizsims:
+            vizsim.finish()
```
hunse: We don't have anything like this in `master`, but it seems unnecessary because `VizSim.finish()` is called by `SimControl.finish()` which in turn is called at the end of `server.ws_viz_component`. The `VizSim` threads seem to exit cleanly on my machine in any case, and if they don't exit cleanly on other machines they will be reported as zombie threads.

hunse:
I think this is all redundant, other than the bug now fixed in #192, so I'll close.

PR #163, opened 2015-05-20
Switched to using viewport for pan/scale

tcstewar:
Started creating a master system for pan/zoom, called viewport

This uses the same coordinate system as the netgraph.  At the moment, only the Components are using the viewport, but it should be straightforward to switch netgraph over too.


tcstewar:
Oops I broke this horribly.  I'm making a new PR.

PR #160, opened 2015-05-20
Fixed dropped websocket messages

tcstewar:
Stops the websocket system from ignoring messages to the server

PR #165, opened 2015-05-21
Sliders move and reset

tcstewar:
Sliders which have a callable Node now move automatically.  You can also reset them to get back to the original function.

### DIFF
--- a/nengo_viz/components/slider.py
+++ b/nengo_viz/components/slider.py
@@ -1,5 +1,8 @@
+import collections
+
 import numpy as np
 import struct
+
 
 from nengo_viz.components.component import Component, Template
 
@@ -13,7 +16,7 @@ class Slider(Component):
         self.label = viz.viz.get_label(node)
         self.start_value = np.zeros(node.size_out)
         self.struct = struct.Struct('<%df' % (1 + node.size_out))
-        self.data = []
+        self.data = collections.deque()
         if not callable(self.base_output):
             self.start_value[:] = self.base_output
 
@@ -44,8 +47,8 @@ class Slider(Component):
 
     def update_client(self, client):
         while len(self.data) > 0:
-            data = self.data.pop(0)
-            client.write(data, binary=True)
+            item = self.data.popleft()
+            client.write(item, binary=True)
 
     def message(self, msg):
         index, value = msg.split(',')
--- a/nengo_viz/components/slider.py
+++ b/nengo_viz/components/slider.py
@@ -1,4 +1,5 @@
 import numpy as np
+import struct
 
 from nengo_viz.components.component import Component, Template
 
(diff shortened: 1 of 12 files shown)


nengo_viz/components/slider.py
```diff
@@ -36,13 +40,21 @@ def javascript(self):
                     label=self.label,
                     start_value=[float(x) for x in self.start_value])
         json = self.javascript_config(info)
-        return 'new VIZ.Slider(main, %s);' % json
+        return 'new VIZ.Slider(main, sim, %s);' % json
+
+    def update_client(self, client):
+        while len(self.data) > 0:
+            data = self.data.pop(0)
```
tbekolay: As noted in #148, this is O(n^2). I dunno if it really matters here, but regardless, should we do the same optimization as in that PR?

nengo_viz/static/viz_slider.js
```diff
@@ -152,8 +164,8 @@ VIZ.Slider.prototype.constructor = VIZ.Slider;
 
 VIZ.Slider.prototype.set_value = function(slider_index, value, immediate) {
     console.assert(typeof slider_index == 'number');
-    console.assert(typeof value == 'number');
-    console.assert(typeof immediate == 'boolean');
+    console.assert(typeof value == 'number', value);
+    //console.assert(typeof immediate == 'boolean');
```
tbekolay: Should this still be commented out?

nengo_viz/static/viz.js
```diff
@@ -378,6 +378,12 @@ VIZ.DataStore.prototype.get_shown_data = function() {
     return shown;
 }
 
+VIZ.DataStore.prototype.is_at_end = function() {
+    var ts = this.sim.time_slider;
+    return (ts.last_time < ts.first_shown_time + ts.shown_time + 0.01);
```
tbekolay: Hmm... I'm a bit confused by this logic. I'm guessing that 0.01 is a fudge factor for float comparison; in that case, shouldn't it be something like
```javascript
ts.last_time >= ts.first_shown_time + ts.shown_time - 0.01
```
?

tbekolay:
Made some comments inline with questions etc, but can confirm that this does indeed cause sliders to move! Pretty sweet.

nengo_viz/static/viz_slider.js
tcstewar: Argh, nope, I need to clear that up -- I added a third option, so it's not a boolean any more....  I'll add something to make that clearer.

nengo_viz/components/slider.py
tcstewar: I still think there's the thread issue.  But here's an alternative:
```python
        count = len(self.data)
        if count > 0:
            for d in self.data[:count]:
                client.write(d, binary=True)
            del self.data[:count]
```

nengo_viz/static/viz.js
tcstewar: Yup, that's for float fudge factor, but I made it way too large...   It does have to be a ```+```, though (with a - sign, the function can never return true).  +0.0000001 seems to work, though.

nengo_viz/static/viz_slider.js
tcstewar: Actually, that whole argument is going to disappear and get reorganized when #160 gets brought in.  So I'm tempted to just put a TODO in here.

tbekolay: Yeah, or just remove the assert entirely, that's fine too.

PR #169, opened 2015-05-21
All is relative

jgosmann:
Eliminates all absolute sizing from components and uses relative scaling/padding/margin based on the font-size. Also, makes the text zoom with everything else.

While doing this, I refactored the components to have less code duplication.

This is based on some intermediate stage of `graphics_overhaul`.

### DIFF
--- a/nengo_viz/components/value.py
+++ b/nengo_viz/components/value.py
@@ -39,4 +39,4 @@ class Value(Component):
 
 class ValueTemplate(Template):
     cls = Value
-    config_params = dict(maxy=1, miny=-1, **Template.default_params)
+    config_params = dict(max_value=1, min_value=-1, **Template.default_params)
--- a/nengo_viz/static/viz_2d_axes.js
+++ b/nengo_viz/static/viz_2d_axes.js
@@ -1,0 +1,88 @@
+/**
+ * Basic 2d axes set.
+ * @constructor
+ *
+ * @param {float} args.width - the width of the axes (in pixels)
+ * @param {float} args.height - the height of the axes (in pixels) 
+ * @param {float} args.min_value - minimum value on y-axis
+ * @param {float} args.max_value - maximum value on y-axis
+ */
+ 
+VIZ.Axes2D = function(parent, args) {
+    var self = this;
+
+    /** draw the plot as an SVG */
+    this.svg = d3.select(parent).append('svg')
+        .attr('width', '100%')
+        .attr('height', '100%');
+
+    /** scales for mapping x and y values to pixels */
+    this.scale_x = d3.scale.linear();
+    this.scale_y = d3.scale.linear();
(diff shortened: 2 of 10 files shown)


jgosmann:
This is just a first pass on refactoring the components. There are probably still a number of improvements possible.

nengo_viz/static/slidercontrol.js
```diff
@@ -0,0 +1,209 @@
+VIZ.SliderControl = function(min, max) {
```
tcstewar: Should this file be ```viz_slidercontrol.js``` for consistency?

tcstewar:
the file ```.DS_Store``` needs to get removed, and then there's the whole rebasing to do, but other than that I like this refactoring.

jgosmann:
The `.DS_Store` file was committed accidentally in the `graphics_overhaul` branch which is partially part of this branch. I would suggest to merge that branch first and then rebase this one. But I could also rebase this one first (I don't think, I depend on anything particular from that branch.)

tcstewar:
> I would suggest to merge that branch first and then rebase this one.

Is that branch ready for merging?  I don't see a PR for it.....

nengo_viz/static/slidercontrol.js
jgosmann: For now this would make sense. But in general I would favor dropping the `viz` prefix in filenames altogether. It is not really informative because everything is prefixed with `viz` and gives only clutter when listing the files. If at all needed I would prefer to use a directory instead of prefixing the files.

It might make sense to also provide some more structure with some directories given the growing number of files, for example move all components to their own subdirectory.

tcstewar: I totally agree with dropping the ```viz_*``` prefix on the files.  We'll put it on the list of things to do once it's all merged... 

And, in fact, if you want to leave this one the way it is (as kind of a marker that we should fix this in the future), I'm good with that too.

jgosmann:
just made one: #180

jgosmann:
Rebased to `master` after `graphics_overhaul` has been merged. I might have broken some slider stuff in resolving the merge conflicts. Will test that tomorrow.

jgosmann:
Fixed the sliders and cleaned up the history. (Commits might be shown in the wrong order on Github, because they are shown by time stamp and not by parent.)

Remaining problems:
* [ ] The size of the sliders is incorrectly initialized on load.
* [ ] The font scaling is incorrectly initialized on load.

jgosmann:
One more to-do: Move the CSS styling of the slider from the JS to a CSS file. I don't think this prevents a merge because it is only about code organization.

tcstewar:
> I don't think this prevents a merge because it is only about code organization.

Looking great!  Now I just have to decide on which order to merge these large things into master in....  That'll be my job for tomorrow after a good night's sleep...  :)

PR #171, opened 2015-05-21
Update interact to current development version.

jgosmann:
This fixes some resize bugs, for example it allows to resize in
x and y direction when clicking the bottom corners which was
sometime not possible with the current stable version.


tcstewar:
Nice.  Looks good to me, and fixes those corner problems

Issue #147, opened 2015-05-19
Graphical overhaul


jgosmann:
* [x] Change cursor on label to default pointer
* [x] Make sure cursor changes work in Firefox
* [x] Cursor state lost after dragging
* [x] Different cursor for resizing and dragging background
* [x] Larger resize margin (esp top left corner; add invisible object?)
* [x] Find a good looking border width
- [x] Standardize text size and font (and don't use pixel values)
- [x] Scale fonts with objects/global zoom
- [x] Nodes: squares
- [x] Networks: rounded rectangles (radius should be percent-based so resizing changes rounding)
- [x] Ensembles: six circle
- [x] Better value plot colors, and other color stuff (Eric)
* [x] Place arrows on lines based on intersection with items
* [x] The labels should go below the items
* [x] Zooming on Macbook track pad.
* [x] Right click for menus
* [x] Left click for... select? info? other suggestions?
* [x] Left align menus
* [x] Double click networks to expand/collapse
* [x] Fix default slider width
* [x] Add borders to ensembles
* [x] Change menus to dropdown menus (instead of button groups)
* [x] Change modal dialogs to bootstrap

PR #173, opened 2015-05-21
Fix cursors in Firefox.

jgosmann:
Addresses #147.


tcstewar:
Also looks good to me.

PR #175, opened 2015-05-21
Style time bar

jgosmann:
Font in time line is untouched, but was changed in `graphics_overhaul`.


tcstewar:
Nifty.  This looks good to me.

PR #176, opened 2015-05-21
Switched to using viewport for pan/scale

tcstewar:
Newly rebased by cherry-picking


tcstewar:
Actually, this is already merged into the undo_stack branch, so let's deal with this as part of that branch.

Issue #170, opened 2015-05-21
Allow semantic pointer plot to plot information for non-default module input and outputs


xchoo:
Currently, the semantic pointer plot is hardcoded to probe only the default input and outputs. Being able to plot information from non-default input and outputs will be helpful for models using custom modules (e.g. Spaun).

tbekolay:
Done?

PR #179, opened 2015-05-21
Draft version of Pointer graphs supporting non-default outputs

xchoo:
Netgraph items now get output list from Module object and passes it to client to use (client only uses first item for now -- waiting for the graphical overhaul branch to be merged before handling multiple possible outputs).
Client passes output name back to server when generating pointer graphs.

### DIFF
--- a/nengo_viz/components/netgraph.py
+++ b/nengo_viz/components/netgraph.py
@@ -103,7 +103,7 @@ class NetGraph(Component):
         self.config[obj].size = width, height
         self.viz.viz.save_config()
 
-    def act_create_graph(self, uid, type, x, y, width, height):
+    def act_create_graph(self, uid, type, x, y, width, height, **args):
         cls = getattr(nengo_viz.components, type + 'Template')
         obj = self.uids[uid]
         template = cls(obj, **args)
--- a/nengo_viz/components/netgraph.py
+++ b/nengo_viz/components/netgraph.py
@@ -1,9 +1,12 @@
+import time
+import struct
+
+import numpy as np
 import nengo
 import json
 
 from nengo_viz.components.component import Component, Template
 import nengo_viz.layout
-
 
 class NetGraph(Component):
     configs = {}
@@ -103,10 +106,10 @@ class NetGraph(Component):
         self.config[obj].size = width, height
         self.viz.viz.save_config()
 
-    def act_create_graph(self, uid, type, x, y, width, height, **args):
+    def act_create_graph(self, uid, type, x, y, width, height, **kwargs):
         cls = getattr(nengo_viz.components, type + 'Template')
         obj = self.uids[uid]
-        template = cls(obj, **args)
+        template = cls(obj, **kwargs)
         self.viz.viz.generate_uid(template, prefix='_viz_')
         self.config[template].x = x
         self.config[template].y = y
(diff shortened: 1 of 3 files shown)


nengo_viz/components/netgraph.py
```diff
@@ -109,7 +106,7 @@ def act_pos_size(self, uid, x, y, width, height):
     def act_create_graph(self, uid, type, x, y, width, height):
         cls = getattr(nengo_viz.components, type + 'Template')
         obj = self.uids[uid]
-        template = cls(obj)
+        template = cls(obj, **args)
```
tcstewar: This ```args``` breaks things completely....  did a commit get missed here?

nengo_viz/static/viz_netgraph_item.js
```diff
@@ -36,7 +36,7 @@ VIZ.NetGraphItem = function(ng, info) {
     this.minHeight = 5;
 
     this.expanded = false;
-    
+
```
tcstewar: It will greatly simplify the merge if we don't include all these whitespace changes.  @tbekolay walked me through a process for trying to get rid of these...

nengo_viz/components/netgraph.py
xchoo: Fixed!

nengo_viz/static/viz_netgraph_item.js
xchoo: Yeah. My editor automatically does this... (removes extra whitespaces at the end of lines).

tcstewar: Even if the editor does that, don't commit those changes!  :)

xchoo: I can select lines to commit? There is a change in this file (further down, lines 219, 226).

nengo_viz/components/pointer.py
```diff
@@ -1,26 +1,27 @@
...
+from nengo.spa.module import Module
 import numpy as np
-import struct
 
 from nengo_viz.components.component import Component, Template
 
 
 class Pointer(Component):
-    def __init__(self, viz, config, uid, obj):
+    def __init__(self, viz, config, uid, obj, **args):
```
tcstewar: Hmm...  this doesn't quite work as it is, since the args don't get saved in the config.  I think this is more a bug in the Template system, though (I think the correct fix is a change to the ```Template.code_python``` function).  So this is probably fine for now.

nengo_viz/static/viz_netgraph_item.js
tcstewar: Yup, that's what ```git add -p filename.py``` is for.  Quite handy.

xchoo: I have to figure out how to do that with tortoisegit. haha

tcstewar: I took a stab at doing this manually the way @tbekolay showed me...  the result is in

https://github.com/nengo/nengo_viz/commits/non_default_pointers_test

It seems to work....  here are the steps I did:

- ```git checkout non_default_pointers```
- ```git rebase -i HEAD~2``` (do an interactive rebase of the last 2 commits in this repository)
- change the text ```pick``` to ```edit``` beside the commit we want to remove things from (this tells the rebase to stop at that point and let us manually do things)
- ```git reset HEAD^``` (undo the commit that we are stopped at)
- ```git add nengo_viz/components/netgraph.py``` (this file is fine)
- ```git add nengo_viz/components/pointer.py``` (so is this one)
- ```git add -p nengo_viz/static/viz_netgraph_item.js``` (we only want part of this file)
- it will ask you what to do with each part of the file.  say ```n``` (no) to all the whitespace changes and ```y``` to the actual changes
- ```git commit``` (you will have to copy in your original commit message)
- ```git checkout .``` (throw away the remaining changes)
- ```git rebase --continue``` (finish the rebase)

drasmuss: In tortoisegit you can double click on the file you're committing, which will open the diff and then you can edit things line by line.  I think that's the easiest way to do it.

tcstewar:
Once the whitespace is removed, I'm good for this one!

nengo_viz/components/pointer.py
hunse: It's traditional to call this `kwargs` instead of `args`.

nengo_viz/static/viz_netgraph_item.js
hunse: `git diff -w` does a diff ignoring whitespace changes, which is useful once you've added all the non-whitespace changes to make sure you've got everything. Unfortunately, there doesn't seem to be anything like `git add -pw`.

xchoo:
Addresses issue #170.

PR #180, opened 2015-05-22
Graphics overhaul

jgosmann:
I think, this should be ready to merge. Am I correct @celiasmith ?


tcstewar:
Is anyone else finding that if you're zoomed in on a small ensemble (usually an ensemble inside a network, like in net.py), you can often get it so that the ensemble can be resized but not moved?  I think this has to do with the global fixed width on the resize area.

nengo_viz/static/viz_netgraph.js
```diff
@@ -104,6 +131,7 @@ VIZ.NetGraph = function(parent, args) {
             self.offsetY = (self.offsetY + yy) / scale - yy;
             
             self.scale = scale * self.scale;
+            interact.margin(30*self.scale);
```
tcstewar: Hmm, this means that if you're zoomed in, the resizable margin changes.  This means that if an object is less than a certain absolute size, you can *never* drag it, even if you zoom in.

tbekolay: I think this is why @celiasmith wanted to make the objects movable by dragging the label instead? When an object is super small, it's kind of impossible to allow two separate behaviors, moving and resizing. If you allow moving, then you can never resize. At least by making resizing the only possibility, you can resize, move, then resize again.

tcstewar: That's true if "super small" means "super small on the screen", and I'd agree that in that case being able to drag using the text is important.  However, here, the problem is that if the object is "super small" in terms of size in the model, but you've zoomed in such that it takes up 50% of your screen, then it's *still* not moveable.  

My feeling is that, as far as the user interface goes, the width of the "resize" area should be fixed in terms of pixels, and not change depending on your global zoom level.  That way if something is small in the absolute sense, but you're zoomed in near it, then you can still move it as expected.  Otherwise, with *any* mildly complicated model, you're never going to be able to move anything by dragging on the item, you'll always have to move it via the text (or some other way).

celiasmith: I'm more than happy to take that out. It was an experiment.  The downside of removing it is that corner drags disappear for scaled up ensembles, but that might be less annoying.

tcstewar: > I'm more than happy to take that out. It was an experiment.  The downside of removing it is that corner drags disappear for scaled up ensembles, but that might be less annoying.

:)  Yay experiments.  :)  I'd also note that for this one, even with interact.margin changing based on global zoom, we also lose corner drags for Ensembles that are just resized to be large.  However, since Ensembles have a fixed aspect ratio anyway, we actually never need a corner drag for them anyway.  So I think switching this to a fixed value makes the most sense.  

The only place I'm a bit worried about it for is the Networks, since corner-drag is useful for them, and this same problem will happen.  I think I'll use that as an argument for rethinking the rounded corner rescaling, but that's a separate experiment and PR. :)

nengo_viz/static/viz_netgraph.js
```diff
@@ -31,7 +33,11 @@ VIZ.NetGraph = function(parent, args) {
     this.svg.id = 'netgraph';
     this.svg.style.height = 'calc(100% - 80px)';
     this.svg.style.position = 'fixed';    
-
+        
+    interact(this.svg).styleCursor(false);
+    //interact(this.svg).preventDefault(true);
+    //interact.margin(30*VIZ.scale.step_size*this.scale);
```
tcstewar: should these commented-out lines be removed?

nengo_viz/static/viz_netgraph.js
jgosmann: Not sure, whether I agree, that we do not need a corner drag for ensembles. Even though the aspect ratio is fixed, the behaviour for a corner drag is slightly different from a side drag.

tcstewar: I agree that it would be nice to have corner drag on ensembles, and it would be slightly different than the edge drag.  However, the current system does not allow that unless the ensemble happens to be about the right size.  If the Ensemble has been shrunk down (via edge drag) smaller than a certain size, then it is impossible to do anything other than a top-left corner drag, because the interact.margin is wider than the object itself.  If the Ensemble has been expanded (via edge drag) larger than a certain size, then you can't do any corner drags, and indeed even edge-drag becomes difficult because the interact.margin can become smaller than the distance from the visible circles to the edge of the invisible circle around the ensemble.  This is because the interact.margin is fixed to the global zoom level, rather than anything to do with the size of the object itself.  This makes the current scaled interact.margin system not work in lots of situations, and is why I think it needs to be removed before merging this PR.  But we should definite look into improving it in future PRs.  I just don't want this PR being blocked on this issue.  :)

tcstewar:
Looking good.   One more small thing:  The ```nengo_viz/static/lib/.DS_Store``` file should be removed.

Other than that, I think it's time to try for a rebase to master.  Can one of you two who wrote that try it?  If not, I'll take a stab at it this afternoon....

jgosmann:
I can do that after lunch :banana:  or so

jgosmann:
Rebased to master, squashed `fixup!` commits, and slightly cleaned up the history. :cake:

PR #183, opened 2015-05-23
Undo stack 2

bjkomer:
Implementing an undo and redo stack for a variety of commands. So far it should support:
-Moving/Resizing netgraph items
-Moving/Resizing components
-Creating/Deleting components
-Expanding/Collapsing networks
-Auto-Layout
-All currently implemented component properties accessible through the click menu (setting ranges, hiding labels, etc)

There are a couple minor bugs that still need to be fixed, and some hacky bits that could be cleaned up, but it is mostly ready to go.


nengo_viz/static/viz_slider.js
```diff
@@ -393,6 +394,10 @@ VIZ.Slider.prototype.set_range = function() {
     }
 };
 
+VIZ.Slider.prototype.update_range = function(min, max) {
+    this.scale.domain([min, max]);
+}
+
 VIZ.Slider.prototype.layout_info = function () {
     var info = VIZ.Component.prototype.layout_info.call(this);
     info.min_value = this.scale.domain()[1];
```
hunse: Is this the cause of the [backwards fishiness](https://github.com/nengo/nengo_viz/pull/183/files#diff-7a1005f176b4a21d3b051a0e8dca1fd6R409)? Seems like the min value should be in `[0]` and the max in `[1]`.

nengo_viz/components/netgraph.py
```diff
@@ -143,7 +177,7 @@ def act_feedforward_layout(self, uid):
                                         pos=self.config[obj].pos,
                                         size=self.config[obj].size))
         self.config[network].has_layout = True
-        self.viz.viz.save_config()
+        self.viz.viz.save_config()"""
```
hunse: All this dead code is to be removed?

hunse:
Perhaps we could squash the commits before review, as well. I'm not sure what kind of structure we want, but I'd be happy having all the undo stack commits as one big one, unless there's some logical way to split it up (can we do joint authorship on commits?). Commits like 5c9950c that are not directly related to the undo stack should stay separate.

tcstewar:
> Perhaps we could squash the commits before review, as well.

+1 on squashing, I think....  probably something to consider for the other large branches too...

nengo_viz/components/netgraph.py
s72sue: Yes, this code which is commented out in netgraph.py should be removed. We just kept it there during development for experimenting with different designs. 

Most of this functionality has been moved to action.py.

tbekolay:
There's no way to really do joint authorship on commits, but you can note in the commit message that this work was co-authored by someone. I've done it on a few commits in Nengo, but not enough to have a well-defined (searchable) template. Some projects probably have it, though.

This reminds me, since we're talking about authorship and stuff, we should have everyone sign the `CONTRIBUTORS.md` file.

nengo_viz/static/viz_slider.js
bjkomer: Yea, the prompt for reading in a new range is backwards, but I think it still works because the slider drag logic is backwards too. It should be an easy fix, I just didn't want to touch too much unrelated stuff in this branch

bjkomer:
Another +1 on squashing. There are a couple commits in here that were duplicated due to weird rebase stuff, so it would be nice if everything got smashed into a couple big commits so that the history is more sensible.

s72sue:
I agree with having a couple big commits to make the history sensible.

Also, I din't make the changes in 5c9950c
<https://github.com/nengo/nengo_viz/commit/5c9950c33fd656cad7a5c8425282aba8fc923391>
I found a bug in the visualizer with my system configuration which Terry
fixed in a separate branch (fix_html_eol). I rebased to that branch to get
the visualizer working on my system. I do agree that these commits should
be kept separate.

nengo_viz/components/action.py
```diff
@@ -0,0 +1,276 @@
...
+    else:
+        return Action(net_graph, **info)
+
+class Action(object):
+    def __init__(self, net_graph, **info):
+        self.net_graph = net_graph
+        self.info = info
+
+    def apply(self):
+        pass
```
tcstewar: These should probably be ```raise NotImplementedError()``` rather than ```pass```

tcstewar:
Just wondering if anyone is currently in charge of making the changes listed above and rebasing this to master.  This is the next big thing I'd like to get merged in....  :)

bjkomer:
I can start on that tonight. I haven't used the squash command before, so I'll back things up to another branch in case I accidentally destroy all of the history :palm_tree:

hunse:
Squashing isn't too hard. Just use `git rebase -i`, and it's pretty self-explanatory. Also, since git keeps a full log of everything you do (see `git reflog`), it's pretty hard to actually mess anything up, but making a back-up branch is a good idea.

s72sue:
Brent, let me know if you need me to work with you. I can stay back in the
lab or we can split things up if required.

-Sue

bjkomer:
Rebase and squash complete! There were conflicts in almost every commit, so I'm sure something wrong slipped through, but hopefully not too much. The big commit also still needs to be updated somehow to reflect multiple authorship.

tcstewar:
This is still marked as a "work in progress".  Is that still the case, or should I look into merging it?

tbekolay:
GH shows 141 commits in this branch, so something went wrong in this rebase... @bjkomer, can you confirm that the only commits that should be in this branch are 212ed3ef5e53253826f28300c630ff334f60eff3 f708030ee9ba0fddf358252e46fa0b9c40ae639c (?) 9db8ea76bae96798385bf038098465cd1799d3fb 81687d3307a9041e4279bcabe4d11173ea6985f8 92aa77c68ad22b7fe6c5778347689bb09d5fc733 5431587d8d0ee4a6bbbe09005f7f21bcb29a4c1e

tbekolay:
If so, then I can fix this up no problemo.

bjkomer:
Yes, those 5 commits should be the only ones. There are still some bugs in this code (sometimes items get popped from an empty list in update_client in the components), but I think this is related to thread safety stuff getting fixed in another pull request #148 

I just noticed that expanding/collapsing nested networks doesn't work anymore. So something must have gone wrong in the rebase and/or the way expand/collapse is handled has changed since this code was written. I'll work on fixing that, and then it should be good to go.

bjkomer:
It should be good to go now!

tbekolay:
Cool, I'll fix up the history.

tbekolay:
I have a WIP fixed up history in [`undo_stack_3`](https://github.com/nengo/nengo_viz/compare/undo_stack_3). This isn't right yet, but I have to poke around to figure out which of the two are correct in all the cases where they differ.

tcstewar:
> but I think this is related to thread safety stuff getting fixed in another pull request #148

That's in master now...  :)

tbekolay:
OK, the history of [`undo_stack_3`](https://github.com/nengo/nengo_viz/compare/undo_stack_3) should be correct now. Also tried to give a decent commit message, and mark the commit at co-authored (sorta). @bjkomer and @s72sue, could you take a look at that diff and make sure it has all of the right changes? Let me know, and if it looks good then I'll replace `undo_stack_2` with `undo_stack_3`.

Even when that happens, I think this is still a WIP; the info windows don't work with these changes (but they should be easy to get working) and I'm doing a little bit of code cleanup.

tbekolay:
Continuing to review this branch... is there going to be an issue between this branch and #159? That branch made it such that the config file is saved at most every 2 seconds. This branch does undo / redo by looking at the difference between old / new configs. Does it do this by looking at the file, or is this an internal representation of the config?

tcstewar:
> Does it do this by looking at the file, or is this an internal representation of the config?

There is an internal representation (it's a nengo.Config object).

s72sue:
Right, we used internal representations for config (not looking at the
file).

tbekolay:
OK, I added two commits to [`undo_stack_3`](https://github.com/nengo/nengo_viz/compare/undo_stack_3):

- 884ccd63a3109615fdfcf97d97373cfc90186 This one that will be squashed. It has some suggested style cleanups, and a few legit fixes (like ensuring that modals work).
- 8a0308d3af402f4 Hooks the undo/redo stuff up to the toolbar, rather than the right-click menu.

I did some rudimentary testing, and this seems to work as intended... but I don't think I know all the things that can be undone / redone, so additional testing would be good! And, again, I wanted to make sure that I didn't miss anything in the rebase, so I won't pushed to this branch (`undo_stack_2`) until @bjkomer or @s72sue confirm that I didn't miss anything in `undo_stack_3`.

tbekolay:
Also made #223 to track something that should happen, but not in this PR.

bjkomer:
Thanks! One bug I noticed is that undoing a change to plot properties (show/hide label, changing range, etc) will make the numbers on the axes disappear in undo_stack_3, but it is working in undo_stack_2. I think it has to do with the update_layout function on the javascript side not doing all the things it is supposed to. I won't have Internet for the next while, but I'll test it some more when I get home tonight

s72sue:
What Brent mentioned is happening only for the value graphs. The spike and
x-y value plots do not have this bug.

Additional bugs:
1) on removing a graph, undoing doesn't bring the graph back
    - Fix: add 'uid_graph' as an attribute of RemoveGraph in action.py
 [ self.uid_graph = None ]

2) on setting the range on a value plot, undoing doesn't set it back to
previous values (only for value graphs)
3) on moving or resizing a value plot, undoing makes the numbers disappear
on the axis

2 and 3 seem to be related to the same bug in update_layout, which Brent
mentioned. I will do another run of testing tomorrow morning to make sure
we are not missing anything.

-Sue

s72sue:
I have fixed the bugs mentioned above in undo_stack_3. Everything else
seems to look good, haven’t found any other bugs yet.

-Sue

tbekolay:
OK, I rebased to master, and squashed my and Sue's commits into the big undo/redo commit, and pushed to this branch, so this should now be totally ready for review!

nengo_viz/components/netgraph.py
```diff
@@ -194,12 +195,44 @@ def message(self, msg):
...
+            else:
+                self.redo()
         else:
             print('received message', msg)
 
+    def undo(self):
+        if self.viz.undo_stack:
+            action = self.viz.undo_stack.pop()
+            re = []
+            for act in action:
```
hunse: Can there ever be more than one action in `action`? It seems to me it's always a list of one. If so, I think we should get rid of the ability to group actions together, since it just makes the code more complex. We can always add it back later if we're going to use it.

nengo_viz/static/viz.js
```diff
@@ -152,14 +149,17 @@ VIZ.Component = function(parent, args) {
             var dz = event.deltaRect.right;
             var da = event.deltaRect.bottom;
 
-            var scale = VIZ.pan.cord_per_px(VIZ.Screen);
-
-            var x = parseFloat(target.getAttribute('data-x')) + ((dx + dz) / 2 * scale.x) ;
-            var y = parseFloat(target.getAttribute('data-y')) + ((dy + da) / 2 * scale.y) ;
-            target.setAttribute('data-x', x);
-            target.setAttribute('data-y', y);
+            var x = self.x + (dx + dz) / 2 / (viewport.w * viewport.scale);
```
hunse: Couldn't we just do `self.x += ...`

tcstewar:
@s72sue @bjkomer and I just tracked down a nasty bug introduced here.  Basically, all .cfg files with any graphs in them are horribly broken at the moment (to the degree that you can't even start nengo_viz).

nengo_viz/static/viz_slider.js
```diff
@@ -260,3 +260,41 @@ VIZ.Slider.prototype.layout_info = function () {
     info.max_value = this.sliders[0].scale.domain()[0];
     return info;
 };
+
+VIZ.Slider.prototype.update_layout = function (config) {
+    //FIXME: this has to be backwards to work. Something fishy must be going on
```
hunse: Can we fix this before merge? It's just due to the min and max values being set backwards in `layout_info` just a few lines above, right?

nengo_viz/components/netgraph.py
```diff
@@ -12,6 +10,8 @@
 from nengo_viz.disposable_js import infomodal
 import nengo_viz.layout
 
+from .action import create_action
```
hunse: Are we avoiding relative imports like in Nengo?

nengo_viz/components/netgraph.py
bjkomer: Expanding nested networks used to create more than one action, which needed to be grouped together. That code has since been changed to use only one action, so the list isn't needed anymore. I can see it being used for something like undoing the 'reset config', but for now we don't need it.

nengo_viz/components/netgraph.py
tcstewar: Possibly...  we should stick with the same standard.  I think I'd prefer ```from nengo_viz.components.action import create_action``` here, but I keep forgetting what standard we've been going with.

nengo_viz/static/viz_slider.js
tcstewar: I think I'd leave this for a separate PR, as there's a bunch of places that need to get swapped.

nengo_viz/static/viz.js
tcstewar: Good idea..  I'll switch that over....

tcstewar:
This is all looking good to me now.  If everyone else is good with it, I'd like to merge shortly...  :)

tbekolay:
This looks good to me! Sorry for breaking the graphs ;)

tcstewar:
> Sorry for breaking the graphs ;)

No worries...  it was a neat combination of different coding styles that caused it...  :)

PR #184, opened 2015-05-24
Fixed bug in reload re:iteritems

tcstewar:
The switchfrom .items() to nengo.utils.compat.iteritems() wasn't
tested well enough here.  The loop changes the dictionary, and
iteritems() really doesn't like that.  This wasn't a problem with
the .items() approach, since it's only ever changing the item
it's on in the loop.  The fix here is to duplicate the dictionary.


tcstewar:
@tbekolay could you take a very quick look at this one-line change?

tbekolay:
This looks good to me! FYI in Python 3, `.items()` is what Python 2 called `iteritems`, so this would have been a Python 3 bug regardless.

tcstewar:
> FYI in Python 3, .items() is what Python 2 called iteritems, so this would have been a Python 3 bug regardless.

Good point!  :)

PR #186, opened 2015-05-25
Modal inputs

celiasmith:
This was branched off of a merge of graphics-overhaul and info-window.  It changes all the value setting and range setting dialogs to bootstrap modals.  It doesn't currently do validation. I'll branch this and do validation in a separate PR.

### DIFF
--- a/nengo_viz/static/viz_modal.js
+++ b/nengo_viz/static/viz_modal.js
@@ -24,11 +24,15 @@ VIZ.Modal.prototype.title = function(title) {
     this.$title.text(title);
 }
 
-VIZ.Modal.prototype.footer = function(type){
+VIZ.Modal.prototype.footer = function(type, ok_function){
     this.$footer.empty();
 
     if (type === "close") {
         this.$footer.append('<button type="button" class="btn btn-default" data-dismiss="modal">Close</button>');
+    } else if (type === "ok_cancel") {
+        this.$footer.append('<button type="button" class="btn btn-default" data-dismiss="modal">Cancel</button>');
+        this.$footer.append('<button id="OK" type="button" class="btn btn-primary" data-dismiss="modal">OK</button>');
+        $('#OK').on('click', ok_function);
     } else if (type === 'confirm_reset') {
         this.$footer.append('<button type="button" id="confirm_reset_button" class="btn btn-primary">Reset</button>');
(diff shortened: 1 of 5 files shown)


tbekolay:
Just to reinforce what Chris said, only the last 4 commits in this branch should be the subject of review; the other stuff is either in master (the `info-window` stuff) or will soon be (the `graphical-overhaul` stuff). This branch should be rebased to master after the `graphical-overhaul` stuff is merged in. (Note: explicitly cherry-picking the last 4 commits here will likely be easier than doing an actual rebase)

jgosmann:
Isn't cherry-picking the last 4 commits pretty much the same as rebasing the last 4 commits?

tbekolay:
No, because this branch contains old versions of the branches that are / will be in master. So, git can't automatically realize that this branch shares history with master, and will instead try to apply all of these commits that have already been applied to master to master. Which you would think would resolve conflict free, but definitely doesn't (especially when you hit merge commits).

jgosmann:
You can tell `git rebase` to only rebase the last 4 commits (`git rebase --onto master e4f7314 modal-inputs`). But yes, without that git will mess up horribly.

tbekolay:
Yeah; I can never remember the `--onto` semantics the first time :rainbow:

tbekolay:
OK! I rebased to master. I think this is a big improvement. Will rebase #195 once this is merged (unless we want to close this and just review #195, rebased to this branch?)

tcstewar:
Is there a way to have it so that if you hit Enter after typing in the value in the modal, it clicks Okay for you?

tbekolay:
There undoubtedly is. Will investigate!

tcstewar:
> There undoubtedly is. Will investigate!

Other than that this looks good to me... if it turns out the hitting Enter thing isn't simple, we can leave it to another PR and merge this one as-is.

celiasmith:
Yes it's not straightforward.  I set it up so it auto highlights the text field, which means when you type it does the right thing.  If you tab twice and hit enter, it will submit.  If you hit enter without that, it will validate it.  It was not obvious how to both validate and submit with enter.

tcstewar:
> Yes it's not straightforward.

Okay, let me merge this for now and then we'll sort out the validate-and-submit once the next PR is in.

PR #187, opened 2015-05-25
Fixed subnetwork expansion after collapse

tcstewar:
Right now if you collapse a sub-network and then re-expand it, the internal items do not re-appear.

This bug was introduced by the reload system, since it tries to avoid re-creating things that already exist, but when collapsing a network it doesn't realize that those items no longer exist on the client.

### DIFF
--- a/nengo_viz/components/netgraph.py
+++ b/nengo_viz/components/netgraph.py
@@ -208,6 +208,14 @@ class NetGraph(Component):
     def act_collapse(self, uid):
         net = self.uids[uid]
         self.config[net].expanded = False
+        for ens in net.ensembles:
+            del self.uids[self.viz.viz.get_uid(ens)]
+        for net in net.networks:
+            del self.uids[self.viz.viz.get_uid(net)]
+        for node in net.nodes:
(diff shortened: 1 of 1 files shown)


tcstewar:
Due to the severity of this bug (and the simplicity of the fix), I'm just going to merge this fix in.

hunse:
This doesn't seem to fix the problem for me. When collapsing an expanded network, I now get a KeyError from [`netgraph.py:216`](https://github.com/nengo/nengo_viz/blob/master/nengo_viz/components/netgraph.py#L216).

tcstewar:
> This doesn't seem to fix the problem for me. When collapsing an expanded network, I now get a KeyError from netgraph.py:216.
Interesting...  what model are you using to get that?

I'll add some checks for those items, in any case.

hunse:
That's in Spaun.

tcstewar:
Damnit, I see my bug....   I'm overwriting the ```net``` variable.

tcstewar:
Okay, I've added a new PR for fixing this:  #189   Does that one work for you, @hunse ?

PR #188, opened 2015-05-25
Fixes refreshing the page

tcstewar:
I broke refreshing the page somewhere along the line.  This seems to fix it for me.

### DIFF
--- a/nengo_viz/server.py
+++ b/nengo_viz/server.py
@@ -45,9 +45,18 @@ class Server(swi.SimpleWebInterface):
     def ws_viz_component(self, client, uid):
         """Handles ws://host:port/viz_component with a websocket"""
         # figure out what component is being connected to
+
+        viz_sim = Server.viz_sim
+
+        component = viz_sim.uids[uid]
         try:
             while True:
-                component = self.viz_sim.uids[uid]
+                if viz_sim.finished:
+                    component.finish()
+                    return
+                if viz_sim.uids[uid] != component:
+                    component.finish()
+                    component = viz_sim.uids[uid]
                 # read all data coming from the component
                 msg = client.read()
                 while msg is not None:


tcstewar:
@tbekolay or @hunse if you have a moment to look at this one, that'd be great...  :)  Thank you!

celiasmith:
If you reload while running a simulation, it gets very slow.  Which doesn't seem to happen if you pause and reload (open up something to reset the sim) and start again.  Not sure how related to this PR that is.  Otherwise, looks good.

tcstewar:
> If you reload while running a simulation, it gets very slow. Which doesn't seem to happen if you pause and reload (open up something to reset the sim) and start again. Not sure how related to this PR that is.

Hmm, looks like hitting reload doesn't stop the previous simulation.  I'll do some poking around and see what I can sort out...  seems to be unrelated to this PR, though.

celiasmith:
Ok, I'll merge this.  I made that observation in #198

nengo_viz/server.py
```diff
@@ -45,9 +45,18 @@ def swi_shutdown(self, *path):
...
         # figure out what component is being connected to
+
+        viz_sim = Server.viz_sim
+
+        component = viz_sim.uids[uid]
         try:
             while True:
-                component = self.viz_sim.uids[uid]
+                if viz_sim.finished:
+                    component.finish()
```
hunse: Sorry I'm late to the game, but it seems strange to have this here since it's in the `finally` block.

nengo_viz/server.py
```diff
@@ -45,9 +45,18 @@ def swi_shutdown(self, *path):
...
+                    return
```
hunse: I'd use `break` instead of `return` here just because I think it's a bit clearer, but no biggie either way.

nengo_viz/server.py
tcstewar: Good point...  removing.... :)

nengo_viz/server.py
tcstewar: Yikes, good point as well...

tcstewar: Since this was already merged, I put these changes in #200

PR #189, opened 2015-05-25
Better fix for expanding collapsed networks

tcstewar:
This removes an overwritten variable bug and handles nested networks


hunse:
Looks good and seems to work.

Issue #190, opened 2015-05-25
Can't move graphs in Chromium 34.0.1847.116 Ubuntu 13.10


studywolf:
Just like the title says, will have a PR in a moment

tcstewar:
Fixed in #191

PR #191, opened 2015-05-25
look for -webkit-transform if transform undefine

studywolf:
quick fix for #190

### DIFF
--- a/nengo_viz/static/viz.js
+++ b/nengo_viz/static/viz.js
@@ -26,7 +26,12 @@ VIZ.set_transform = function(element, x, y) {
 
 
 VIZ.get_transform = function(element) {
-    var holde = $(element).css('transform').match(/(-?[0-9\.]+)/g); //Ugly method of finding the current transform of the element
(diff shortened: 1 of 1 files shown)


tcstewar:
It looks like on some older versions of Chrome, the ```transform``` attribute gets ignored and it uses ```-webkit-transform``` instead.  We're handling that in ```VIZ.set_transform``` and this PR makes it so we handle it in ```VIZ.get_transform``` as well.

tcstewar:
(and it looks good to me, but I wouldn't mind another set of eyes on it before I merge it)

tcstewar:
I took @hunse 's suggestion and also reformatted a bit to respect line lengths....  

@studywolf does this still fix the problem for you?  If you could quickly confirm that I'll merge it in.

hunse:
Also fix up the commit message a bit before merge. "undefined" is missing its second 'd', and it would be nice to have a bit of a body to say that this was a problem that happened on Chrome, etc.

studywolf:
yup can drag things!

PR #192, opened 2015-05-25
Fixed typo in error message generation for swi

tcstewar:
This stopped swi from generating useful error messages


hunse:
Pretty obvious bug, and the fix looked fine to me, so I went ahead and merged.

PR #194, opened 2015-05-25
Moved Modal declaration from page into viz_modal.js

studywolf:
Just to clean up the page.html a bit, created an addModal function in viz_modal and called it there to add all the modal declaration stuff to the page. Nothing tricky, did basic testing and everything worked.

### DIFF
--- a/nengo_viz/static/viz_modal.js
+++ b/nengo_viz/static/viz_modal.js
@@ -416,3 +416,51 @@ VIZ.Modal.make_conn_path_dropdown_list = function($container, others_uid, obj_ty
         }
     }
 }
+
+
+function addModal() {
+    var modalFade = document.createElement('div');
+    modalFade.className = 'modal fade';
+    modalFade.setAttribute('style','z-index:100000001');
+
+    var modalDialog = document.createElement('div');
+    modalDialog.className = 'modal-dialog';
+
+    var modalContent = document.createElement('div');
+    modalContent.className = 'modal-content';
+
+    var modalHeader = document.createElement('div');
+    modalHeader.className = 'modal-header';
+
+    var modalHeaderButton = document.createElement('button');
+    modalHeaderButton.className = 'close';
+    modalHeaderButton.type='button';
+    modalHeaderButton.setAttribute('data-dismiss','modal');
+    modalHeaderButton.setAttribute('aria-label','Close');
+
+    var modalHeaderButtonSpan = document.createElement('span');
+    modalHeaderButtonSpan.setAttribute('aria-hidden','true');
+    modalHeaderButtonSpan.innerHTML = '&times';
+
+    var modalHeaderH4 = document.createElement('h4');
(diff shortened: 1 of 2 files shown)


tbekolay:
I had it in the page because it's much shorter (14 lines vs. 48 lines) and free (it takes a few milliseconds to execute this javascript, vs. basically nothing to parse the HTML). Additionally, we never want more than one modal to be created -- Bootstrap explicitly doesn't support this (though [this enhanced version](https://github.com/jschr/bootstrap-modal) does).

If anything, I think that we should start shoving more stuff into `page.html`, as constructing HTML through Javascript is super cumbersome, even with JQuery. Browsers read and parse HTML in essentially no time, so we get a lot from putting stuff into `page.html`.

So yeah, I'm -3 on this PR. Sorry! :rainbow:

studywolf:
maxresdefault

xchoo:
Question: What does moving the modal code into viz_modal give us? We are only calling the function once, yes?

studywolf:
@xchoo, so glad you asked. 
Basically my thought was to bring everything in line, so that all the things that are only created once are all declared in the same way. After talking with Trevor it seems like the better way would be instead to bring code for sim_control and top_toolbar out of the javascript and into the page.html, rather than pushing the modal creation into viz_modal. So I'll close this and look and that instead!

cchan987:
I would think this approach would be a cleaner way to add the modal.

Currently everything is being created in js files and that helps when trying to locate the code that created a particular element.

tbekolay:
There's definitely an argument to be made for keeping everything in JS files. There's additional flexibility there, but IMO it comes at the cost of readability and performance. JS is not great at creating HTML; no matter what nice APIs you use (JQuery, etc) it's harder to read the JS that constructs a block of HTML than it is to read the block of HTML that is created in the end. Hence, I spend more time in the Chrome inspector looking at the HTML I've constructed than I do looking at the JS constructing it.

With web stuff there's honestly no way around having to deal with HTML, CSS and JS files simultaneously. Given that, you want to use each technology for what it's good at. HTML is good for defining the parts of your page that won't change (or, rather, the core set of things that will be mutated by JS). CSS is good for defining all of the style-related things. JS is good for the dynamic bits. If you don't need something to be dynamic (e.g., the modal window, sim control, and top toolbar) then you can provide a much nicer JS API for those things by defining the structure once (in HTML) and mutating it, rather than constructing it from scratch every time. It's also faster, though for what we're doing, that kind of speed isn't a huge concern (yet).

The situation where it would be best in JS is if we want to be making many of these, dynamically, at run-time (i.e., not just when the page is first loaded). It doesn't seem like that's the case for these three things, but if that changes later on, we can always move it into JS later.

xchoo:
I think this raises the bigger question of going through the existing code base to make sure that this specification (the one that Trevor is describing) is adhered to.

tbekolay:
@xchoo Yeah, that's what Travis is doing.

PR #195, opened 2015-05-25
Modal validation

celiasmith:
This was branched off of modal-inputs. It validates user input before allowing the input to be submitted, greying things out, providing error messages, etc.  Right now, it will wait 5s before providing any error messages (or provides them immediately on hitting return).  The default was 500ms, but error messages are pretty much guaranteed to show up, and don't seem relevant til you hit return.  (To submit the form, you have to click on 'ok' or hit return when 'ok' is selected (2 tabs away from the input field).)  I wondered about having 'return' submit the form, but haven't implemented that (it might conflict with the validation).

### DIFF
--- a/nengo_viz/static/lib/js/validator.js
+++ b/nengo_viz/static/lib/js/validator.js
@@ -1,0 +1,303 @@
+/*!
+ * Validator v0.8.1 for Bootstrap 3, by @1000hz
+ * Copyright 2015 Cina Saffary
+ * Licensed under http://opensource.org/licenses/MIT
+ *
+ * https://github.com/1000hz/bootstrap-validator
+ */
+
++function ($) {
+  'use strict';
+
+  var inputSelector = ':input:not([type="submit"], button):enabled:visible'
+  // VALIDATOR CLASS DEFINITION
+  // ==========================
+
+  var Validator = function (element, options) {
+    this.$element = $(element)
+    this.options  = options
+
+    options.errors = $.extend({}, Validator.DEFAULTS.errors, options.errors)
+
+    for (var custom in options.custom) {
+      if (!options.errors[custom]) throw new Error('Missing default error message for custom validator: ' + custom)
+    }
+
+    $.extend(Validator.VALIDATORS, options.custom)
+
+    this.$element.attr('novalidate', true) // disable automatic native validation
+    this.toggleSubmit()
+
+    this.$element.on('input.bs.validator change.bs.validator focusout.bs.validator', $.proxy(this.validateInput, this))
+    this.$element.on('submit.bs.validator', $.proxy(this.onSubmit, this))
+
+    this.$element.find('[data-match]').each(function () {
+      var $this  = $(this)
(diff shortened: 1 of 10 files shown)


tbekolay:
Rebased and ready for review!

Personally, I was very confused for a while because I put in some invalid data and then couldn't push OK. But it also didn't give any feedback. Eventually the error message appeared, so I believe this is because of the explicit 5 second delay [here](https://github.com/nengo/nengo_viz/pull/195/files#diff-7356065b68258d6522ff4fd8032a02b9R479). @celiasmith, what's the purpose of the 5 second delay? Could we maybe make it 0.5 seconds or something?

tcstewar:
> Personally, I was very confused for a while because I put in some invalid data and then couldn't push OK.

Yup, that behaviour is confusing to me too.

And I'd very much like to have hitting enter try to both validate and (if successful) submit the form...  the extra tabbing or mouse moving is odd...

That said, it's really nice to have this validation stuff!  Works great!

nengo_viz/static/lib/js/validator.js
```diff
@@ -0,0 +1,303 @@
+/*!
+ * Validator v0.8.1 for Bootstrap 3, by @1000hz
+ * Copyright 2015 Cina Saffary
+ * Licensed under http://opensource.org/licenses/MIT
```
tcstewar: This should be added to LICENSE.md

nengo_viz/static/viz_xyvalue.js
```diff
@@ -107,8 +107,8 @@ VIZ.XYValue.prototype.on_resize = function(width, height) {
 VIZ.XYValue.prototype.generate_menu = function() {
     var self = this;
     var items = [];
-    items.push(['Set range', function() {self.set_range();}]);
-    items.push(['Set X, Y indices', function() {self.set_indices();}]);
+    items.push(['Set range...', function() {self.set_range();}]);
+    items.push(['Set X, Y indices...', function() {self.set_indices();}]);
```
tcstewar: Should set_indices also have a validator?

celiasmith:
> what's the purpose of the 5 second delay? Could we maybe make it 0.5 seconds or something?

I put that in because it was annoying to have the validation happening all the time when i wasn't done typing.  We can make it whatever we want.  The long delay was basically to turn it 'off' since validation errors show up immediately on hitting return (you also know it's invalid because there's no 'ok' button... but yeah, maybe too subtle).

I can make it .5s .... or i could probably make it so 'ok' is always lit and errors reported when you hit 'ok' (and hopefully 'return').  That sounds like it might be the best option.... thoughts?

tbekolay:
> make it so 'ok' is always lit and errors reported when you hit 'ok' (and hopefully 'return')

Sounds like the best option to me!

celiasmith:
The modals now work as promised... also did the two changes requested by @tcstewar in line notes.

tcstewar:
looks great!  merging....

Issue #196, opened 2015-05-25
Auto layouts assumes a specific scaling


celiasmith:
If you do auto layout of the following .cfg file (2d.py.cfg), you will see a behaviour where everything is made small when laid out (even though there's lots of screen available).  I believe this happens because auto-layout is assuming a particular screen size or zoom level.  I think it would be good to either:

1. layout with an object that is fixed-size wrt the current zoom.  (This is done with e.g. sliders right now.); or 

2. perhaps the zoom should be changed on an auto-layout (presumably to 1).  This option could mess up the relative size of open graphs/sliders (unless they are also layed out somehow).

```
_viz_net_graph = nengo_viz.components.NetGraphTemplate()
_viz_sim_control = nengo_viz.components.SimControlTemplate()
_viz_config[_viz_sim_control].kept_time = 4
_viz_config[_viz_sim_control].shown_time = 0.5
_viz_config[ens].pos=(1.2796636011548939, -0.06306345847528876)
_viz_config[ens].size=(0.5529063510738321, 1.2538348748760917)
_viz_config[model].pos=(6.324660393025442, 5.872174709161818)
_viz_config[model].size=(0.08390545288824205, 0.08390545288824205)
_viz_config[model].expanded=True
_viz_config[model].has_layout=True
_viz_config[stimulus].pos=(-2.3842674505317256, -0.027871992320583117)
_viz_config[stimulus].size=(0.7781600920340797, 0.8920988343758754)
```

tcstewar:
Ooo...  good point.   I think option 1 works best -- the auto-layout should target the current zoom level.  That should also be pretty easy to implement, I think.

tcstewar:
I've done a quick fix in #197.  It goes with option #1 here.

celiasmith:
The current fix breaks on reload. (Just run an example and hit 'reload').  I didn't revert the merge as I assume it's faster/easier to just do another(?). But maybe @tbekolay will hate that...

tcstewar:
That's not anything to do with this fix.  That's the bug that #188 is meant to fix.

tcstewar:
(basically, reload has been broken for a while and no one noticed till this morning)

celiasmith:
kewl, then no need to revert.  i'll close this and review the other.

PR #197, opened 2015-05-25
Made autolayout use the full available space

tcstewar:
This addresses #196 and simply adjusts the top-level autolayout such that takes into account the global pan/zoom state.


celiasmith:
Looks good to me!

PR #200, opened 2015-05-26
Tiny cleanup to websocket loop

tcstewar:
Follows up on @hunse 's comments in #188 that were missed before it was merged.

Basically just removes a redundant ```component.finished()``` and uses a ```break``` instead of a ```return```.


hunse:
LGTM

PR #201, opened 2015-05-26
Top toolbar

cchan987:
Creates a top toolbar with:
-Ability to open new file
-Reset the model layout
-undo button (inactive)
-redo button (inactive)
-Config options menu (active, needs to be populated)


tcstewar:
Cool!  If you could take a stab at rebasing this to master, I'll start going through the code....

studywolf:
might be worth getting the undo PR in first, so that the inactive buttons can be hooked up?

tcstewar:
> might be worth getting the undo PR in first, so that the inactive buttons can be hooked up?

That is a good point.  I'll push on the undo people more....

nengo_viz/components/toolbar.py
```diff
@@ -0,0 +1 @@
+##toolbar
```
tcstewar: Should this file exist?

studywolf:
OK, rebased to master. Fixed things I broke during rebasing. Then I rebased to takehtmloutofjs and took the html out of viz_top_toolbar and put it into page.html.

I believe that everything works again, the config window is just empty, and the undo/redo have a TODO in viz_top_toolbar for where they should be hooked up.

studywolf:
Oh one thing that came up. In viz_sim_control.css, to make things look right on my desktop I deleted the margin-top: 40px inside the .metrics-container.table. But now it's up too high on my laptop. Not sure what's going on there. That's actually a thing from takehtmloutofjs.

PR #207, opened 2015-05-27
Confirm model reset

cchan987:
The user is now prompted with a modal to confirm when they click the reset model button.

### DIFF
--- a/nengo_viz/static/viz_modal.js
+++ b/nengo_viz/static/viz_modal.js
@@ -26,7 +26,7 @@ VIZ.Modal.footer = function(type){
     else if (type === 'confirm_reset') {
         $footer.append('<button type="button" id="confirm_reset_button" class="btn btn-primary">Reset</button>');
         $footer.append('<button type="button" class="btn btn-default" data-dismiss="modal">Close</button>');
-        $('#confirm_reset_button').on('click', toolbar.reset_model_layout();});
+        $('#confirm_reset_button').on('click', function() {toolbar.reset_model_layout();});
     }
     else {
         console.warn('Modal footer type ' + type + ' no recognized.')
--- a/nengo_viz/static/viz_modal.js
+++ b/nengo_viz/static/viz_modal.js
@@ -22,7 +22,13 @@ VIZ.Modal.footer = function(type){
 
     if (type === "close") {
         $footer.append('<button type="button" class="btn btn-default" data-dismiss="modal">Close</button>');
-    } else {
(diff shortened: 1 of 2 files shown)


nengo_viz/static/viz_top_toolbar.js
```diff
@@ -20,7 +20,13 @@ VIZ.Toolbar = function(filename) {
     this.toolbar = $('#top_toolbar')[0];
 
     $('#Open_file_button')[0].addEventListener('click', function () {self.file_browser()});
-    $('#Reset_layout_button')[0].addEventListener('click', function () {self.reset_model_layout()});
+    $('#Reset_layout_button')[0].addEventListener('click', function () {
+        VIZ.Modal.title('Are you sure?');
```
tcstewar: As a general design principle, dialog boxes like this should be explicit about what you're confirming.  So instead of "Are you sure?", this should be something like "Are you sure you wish to reset this layout, removing all the graphs and resetting the position of all items?"

nengo_viz/static/viz_modal.js
```diff
@@ -22,7 +22,13 @@ VIZ.Modal.footer = function(type){
 
     if (type === "close") {
         $footer.append('<button type="button" class="btn btn-default" data-dismiss="modal">Close</button>');
-    } else {
+    } 
+    else if (type === 'confirm_reset') {
+        $footer.append('<button type="button" id="confirm_reset_button" class="btn btn-primary">Reset</button>');
+        $footer.append('<button type="button" class="btn btn-default" data-dismiss="modal">Close</button>');
+        $('#confirm_reset_button').on('click', function() {toolbar.reset_model_layout();});
```
tbekolay: This function should be passed in as a parameter, not hard-coded into this function. See #186 for an example.

tcstewar: Oops, I just merged this.  Damnit.  I completely agree with your comment but I figured people were waiting till a future PR to fix that up.  In any case, if it gets fixed here then I can cherry-pick the change into master too.

tbekolay: No, it's merged, just do it in another PR. #186 will have to be rebased to master, and it does the same thing, so we can do it there.

tcstewar: > No, it's merged, just do it in another PR. #186 will have to be rebased to master, and it does the same thing, so we can do it there.

Sounds good to me.  :)

PR #208, opened 2015-05-27
Added 'nengo_viz' script to start the visualizer

tcstewar:
This should act like nengo_gui, in that you can specify a file to open, otherwise it will open a default.py file.

One change from nengo_gui is that the file browser shows files from where you ran the script, and in addition it also shows the built-in examples folder.  This way people have access to the built-in examples as well as a more typical file browsing in their local folders.

### DIFF
--- a/nengo_viz/server.py
+++ b/nengo_viz/server.py
@@ -23,7 +23,7 @@ class Server(swi.SimpleWebInterface):
         d = unquote(dir)
         ex_tag = '//examples//'
         if d == '.':
-            r.append('<li class="directory collapsed">'
+            r.append('<li class="directory collapsed examples_dir">'
                      '<a href="#" rel="%s">%s</a></li>' % (ex_tag, ex_tag))
             path = '.'
         elif d.startswith(ex_tag):
--- a/nengo_viz/server.py
+++ b/nengo_viz/server.py
@@ -22,9 +22,10 @@ class Server(swi.SimpleWebInterface):
         r = ['<ul class="jqueryFileTree" style="display: none;">']
         d = unquote(dir)
         ex_tag = '//examples//'
+        ex_html = '<em>built-in examples</em>'
         if d == '.':
             r.append('<li class="directory collapsed examples_dir">'
-                     '<a href="#" rel="%s">%s</a></li>' % (ex_tag, ex_tag))
+                     '<a href="#" rel="%s">%s</a></li>' % (ex_tag, ex_html))
             path = '.'
         elif d.startswith(ex_tag):
             path = os.path.join(nengo_viz.__path__[0],
--- a/nengo_viz/server.py
+++ b/nengo_viz/server.py
@@ -18,18 +18,31 @@ class Server(swi.SimpleWebInterface):
     """Web server interface to nengo_viz"""
 
     def swi_browse(self, dir):
-        self.script_path = '.'
         if self.user is None: return
         r = ['<ul class="jqueryFileTree" style="display: none;">']
         d = unquote(dir)
-        for f in sorted(os.listdir(os.path.join(self.script_path, d))):
-            ff = os.path.relpath(os.path.join(self.script_path, d,f), self.script_path)
-            if os.path.isdir(os.path.join(self.script_path, d, ff)):
-                r.append('<li class="directory collapsed"><a href="#" rel="%s/">%s</a></li>' % (ff,f))
+        ex_tag = '//examples//'
+        if d == '.':
+            r.append('<li class="directory collapsed">'
+                     '<a href="#" rel="%s">%s</a></li>' % (ex_tag, ex_tag))
+            path = '.'
+        elif d.startswith(ex_tag):
+            path = os.path.join(nengo_viz.__path__[0], 
+                                'examples', d[len(ex_tag):])
+        else:
+            path = os.path.join('.', d)
+
+        for f in sorted(os.listdir(path)):
+            ff = os.path.relpath(os.path.join(path, f), '.')
+            ff = os.path.join(path, f)
+            if os.path.isdir(os.path.join(path, f)):
+                r.append('<li class="directory collapsed">'
+                         '<a href="#" rel="%s/">%s</a></li>' % (ff,f))
             else:
                 e = os.path.splitext(f)[1][1:] # get .ext and remove dot
                 if e == 'py':
-                    r.append('<li class="file ext_%s"><a href="#" rel="%s">%s</a></li>' % (e,ff,f))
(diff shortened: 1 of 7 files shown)


tcstewar:
And this one's been rebased too, ready for review.

nengo_viz/server.py
```diff
@@ -11,24 +11,38 @@
     from urllib.parse import unquote
 
 import nengo_viz.swi as swi
+import nengo_viz
 
 
 class Server(swi.SimpleWebInterface):
     """Web server interface to nengo_viz"""
 
     def swi_browse(self, dir):
```
tbekolay: This isn't part of this PR, but we really shouldn't be using the variable `dir` since it shadows the builtin `dir`.

nengo_viz/server.py
```diff
@@ -11,24 +11,38 @@
...
         if self.user is None: return
         r = ['<ul class="jqueryFileTree" style="display: none;">']
         d = unquote(dir)
-        for f in sorted(os.listdir(os.path.join(self.script_path, d))):
-            ff = os.path.relpath(os.path.join(self.script_path, d,f), self.script_path)
-            if os.path.isdir(os.path.join(self.script_path, d, ff)):
-                r.append('<li class="directory collapsed"><a href="#" rel="%s/">%s</a></li>' % (ff,f))
+        ex_tag = '//examples//'
+        if d == '.':
+            r.append('<li class="directory collapsed examples_dir">'
```
tbekolay: It seems weird to me to have HTML being sent down the wire... I would expect a list of files to be sent down, and the templating engine to generate this HTML.

nengo_viz/server.py
```diff
@@ -43,8 +57,22 @@ def swi_favicon_ico(self):
         icon = pkgutil.get_data('nengo_viz', 'static/favicon.ico')
         return ('image/ico', icon)
 
+    def create_login_form(self):
+        if self.attempted_login:
+            message = 'Invalid password.  Try again.'
```
tbekolay: Single space after period (maybe would get rendered well with HTML, but looks weird in fixed-width font)

nengo_viz/server.py
```diff
@@ -11,24 +11,38 @@
...
     def swi_browse(self, dir):
-        self.script_path = '.'
         if self.user is None: return
         r = ['<ul class="jqueryFileTree" style="display: none;">']
         d = unquote(dir)
-        for f in sorted(os.listdir(os.path.join(self.script_path, d))):
-            ff = os.path.relpath(os.path.join(self.script_path, d,f), self.script_path)
-            if os.path.isdir(os.path.join(self.script_path, d, ff)):
-                r.append('<li class="directory collapsed"><a href="#" rel="%s/">%s</a></li>' % (ff,f))
+        ex_tag = '//examples//'
```
tbekolay: I'm not sure if this is the intent, but for me this shows up literally `//examples//` in the file browser. It looks super weird... I might expect something more like a bold or italics on the folder name?

I think at some point (probably not in this PR) we should be making a more stark distinction between user generated scripts (which we can be reasonably sure are in some place that the user can modify) and our provided examples (which are likely to be in a system directory, and therefore are readonly). Something like a separate dropdown menu for examples, rather than having it be part of the same file tree browser.

tbekolay:
Code comments inline. Other comments from using these changes:

- I started `nengo_viz` in my home folder, and navigated to `Code/nengo_viz/examples/2d.py`. This shows up in the top-right corner as `././././Code/nengo_viz/examples/2d.py`, which, while technically correct, should probably be shortened. We could use `os.path.realpath` or `os.path.normpath` for that (though I'm not sure exactly where)
- Changing the port and the password via command line worked great! However, when I opened a second tab and went to the page again, I didn't get prompted for a password the second time. Is this because it's doing something smart like remembering the session, or using a cookie? I expected it to ask me for a password each time, but obviously it'd be better not to do that if it's still semi-secure.

tcstewar:
> I started nengo_viz in my home folder, and navigated to Code/nengo_viz/examples/2d.py. This shows up in the top-right corner as ././././Code/nengo_viz/examples/2d.py, which, while technically correct, should probably be shortened. We could use os.path.realpath or os.path.normpath for that (though I'm not sure exactly where)

Good point.  I'll see where I can sneak that in...  probably just in the toolbar constructor.



> Changing the port and the password via command line worked great! However, when I opened a second tab and went to the page again, I didn't get prompted for a password the second time. Is this because it's doing something smart like remembering the session, or using a cookie? I expected it to ask me for a password each time, but obviously it'd be better not to do that if it's still semi-secure.

It's a browser-level cookie, with the same sort of behaviour as normal website logins.  So when I log in to gmail and open up a separate tab, it doesn't ask me to log in again.  However, if you open up a separate browser and go to it, then it will ask you to log in again.

tbekolay:
> It's a browser-level cookie, with the same sort of behaviour as normal website logins.

Excellent! :rainbow:

nengo_viz/server.py
tcstewar: > It seems weird to me to have HTML being sent down the wire... I would expect a list of files to be sent down, and the templating engine to generate this HTML.

This is just how jQueryFileTree works:  http://www.abeautifulsite.net/jquery-file-tree/

nengo_viz/server.py
tcstewar: italics sounds good to me.  And maybe centred instead of left-justified?  I'd like the name to show up visually differently somehow so there's at least an indication that something weird is happening.  And yup, I think a separate menu for the builtin examples would be worth trying in a future PR.

tbekolay: Cool, will make an issue so I don't forget about it...

nengo_viz/server.py
tcstewar: Fixed

nengo_viz/server.py
tcstewar: > italics sounds good to me.

It's now in italics and says "built-in examples".  I tried bold and I tried centered, and those both looked weird to me, but I like this version with italics and with a non-standard directory name.

tcstewar:
> We could use os.path.realpath or os.path.normpath for that (though I'm not sure exactly where)

I'm now doing an ```os.path.relpath``` in ```Viz.load()```.  Does that work for your case?

tcstewar:
This one's been rebased again.  I think the only outstanding comment is @tbekolay 's problem with ```././././Code/nengo_viz/examples/2d.py```.  Could you check if that works better for you now?

tbekolay:
Just tried it out, and it looks great now (`Code/nengo_viz/examples/2d.py`). Good to merge!

Issue #215, opened 2015-05-28
Setting slider range to a range that does not include current slider value makes out of range slider bar


cchan987:
![image](https://cloud.githubusercontent.com/assets/10402911/7866340/d0a62ec8-053c-11e5-8a53-497298827456.png)

tcstewar:
Does #220 fix this for you, @cchan987 ?

PR #217, opened 2015-05-28
Prevent setting slider values outside of range.

### LINKED ISSUE #213: Slider value can be lowered past range bounds
Hard to describe:
If you move the slider to the bottom so it is at -1.00
then while still holding the click down, move the mouse further down, out of the slider, it will lower the value even more (lower than -1.00), while still displaying -1.

image

jgosmann:
Fixes #213, #215.


tcstewar:
> Fixes #213, #215.

Sweet.  Pretty good for a 5-character change....  :)

cchan987:
Looking good on  #213 but #215 seems still to be broken

jgosmann:
Slightly misunderstood #215, I think ...

Issue #205, opened 2015-05-27
Shut down the server when the last browser window closes


tcstewar:
We should be able to detect when there are no browser windows open, and shut down the server at that point (with perhaps a few seconds of grace time).

tbekolay:
This would be cool for sure for most use cases! I think should also be configurable though (probably a command-line option for it to stay alive, or an RC file thing). Especially for something like doing a live presentation, you might be relying on the server staying alive when opening and closing browser tabs live.

tcstewar:
+1 for configurable option.  The default would be to automatically close, though, I think (this might also be a good thing to add for the global .config file and menu)

tbekolay:
+1 auto-close by default!

PR #218, opened 2015-05-28
Shutdown server when last webpage closes

tcstewar:
Addresses #205 

The Viz object now keeps track of a list of VizSim objects.  VizSims remove themselves from that list when they are done.

When a websocket closes, it checks that list to see if it's empty.  If it is, then it shuts down the server.

For this to work, the SimControl on the server side has been modified to send out WebSocket ping events every 2 seconds (otherwise there's no way to realize the webpage has closed).

There's no configuration option yet, but I'll add that when #208 is merged.

### DIFF
--- a/nengo_viz/components/sim_control.py
+++ b/nengo_viz/components/sim_control.py
@@ -24,6 +24,7 @@ class SimControl(Component):
         self.skipped = 1
         self.time = 0.0
         self.last_status = None
+        self.next_ping_time = None
         self.reload = False
         self.send_config_options = False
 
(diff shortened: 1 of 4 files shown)


tcstewar:
Rebased this to master so it can be reviewed...

tbekolay:
I can't speak too much to the code (looks good to me?) but the user experience of this is excellent! It feels much cleaner for the server to shut down once the tabs are closed. The only thing that I would like is for an explanatory message to be printed to the console. Right now, once you close all the tabs it just prints `Shutting down server...` like before. It'd be good for it to explain why it's shutting down the server; something like `No more connections to the nengo_viz server. Shutting down server...`. As a user, I'd be impressed to see that message ;)

tcstewar:
> It'd be good for it to explain why it's shutting down the server;

I've added that message now.  I also had to change the logic slightly so that it wasn't trying to shut the server down multiple times (and thus printing multiple messages).  Now only the SimControl websocket checks for shutdowns

tbekolay:
Awesome, LGTM! I'd be keen to merge this in quickly, it's quite an improvement for user experience.

tcstewar:
> Awesome, LGTM! I'd be keen to merge this in quickly, it's quite an improvement for user experience.

Your wish is my command!  :)

studywolf:
goodness this makes things better!

PR #219, opened 2015-05-28
Styling

jgosmann:
See commit messages.

Summary:
* Fix speed and time display in sim control.
* Style the top toolbar to match sim control.

### DIFF
--- a/nengo_viz/static/viz_sim_control.css
+++ b/nengo_viz/static/viz_sim_control.css
@@ -53,6 +53,8 @@
     margin: -20px 15px 0 15px;
     position:fixed;
     width:60px;
+    border-color: #eee !important;
+    background-color: #eee !important;
 }
 
 .play-pause-button:hover {
--- a/nengo_viz/static/viz_sim_control.css
+++ b/nengo_viz/static/viz_sim_control.css
@@ -48,12 +48,16 @@
     font-family: sans-serif;
 }
 
+#pause_button {
+    border-radius:0px;
+    border-width:0px;
+}
+
 .play-pause-button {
     height:60px;
     margin: -20px 15px 0 15px;
     position:fixed;
     width:60px;
-    border-color: #eee !important;
     background-color: #eee !important;
 }
 
--- a/nengo_viz/static/viz_sim_control.css
+++ b/nengo_viz/static/viz_sim_control.css
@@ -59,15 +59,19 @@
     font-size: 30px;
 }
 
-.metrics-container.table {
+#time_table {
     display: inline-block;
-    margin-top: 40px;
+    margin-top: 10px;
     margin-left: 10px;
+    width: 130px;
     overflow: hidden;
-    transform: translate(0, -50%);
-    width: 100px;
+}
+
+.metrics-container {
+    width: auto !important;
 }
 
 .metrics-container td, .metrics-container th {
-    padding: 0.2em;
+    padding: 0.2em !important;
+    border: none !important;
 }
--- a/nengo_viz/static/viz_sim_control.css
+++ b/nengo_viz/static/viz_sim_control.css
@@ -51,8 +51,13 @@
 .play-pause-button {
     height:60px;
     margin: -20px 15px 0 15px;
-    position:fixed; 
-    width:60px; 
+    position:fixed;
+    width:60px;
+}
+
+.play-pause-button:hover {
+    background: #666 !important;
+    color: #eee !important;
 }
 
 .play-pause-button span {
--- a/nengo_viz/static/viz_sim_control.js
+++ b/nengo_viz/static/viz_sim_control.js
@@ -31,8 +31,8 @@ VIZ.SimControl = function(div, args) {
     this.ws.onmessage = function(event) {self.on_message(event);}
     
     /** Create the TimeSlider */
-    this.time_slider = new VIZ.TimeSlider({x: 140, y: 10, sim:this,
-                                           width: this.div.clientWidth-240, 
+    this.time_slider = new VIZ.TimeSlider({x: 150, y: 10, sim:this,
+                                           width: this.div.clientWidth-250, 
                                            height: this.div.clientHeight-20,
                                            shown_time: args.shown_time,
                                            kept_time: args.kept_time});
--- a/nengo_viz/static/viz_top_toolbar.css
+++ b/nengo_viz/static/viz_top_toolbar.css
@@ -2,6 +2,7 @@
     color: #333;
     cursor: default;
     top: 0;
+    border-radius: 0;
 }
 
 #top_toolbar .glyphicon:hover {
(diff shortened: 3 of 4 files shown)


tcstewar:
Nice.   :)

I do find the blue highlighting on hover a little too subtle....   perhaps a border around the button on hover?

```
#top_toolbar .glyphicon {
    color: #333;
    cursor: default;
    top: 0;
    border: solid 1px #eee;
}

#top_toolbar .glyphicon:hover {
    background: #d9edf7;
    border: solid 1px #ccc;
}
```

jgosmann:
Oh yeah, I forgot to mention that we might want to make the highlighting a bit less subtle.
I chose the color because it matches the color in the time bar ...

jgosmann:
Not sure how much I like the border. I changed it to invert fore- and background color on hover. What do you think about that?

tcstewar:
> Not sure how much I like the border. I changed it to invert fore- and background color on hover. What do you think about that?

Interesting..  it's a bit harsh of a contrast, but I like the general idea.  Maybe less dark a background on hover?

Also, I think the play/pause/build button in the bottom-right should by styled similarly to the top toolbar buttons.

jgosmann:
Reduced the contrast and applied the same hover style to the play/pause button. (I don't think the normal style should be changed, because the brighter background makes it more easy to identify as a clickable button. Otherwise it would have the same color as the background.)

celiasmith:
+1 to both of terry's comments

celiasmith:
now it doesn't look the same as the other buttons.... i think the point was to make it look like those other ones (no background)

celiasmith:
not sure if this is styling as well, but the filename cursor changes (to a text selector, which you can't do)... either you should be able to select the text or the cursor shouldn't change.

jgosmann:
Filename cursor should be fixed with the last commit.
As I said, I don't agree that the play/pause button should have no background.

celiasmith:
Do you want to make them all have a background then?  Or do you think they're good just being different?

celiasmith:
The cursor is still not fixed for me.

celiasmith:
I just did a commit to see what it looks like to make all the buttons the same.  Thoughts of this vs having the bottom one different?

tbekolay:
One style thing about the top toolbar that I really dislike is the rounding on the toolbar buttons. It makes it seem like they're not actually supposed to be lined up beside each other, because there's a weird bit of negative space were they all to be hovered on at the same time.

Setting border-radius to 0px, or none, should fix it.

studywolf:
i liked the blue,... it looks so boring as light grey...

celiasmith:
I thought that, but all the colours come from data!  so then i felt better...

studywolf:
data?

celiasmith:
run the model and the graphs are all coloured :)

jgosmann:
> The cursor is still not fixed for me. I think it should be 'default' instead of 'pointer'

Depends on what the cursor should be. I think, when I tested it first it was 'pointer' and I like it that way. But if people prefer 'default', that's good too.

> I just did a commit to see what it looks like to make all the buttons the same. Thoughts of this vs having the bottom one different?

I don't like it. It's common to have toolbars without visible button borders, so it's obvious enough that icons in a row at the top of a window are clickable. But this is not as obvious for the sim control. It could just be a text marker direction of time. And I can't think of any software where non-toolbar buttons don't have any border.

> i liked the blue,... it looks so boring as light grey...

I don't think control elements like the toolbar should be too brightly colored because that's not what the focus of attention should be. Having some more color in there could be nice, but we should be careful where we add it. One thing I could imagine is to use a different hover background color.

jgosmann:
and I removed the rounded corners from the toolbar buttons

studywolf:
Color can be a settings thing!

jgosmann:
We could make everything styleable with user CSS. :P

As a general note: I don't think, we should make every conceivable thing a user option.

tcstewar:
> As a general note: I don't think, we should make every conceivable thing a user option.

I'd second that note...  it's the same sort of principle as Python's "There should be one-- and preferably only one --obvious way to do it."  :)

studywolf:
I don't think anyone is advocating for every conceivable option, friends.
Color is clearly a preference and rather than argue about it i thought we
could have an easy fix where users can set it however they want.

As a general note: I don't think, we should make every conceivable thing a
user option.

I'd second that note... it's the same sort of principle as Python's "There
should be one-- and preferably only one --obvious way to do it." :)

—
Reply to this email directly or view it on GitHub
<https://github.com/nengo/nengo_viz/pull/219#issuecomment-106621893>.

tcstewar:
> Color is clearly a preference and rather than argue about it i thought we
> could have an easy fix where users can set it however they want.

True, but even if we did have such an option, we'd still have to have exactly this argument over what the default should be.  And once we've decided on a default, I'm not sure we gain much by having it be configurable.  Mostly I'm trying to avoid the "giant page of configuration options" problem that's prevalent in other software, so I don't think we should get in the habit of creating new configuration options.  I'd much prefer to decide on a look and feel and stick with that.

That said, I really shouldn't be allowed a vote on any colour-based look and feel options.  :)

celiasmith:
(Hopefully not a duplicate comment, as I emailed it, but it's not showing up here)

On button borders: The email reader I'm using now has no borders on both top and bottom bar buttons.  This is very common in mobile. If you google flat ui design you'll see many examples.

Adding: On the cursor, the original change to 'pointer' didn't do anything (which is why I suggested default), the commit I added fixed that problem (it was different css that had to change).  As for pointer vs default, I suspect we need to compare these across OSs, 'pointer' on a mac is a finger...

On colours, we should talk in person.  My impression is that Linux apps tend to have a .settings file where you can change everything about the app (often not in a menu); not so for windows and mac, where look and feel is more constrained.  User css would kind of do the linux thing, for better or worse.

jgosmann:
The examples I find on Google seem to support my view. But I think, I explained my view and if the majority prefers no border, I'm fine with that.

Introducing settings for colors could lead to a vast amount of options and might be wasted developer time. How many users will actually care enough to change those options? Allowing user CSS seems to be the best option to me, because it should be fairly easy to implement and gives all the flexibility. But before that we might want to clean up the HTML structure, IDs, and class names a bit to have a clear structure for styling.

That being said, I think there is one case where configurable colors of some sort are quite important: The syntax highlighting in the editor. And while I'm talking about it: I propose [Solarized](http://ethanschoonover.com/solarized) (which comes in a dark and a light version) as default.

studywolf:
Sure, in which case I vote for a darker gray or ideally a color for the top
bar. It kind of blends into the white background at the moment and that is
a bit headache-y.

I would actually think that the syntax highlighting would be a case where
the configuration wouldn't matter really, since it's also designed for
people to use their own editor instead...

jgosmann:
> since it's also designed for people to use their own editor instead...

I guess, that's true

tcstewar:
Cool, this is definitely an improvement now, so I'm going to merge it.  I'm quite sure there will be more styling issues for future PRs....  :)

PR #221, opened 2015-05-28
Embedded visualization in IPython notebook

jgosmann:
Opening this PR for people to take a look at and try out. It seems to be working quite well already, but there are a few more things which should maybe addressed.

This allows to embed the nengo_viz visualization easily in IPython notebooks. Just do
```python
from nengo_viz.ipython import IPythonViz
IPythonViz(model)
```

Assigning the `IPythonViz` object to a variable and displaying it multiple times should work (it's like opening multiple tabs for nengo_viz). Different `IPythonViz` objects should be able to coexist.

Potential things to do:
* [x] On shutdown the server produces output which will be displayed the next time a cell is executed after server shutdown. Probably that output should be suppressed.
* [x] There always seems to be Zombie threads remaining after shutdown. Maybe the timeout is just too short? :point_right: Solved in PR #227.
* [ ] Do we want to automatically display model with nengo_viz in IPy Notebooks? (without having it to enclose in `IPythonViz(...)`? :point_right: We probably don't have to decide this now and can leave it for a potential later PR.
* [x] The config should probably live longer than just one IPythonViz lifecycle ... but I'm not sure whether one config per IPython kernel is the correct approach. There might be different models defined in one kernel? :point_right: It is now possible to pass a filename to persist the config. I don't think there is a way to figure out which config to use without this additional information because we don't know whether two models are supposed to be the same.
* [x] The file open toolbar button should probably removed for the embedded visualization. (It should work fine, but might be confusing.)
* [x] The filename in the toolbar isn't helpful either, and actually non-existent.
* [x] something related to the server side, but I forgot :point_right: I believe it was about what I commented on in f97b743a10f057290741aca1befa9d6d47a8826d. If not, it probably wasn't that important.

### DIFF
--- a/nengo_viz/components/netgraph.py
+++ b/nengo_viz/components/netgraph.py
@@ -27,15 +27,21 @@ class NetGraph(Component):
         self.parents = {}
         self.networks_to_search = [self.viz.model]
         self.initialized_pan_and_zoom = False
-        self.last_modify_time = os.path.getmtime(self.viz.viz.filename)
+        try:
+            self.last_modify_time = os.path.getmtime(self.viz.viz.filename)
+        except:
+            self.last_modify_time = None;
         self.last_reload_check = time.time()
 
     def check_for_reload(self):
-        t = os.path.getmtime(self.viz.viz.filename)
+        try:
+            t = os.path.getmtime(self.viz.viz.filename)
 
-        if self.last_modify_time < t:
-            self.reload()
-            self.last_modify_time = t
+            if self.last_modify_time < t:
+                self.reload()
+                self.last_modify_time = t
+        except:
+            pass
 
     def reload(self):
         locals = {}
--- a/.gitignore
+++ b/.gitignore
@@ -57,6 +57,7 @@ target/
 
 # nengo_viz specific
 *.py.cfg
+.ipynb_checkpoints
 
 # Vagrant
 .vagrant
--- a/examples/net.ipynb
+++ b/examples/net.ipynb
@@ -1,0 +1,62 @@
+{
+ "metadata": {
+  "name": "",
+  "signature": "sha256:449f5777c8db5b6d5daf37b3d65fc3848ca984f8862ebb50760a896deda5adea"
+ },
+ "nbformat": 3,
+ "nbformat_minor": 0,
+ "worksheets": [
+  {
+   "cells": [
+    {
+     "cell_type": "code",
+     "collapsed": false,
+     "input": [
+      "import numpy as np\n",
+      "import nengo\n",
+      "\n",
+      "model = nengo.Network()\n",
+      "with model:\n",
+      "    stimulus_A = nengo.Node([1], label='stim A')\n",
+      "    stimulus_B = nengo.Node(lambda t: np.sin(2*np.pi*t))\n",
+      "    ens = nengo.Ensemble(n_neurons=1000, dimensions=2)\n",
+      "    result = nengo.Ensemble(n_neurons=50, dimensions=1)\n",
+      "    nengo.Connection(stimulus_A, ens[0])\n",
+      "    nengo.Connection(stimulus_B, ens[1])\n",
+      "    nengo.Connection(ens, result, function=lambda x: x[0] * x[1],\n",
+      "                     synapse=0.01)\n",
+      "\n",
+      "    with nengo.Network(label='subnet') as subnet:\n",
+      "        a = nengo.Ensemble(100, 1)\n",
+      "        b = nengo.Ensemble(100, 1)\n",
+      "        nengo.Connection(a, b)\n",
+      "        nengo.Connection(b, b)\n",
+      "\n",
+      "        with nengo.Network() as subsubnet:\n",
+      "            c = nengo.Ensemble(100, 1)\n",
+      "            d = nengo.Ensemble(100, 1)\n",
+      "            nengo.Connection(c, d)\n",
+      "        nengo.Connection(b, c)\n",
+      "        nengo.Connection(d, a)\n",
+      "    nengo.Connection(result, a)"
+     ],
+     "language": "python",
+     "metadata": {},
+     "outputs": []
+    },
+    {
+     "cell_type": "code",
+     "collapsed": false,
+     "input": [
+      "from nengo_viz.ipython import IPythonViz\n",
+      "IPythonViz(model)"
+     ],
+     "language": "python",
+     "metadata": {},
+     "outputs": []
+    }
+   ],
+   "metadata": {}
+  }
+ ]
+}
--- a/nengo_viz/components/sim_control.py
+++ b/nengo_viz/components/sim_control.py
@@ -125,8 +125,6 @@ class SimControl(Component):
                 os.remove(self.viz.viz.filename + '.cfg')
             self.viz.viz.load(self.viz.viz.filename)
             self.reload = True
-        else:
-            print(msg)
 
 class SimControlTemplate(Template):
     cls = SimControl
--- a/nengo_viz/components/sim_control.py
+++ b/nengo_viz/components/sim_control.py
@@ -5,6 +5,7 @@ import traceback
 import numpy as np
 import nengo
 import os
+import os.path
 import json
 
 from nengo_viz.components.component import Component, Template
@@ -121,9 +122,12 @@ class SimControl(Component):
             except:
                 traceback.print_exc()
         elif msg == 'reset':
-            if os.path.isfile(self.viz.viz.filename + '.cfg') :
-                os.remove(self.viz.viz.filename + '.cfg')
-            self.viz.viz.load(self.viz.viz.filename)
+            if os.path.isfile(self.viz.viz.config_name()) :
+                os.remove(self.viz.viz.config_name())
+            self.viz.viz.config = self.viz.viz.load_config()
+            self.viz.viz.load(
+                self.viz.viz.filename, self.viz.viz.model,
+                self.viz.viz.orig_locals)
             self.reload = True
 
 class SimControlTemplate(Template):
--- a/nengo_viz/ipython.py
+++ b/nengo_viz/ipython.py
@@ -56,11 +56,12 @@ class IPythonViz(object):
     def start_server(cls, cfg, model):
         name = model.label if model.label is not None else ''
         viz = nengo_viz.Viz(
-            name, cfg=cfg, model=model, locals=get_ipython().user_ns)
+            name, cfg=cfg, model=model, locals=get_ipython().user_ns,
+            interactive=False)
         server = viz.prepare_server(port=0, browser=False)
         server_thread = threading.Thread(
             target=viz.begin_lifecycle,
-            kwargs={'server': server, 'interactive': False})
+            kwargs={'server': server})
         server_thread.start()
         cls.servers[cfg] = server
         cls.threads[cfg] = server_thread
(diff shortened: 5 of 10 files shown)


jgosmann:
This PR would supersede nengo/nengo#147.

tbekolay:
Whoa, this sounds super cool!!

tcstewar:
What the....   how the hell is this even possible!   

This is way cool.  Way way way cool.  Still some rough edges, but wow is this incredible.  It didn't even cross my mind that this would be doable.

jgosmann:
> What the.... how the hell is this even possible!

iframes ... and some luck that IPython exposes a dictionary of user defined variables.

jgosmann:
This PR also supersedes nengo/nengo#524 (that's actually the one I wanted to link in the first place)

jgosmann:
Updated the issue list above.

jgosmann:
This is probably good for review now.

nengo_viz/components/netgraph.py
```diff
@@ -27,15 +27,21 @@ def __init__(self, viz, config, uid):
...
+            t = os.path.getmtime(self.viz.viz.filename)
 
-        if self.last_modify_time < t:
-            self.reload()
-            self.last_modify_time = t
+            if self.last_modify_time < t:
+                self.reload()
+                self.last_modify_time = t
+        except:
+            pass
```
tcstewar: I'd much prefer if this didn't fail silently like this.  Is this meant to dealing with the filename not existing, or any possible error in the reload process?

jgosmann: With the filename not existing (because there is no physical file of the model for the IPython notebook). But also in the regular case, it would be better if the server continues running if the file is deleted.

tcstewar:
This is really quite cool.  :)  

I'm occasionally getting an error on load (so far always after re-executing a notebook block) on this line:

https://github.com/nengo/nengo_viz/blob/master/nengo_viz/viz.py#L251

The core thing is a bug on my part in that if ```label``` is None, then I can't look for a ```'.'``` in it, so I should be checking for None before doing that search.  But I'm not sure why it's coming up at all, since ```obj``` should always be in ```default_labels```.   But I'm having problems replicating the bug, so I'm not sure yet.

nengo_viz/components/netgraph.py
tcstewar: Hmm..  but with this, it also silently ignores any exceptions that happen in ```reload``` that aren't due to the file suddenly going missing, which may make it more difficult to debug.  Perhaps if it was just catching ```OSError```?

tcstewar:
Also, it'd be great in the future if there was some way for it to maintain a .cfg file for the layout, but I think that's a separate PR.

jgosmann:
Just found two more problems:
* Reset is not working.
* If a model is displayed twice with the same config file and updated in-between those changes won't be shown. :/ Not sure what to do about it. Maybe the existing server should not be reused, but than two servers might write to the same config ...

jgosmann:
And there is again some output which should be hidden ... at some point I would like to replace all the print statements with something where it is easier to control where that output appears (if at all).

jgosmann:
The last commit might have solved your problem, @tcstewar.
Apart from that the commit makes sure that each visualization uses a different config. If the same filename is past twice only the last started visualization will update the config file (the first time this happens a warning will be printed).

jgosmann:
Problems should be fixed.

tcstewar:
> Problems should be fixed.

Haven't been able to make it crash like that again since you made those changes.  Yay!

tcstewar:
One thing I just realized about the config file...  if we're never going to load the config file, we shouldn't actually have to make it.  I'm highly tempted to put some logic in to Viz such that if ```filename is None``` it just doesn't bother even trying to save anything to the file.  Everything should still work without that, though, since there's a copy held in memory.

jgosmann:
At the moment I pass the network label as filename (if it's defined) to have that show up in the top right corner. But that logic could be moved to core nengo_viz to do the `filename is None` thing (which might be a good idea).

Just not sure, whether we should do that before the release.

tcstewar:
> But that logic could be moved to core nengo_viz to do the filename is None thing (which might be a good idea).
> 
> Just not sure, whether we should do that before the release.

I'm good to put it off till the next version...  :)

jgosmann:
So, is this then good to merge? :)

tbekolay:
I got into a weird situation when I reran a cell with the visualizer... the netgraph was normal sized, but the sliders and graphs were huge!! I zoomed out, removed them, and then when I zoomed in this happened:

screen shot 2015-06-03 at 6 02 10 pm

Not a deal-breaker I think; honestly it's crazy but super cool that this works ;) I think we can just note that this may be slightly buggy, and have people file issues when they come up.

jgosmann:
Did you try the reset button? I just discovered that it's quite broken in this branch.

jgosmann:
For some reason graphs and sliders won't get removed ... but I don't see why.

jgosmann:
Found out why and fixed it.

tbekolay:
I did try the reset button, so that might have done it. I checked this out and now reset works perfectly!

After running it for a while, I re-evalated the same cell, and when nengo_viz loaded up again it was basically in an inactive state. I couldn't move any objects, couldn't press play, and when I tried to press reset, the modal popped up but I couldn't press reset. Seemed like the server wasn't runing perhaps?

jgosmann:
Yeah, that seems like the server died ...
If you evaluate another cell afterwards, it might show you an exception that was raised (if any).

tbekolay:
It has in the past, but not this time. In any case, I think it's fine for now! Any other outstanding issues for this PR?

jgosmann:
I think, the problem is that I connect to the server to see whether it has started and then I disconnect which shuts down the server (I thought the server would only consider websocket connections ...)

tcstewar:
> I think, the problem is that I connect to the server to see whether it has started and then I disconnect which shuts down the server (I thought the server would only consider websocket connections ...)

Would it help if the server had an option as to whether or not to auto-shutdown?  I was strongly tempted to toss that in anyway....

jgosmann:
Actually, it should auto-shutdown, otherwise we could end up with a lot of server instances.

jgosmann:
:rage: The whole server organization needs to be refactored!!1one! If I'm not mistaken, **all** servers will be shutdown once no connections remain and that might include a server which is just starting up.

tcstewar:
> The whole server organization needs to be refactored!!1one! If I'm not mistaken, all servers will be shutdown once no connections remain and that might include a server which is just starting up.

Argh.  Yes, I think you're right here.  I'm pretty sure we've never worried about the case where there are multiple Servers at the same time.

Can you wait to start the new Server until the previously existing ones have finished shutting down?

jgosmann:
No, because I don't have any idea when they will shut down or if they will shut down at all

tcstewar:
> No, because I don't have any idea when they will shut down or if they will shut down at all

Hmm...  how about only ever creating one server?  When a second one is needed, we can re-use the existing one the same way multiple browser windows work (one Viz, multiple VizSims)

jgosmann:
One server can't handle more than one model, but we might have more than one model.

tbekolay:
Could the IPython stuff keep track of the last port used, and just increment it for each time you call `IPythonViz`?

tcstewar:
> One server can't handle more than one model, but we might have more than one model.

It'll work as well as multiple models in different browser windows works anyway (which isn't well, I admit).

I should be able to fix some of that behaviour though...  I'll see what I can do.

tcstewar:
> Could the IPython stuff keep track of the last port used, and just increment it for each time you call IPythonViz?

It's not a port problem.  It's the problem that the nengo_viz.Server is only expecting one instance of nengo_viz.Viz to exist.

jgosmann:
I think, it's fixed for now.

tbekolay:
Cool; it works in that I can rerun a cell and it'll bring up a nengo_viz window that works!

Here's something weird though... let me know if it's reproducible. I run the `cominbing` Nengo example, add a value plot for '2D Population', and run. Works great. Rerun the cell, and now the view is reset, and it doesn't look like the value plot is there anymore. But it is there; in fact, it's just super gigantic, and takes up many times the screen real estate. If I zoom out a lot, I can then see the lines being drawn, but the netgraph is tiny at this point.

I'd prefer to not try to save the config than to have weird stuff like this happening... is it a quick fix?

jgosmann:
Where does the server store those plots ...?

jgosmann:
Arg, the global namespace gets affected ...

jgosmann:
But that was an easy fix. :)

tbekolay:
Great, I can now rerun this cell as many times as I want! :rainbow: Seems ready to merge IMO, any dissenting opinions?

Also, I can't overstate how cool this is. Minds will be blown when we do this at the summer school ;)

jgosmann:
> Minds will be blown when we do this at the summer school ;)

Good thing they'll learn how to build a new one. ;)

jgosmann:
I will do the rebase now.

jgosmann:
rebased

tbekolay:
@celiasmith was having an issue with this branch, but it seems like probably an environment issue (maybe an IPython version issue?) Will merge when we get it working on his machine.

jgosmann:
What kind of issue?

tbekolay:
I rebased to master again; no conflicts, but now I get this error:

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-2-644f47eb8289> in <module>()
      1 from nengo_viz.ipython import IPythonViz
----> 2 IPythonViz(model)

/Users/Trevor/Code/nengo_viz/nengo_viz/ipython.pyc in __init__(self, model, cfg, height)
     35             cfg = get_ipython().mktempfile()
     36 
---> 37         self._server_thread, server = self.start_server(cfg, model)
     38         self.port = server.server_port
     39 

/Users/Trevor/Code/nengo_viz/nengo_viz/ipython.pyc in start_server(cls, cfg, model)
     66         viz = nengo_viz.Viz(
     67             name, cfg=cfg, model=model, locals=get_ipython().user_ns,
---> 68             interactive=False, allow_file_change=False)
     69         server = viz.prepare_server(viz, port=0, browser=False)
     70         server_thread = threading.Thread(

/Users/Trevor/Code/nengo_viz/nengo_viz/viz.pyc in __init__(self, filename, model, locals, cfg, interactive, allow_file_change)
    141                                     'default.py')
    142 
--> 143         self.load(filename, model, locals)
    144 
    145     def load(self, filename, model=None, locals=None):

/Users/Trevor/Code/nengo_viz/nengo_viz/viz.pyc in load(self, filename, model, locals)
    144 
    145     def load(self, filename, model=None, locals=None):
--> 146         filename = os.path.relpath(filename)
    147         if locals is None:
    148             locals = {}

/Users/Trevor/.virtualenvs/nengo/lib/python2.7/posixpath.pyc in relpath(path, start)
    433 
    434     if not path:
--> 435         raise ValueError("no path specified")
    436 
    437     start_list = [x for x in abspath(start).split(sep) if x]

ValueError: no path specified
```

Also, I pushed a commit that adds `net.ipynb` to the examples folder. Helps for testing out this branch.

jgosmann:
wil look into it

jgosmann:
fixed it, also fixed @celiasmith config loading problem.

tbekolay:
Okay cool; works for me now, so merging!

Issue #101, opened 2015-04-14
Requires `if __name__ == '__main__'`


hunse:
@cchan987 told me to put this at the end of my file to run the visualizer

``` python
if __name__ == '__main__':
    import nengo_viz
    nengo_viz.Viz(__file__).start()
```

but if I just put this

``` python
import nengo_viz
nengo_viz.Viz(__file__).start()
```

it doesn't work. I get: 

```
Traceback (most recent call last):
  File "test_split.py", line 39, in <module>
    nengo_viz.Viz(__file__).start()
  File "/home/ehunsber/workspace/nengo_viz/nengo_viz/viz.py", line 231, in __init__
    exec(code, locals)
  File "<string>", line 39, in <module>
NameError: name '__file__' is not defined
```

tbekolay:
I also ran into this issue, and it's not an easy one to debug. FYI the reason is because `nengo_viz` takes the file pointed to in `__file__`, reads it into a string, and `exec`s it. `__file__` exists when you run `python my_file.py`, and is `my_file.py`, but not when the file is `exec`d.

Ideally, we would either

- Make it possible to start visualizing a network without having to re-execute that network's file (in general, it would be nice to not have to have the file requirement, e.g., for starting a visualization in an IPython notebook).
- Provide a simple script and recommend starting the visualizer by running that script; i.e., `nengo_viz my_file.py`. This can be done just by editing `setup.py`.

tcstewar:
> Make it possible to start visualizing a network without having to re-execute that network's file (in general, it would be nice to not have to have the file requirement, e.g., for starting a visualization in an IPython notebook).

That's doable right now.  You do it this way:

```python
nengo_viz.Viz(file='whatever.py', locals=locals())
```

the only thing it uses ```file``` for in this case is to generate a file name for the ```.cfg``` file.  If you pass in ```locals``` it uses that, otherwise it executes ```file``` to generate the locals.



> Provide a simple script and recommend starting the visualizer by running that script; i.e., nengo_viz my_file.py. This can be done just by editing setup.py

That is definitely the right thing to have, and it would be exactly like how ```nengo_gui``` does it now.

One other thing that I would very much like to add is to get rid of the ```if __name__ == '__main__':``` requirement.  That would be done by adding some hooks into the code that is executing the script to have it stop if it notices a nengo_viz.Viz object being created.  It should also stop if a nengo.Simulator object is created.

tbekolay:
> That's doable right now.

[...]

PR #224, opened 2015-05-28
Handle loading bad files

tcstewar:
This now handles some of the cases of weirdnesses that can happen when loading

- Any exception is printed to the console and we recover gracefully.  In a future PR this exception should be sent to the browser and displayed somehow
- If the script creates a Simulator object (either nengo.Simulator or nengo_ocl.Simulator), it stops executing the script at that point (before build)
- If the script creates a Viz object, it stops executing at that point

This means we no longer need ```if __name__ == '__main__':``` protection in the example scripts (issue #101).

### DIFF
--- a/nengo_viz/components/sim_control.py
+++ b/nengo_viz/components/sim_control.py
@@ -1,5 +1,6 @@
 import time
 import struct
+import traceback
 
 import numpy as np
 import nengo
@@ -115,11 +116,11 @@ class SimControl(Component):
                 self.viz.rebuild = True
             self.paused = False
         elif msg[:4] == 'open':
-            if os.path.isfile(self.viz.viz.filename):
+            try:
                 self.viz.viz.load(msg[4:])
                 self.reload = True
-            else:
-                print('File does not exist')
+            except:
+                traceback.print_exc()
         elif msg == 'reset':
             if os.path.isfile(self.viz.viz.filename + '.cfg') :
                 os.remove(self.viz.viz.filename + '.cfg')
--- a/nengo_viz/monkey.py
+++ b/nengo_viz/monkey.py
@@ -1,17 +1,21 @@
+import contextlib
 import importlib
 import threading
-import sys
 import traceback
-import contextlib
+import sys
+
 
 # list of Simulators to check for
-known_modules = ['nengo', 'nengo_ocl']
+known_modules = ['nengo', 'nengo_ocl', 'nengo_brainstorm', 'nengo_distilled']
+
 
 class StartedSimulatorException(Exception):
     pass
 
+
 class StartedVizException(Exception):
     pass
+
 
 # create a wrapper class that will throw an exception if we are
 # currently executing a script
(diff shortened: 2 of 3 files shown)


tcstewar:
Rebased this to master so it can be reviewed...  :)

tbekolay:
Hm... two things jump out at me; not sure if I'm interpreting the code right, so definitely correct if I'm wrong!

First, it looks like the code will stop the moment that it encounters a `*.Simulator` call or a `nengo_viz.Viz` call? There's a builtin assumption here that the Simulator or nengo_viz stuff will be at the end of the script, and the model will be completely built by that point. I don't think that that will always be the case causing very strange bugs where half of someone's model will be there in the GUI.

As for a fix, I don't think we should re-implement `eval`, that would be cray. Previously I suggested just stripping lines from the file after it's been read in (or passed in) but I agree that that's also quite an error prone process that could easily miss things. So, I think for the time being we should at least raise a warning or print something to the console letting the user know that we stopped evaluating the model at a certain point. It would be great if we could actually say something "We encountered a Simulator at line 34 in mymodel.py. Ignoring this line, and all subsequent lines. If your model was complete at this point, then you can ignore this message." I think we have access to that info...

Second, the monkey patching in this seems a bit heavy handed. I definitely wouldn't expect anything to be money patched when I do `import nengo_viz`. But once I actually hand control over to `nengo_viz`, then it can do whatever.

If I'm reading the code correctly, the only time that the monkey patching is actually necessary is in the `Viz.load` function. If that's the case, then I suggest we do the monkey patch in a context manager, so the load looks something like this:

```python
def load(self, filename, model=None, locals=None):
    with nengo_viz.monkey.patch_sims():
        ... do the loading stuff ...
```

The context manager would allow us to return `nengo.Simulator` and `nengo_ocl.Simulator` to its original state after loading the file.

tcstewar:
> It would be great if we could actually say something "We encountered a Simulator at line 34

Ooo, excellent idea.  And yup, that's easy info to get -- we can just ask the exception for its stack trace and check its line number.  I'll sort that out...

tcstewar:
> The context manager would allow us to return nengo.Simulator and nengo_ocl.Simulator to its original state after loading the file.

Hmm...  I could see something like that working in addition to what there is here.  My only worry is that the monkey-patch effects will still bleed over across threads, so there's still the potential for confusing object changes.  It might be possible to muck with the locals in ```load``` enough such that there's no way any other thread might get access to these monkey-patched objects, though.  I'll poke at it more....  :)

tcstewar:
There's now a message printed to the console letting us know we're ignoring everything after a Simulator or Viz is created in the script.  In a future PR, this should be sent to the javascript client.

tcstewar:
> The context manager would allow us to return nengo.Simulator and nengo_ocl.Simulator to its original state after loading the file.

It's now a context manager.  :)  This helps clean it up nicely, and gets rid of the majority of cases where the monkey patch will confuse things.

It is still possible for another thread to get a DummySimulator object, if it creates a Simulator while this thread is in the midst of executing a file.  However, if it does so, that DummySimulator object should still work fine, since that other thread doesn't have its flag set.

tbekolay:
Cool cool, I'm not too worried about other threads accessing that stuff, as that's either rare or advanced usage, and in most cases it'll still work. Aside from the inline comments, this PR looks good to me!

PR #225, opened 2015-05-28
Make Modal a class instead of singleton

tbekolay:
While we won't do this, it makes it possible to make multiple modals. The major advantage here is just that the Javascript code looks more idiomatic.

The main reason I did this was because info_body was being used to clear the body, which is overkill (and adds divs we don't want). So I also added a clear_body function. And a text_body function, which the reset modal now uses (it is resettable right?)

I know that this'll make merging #186 and #195 a bit more complicated, but the history of those branches needs to be fixed up anyhow (which I'm happy to do!)

### DIFF
--- a/nengo_viz/disposable_js.py
+++ b/nengo_viz/disposable_js.py
@@ -20,18 +20,18 @@ def infomodal(ng, uid, **args):
         raise NotImplementedError()
 
 def add_modal_title_js(title_text):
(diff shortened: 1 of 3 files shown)


tbekolay:
FYI I just force pushed this branch to change the first commit, since I realize now that removing the config with the reset button in the toolbar is not, in fact, resettable (yet!) I also fixed some CSS stuff in another commit, as the modals' icon colors were messed up.

PR #226, opened 2015-05-28
Normalize mouse scroll speeds across browsers

celiasmith:
It’d be nice to have the scroll speed set in config.  

This commit definitely makes the mac sensitivity issue better.  It makes scrolling consistent on firefox, safari, and chrome from what I can tell.

Solution from: http://stackoverflow.com/questions/5527601/normalizing-mousewheel-speed-across-browsers

### DIFF
--- a/nengo_viz/static/viz_netgraph.js
+++ b/nengo_viz/static/viz_netgraph.js
@@ -114,9 +114,7 @@ VIZ.NetGraph = function(parent, args) {
             var x = (event.clientX / $(self.svg).width())
             var y = (event.clientY / $(self.svg).height());
 
-            self.normalize_mousewheel(event);
             var delta = event.wheelDeltaY || -event.deltaY;
-
             var scale = delta > 0 ? VIZ.scale.step_size : 1.0 / VIZ.scale.step_size; // will either be 1.1 or ~0.9
 
             //scale and save components
@@ -374,4 +372,3 @@ VIZ.NetGraph.prototype.detect_collapsed_conns = function(uid) {
         }
     }
 }
-
--- a/nengo_viz/static/viz_netgraph.js
+++ b/nengo_viz/static/viz_netgraph.js
@@ -114,7 +114,7 @@ VIZ.NetGraph = function(parent, args) {
             var x = (event.clientX / $(self.svg).width())
             var y = (event.clientY / $(self.svg).height());
 
-            this.normalize_mousewheel(event);
+            self.normalize_mousewheel(event);
             var delta = event.wheelDeltaY || -event.deltaY
             VIZ.scale.step_size = 1.05
             var scale = delta > 0 ? VIZ.scale.step_size : 1.0 / VIZ.scale.step_size; // will either be 1.1 or ~0.9
--- a/nengo_viz/static/viz_netgraph.js
+++ b/nengo_viz/static/viz_netgraph.js
@@ -8,9 +8,9 @@
  */
 VIZ.NetGraph = function(parent, args) {
     this.scale = 1.0;          // global scaling factor
-    this.offsetX = 0;          // global x,y pan offset 
+    this.offsetX = 0;          // global x,y pan offset
     this.offsetY = 0;
-    
+
     this.svg_objects = {};     // dict of all VIZ.NetGraphItems, by uid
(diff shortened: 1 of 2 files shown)


jgosmann:
There's a [deltaMode property on the wheel event](https://developer.mozilla.org/en-US/docs/Web/Events/wheel) is that set to pixels in all cases? Maybe that explains some of the differences between browsers ... but I don't have much hope. It's strange that the delta values are such a mess across browsers. :cry: 

Furthermore, does this make db136af32dcb869ae2f1b3b741648bdcf96e99c4 obsolete? Or is that still needed?

celiasmith:
No this doesn't make that obsolete.  That other commit stops the 'intertia' that's a bit nuts on macs. (Does it make other OSs jerky, though?  If so, maybe we can just do it for macs?)  This one makes the zooming the same sensitivity across browsers.

nengo_viz/static/viz_netgraph.js
```diff
@@ -114,7 +114,9 @@ VIZ.NetGraph = function(parent, args) {
             var x = (event.clientX / $(self.svg).width())
             var y = (event.clientY / $(self.svg).height());
 
+            self.normalize_mousewheel(event);
             var delta = event.wheelDeltaY || -event.deltaY
+            VIZ.scale.step_size = 1.05
             var scale = delta > 0 ? VIZ.scale.step_size : 1.0 / VIZ.scale.step_size; // will either be 1.1 or ~0.9
```
tcstewar: I'm confused...  it looks to me like ```normalize_mousewheel``` does some wacky math in order to compute a nice ```event.delta```, which we then throw away because this ```scale``` calculation completely ignores it.

tcstewar:
Hmm, this makes scrolling unbearably slow on Windows (Chrome and Firefox).  :(

celiasmith:
Could you try this out... hopefully it speeds up scrolling on non Macs (and keeps it good for macs).  I've tested it across all 3 browsers.  And in the end, the original method is better than the new event normalizing i tried, so i switched it back.  So, really, this is just scaling the speed differently across mac vs non-mac

tcstewar:
> Could you try this out... hopefully it speeds up scrolling on non Macs

That's good on Chrome and Firefox on WIndows for me...  :)

celiasmith:
Sweet, then it should be good to go

tcstewar:
> Sweet, then it should be good to go

Sounds good to me.  Although we should probably post an Issue saying that at some point this should get revisited, as we really are throwing out a lot of information here.  What's controlling the speed of the zoom is the rate at which the browser generates mousewheel events, not anything to do with the data in the mousewheel even that says how big the motion should be (the only thing we look at is the sign of this value).

Note that this is now a 3-line change, but requires rebasing, so you may want to just do it in a totally new branch.  ;)  Or do the rebase.

jgosmann:
Scrolling seems consistent across browsers on OS X. But to me it feels not sensitive enough with a mouse wheel. The trackpad is ok, but could also be a bit more sensitive.

The feel of the inertia on the trackpad doesn't feel right, but will be probably hard to get right anyways.

jgosmann:
On Windows when accessed via remote desktop a sound is played occasionally while zooming. :open_mouth: It doesn't seem to happen without the remote desktop connection.

Scrolling feels too insensitive for both trackpad and mouse on Windows, but take this with a grain of salt as I was using a virtual machine which might mess with the events from the hardware.

celiasmith:
>Note that this is now a 3-line change, but requires rebasing, so you may want to just do it in a totally new branch. ;) Or do the rebase.

Tried the rebase... got ugly... moved to a new PR #240

celiasmith:
@jgosmann That's super weird (sounds).  On macs, I agree it's good on trackpads.  I find the sensitivity a bit much on the mouse (magic mouse), myself, so this is probably a good compromise (since we don't want everything to be an option :) )...  

In general it's a shame to make the scrolling so jerky, but I couldn't figure out anything better.

jgosmann:
I don't think the magic mouse counts as a normal mouse, but more like a track pad.

celiasmith:
Well, it's one of the most common mice used by mac users... (they tend to be apple snobs, you know :smiling_imp: )

PR #227, opened 2015-05-29
Timely thread shutdown

jgosmann:
This solves problems resulting in the frequent 'x zombie threads killed' message on shutdown (cp. PR #221). See commits (and commit messages) for details. Essentially, `close()` on a socket might take until the object is garbage collected. For an instant close `shutdown` has to be called beforehand, but that can lead to exceptions because threads are suddenly missing their sockets, so that has to be taken care of. And lastly, their where iterations over a list while it was modified. That's never a good idea.

Time to sleep now ... :sleeping:

### DIFF
--- a/nengo_viz/swi.py
+++ b/nengo_viz/swi.py
@@ -406,8 +406,10 @@ class SimpleWebInterface(BaseHTTPServer.BaseHTTPRequestHandler):
             print("Shutting down server...")
 
             # shut down any remaining threads
+            # server.requests might be modified from other threads, so we need
+            # a copy which we get in a thread-safe way be slicing it with [:].
             if asynch and server.requests is not None:
-                for _, sock in server.requests:
+                for _, sock in server.requests[:]:
                     try:
                         sock.shutdown(socket.SHUT_RDWR)
                         sock.close()
@@ -415,7 +417,7 @@ class SimpleWebInterface(BaseHTTPServer.BaseHTTPRequestHandler):
                         pass
 
                 first = True
-                for thread, _ in server.requests:
+                for thread, _ in server.requests[:]:
                     if thread.is_alive():
                         # giving the first thread more time to close
                         # effectively gives all threads more time to close
@@ -423,7 +425,7 @@ class SimpleWebInterface(BaseHTTPServer.BaseHTTPRequestHandler):
                         first = False
 
                 n_zombie = sum(thread.is_alive()
-                               for thread, _ in server.requests)
+                        for thread, _ in server.requests[:])
                 if n_zombie > 0:
                     print("%d zombie threads will close abruptly" % n_zombie)
 
@@ -447,22 +449,14 @@ class AsyncHTTPServer(SocketServer.ThreadingMixIn, BaseHTTPServer.HTTPServer):
         BaseHTTPServer.HTTPServer.__init__(self, *args, **kwargs)
 
         # keep track of open threads, so we can close them when we exit
-        self._requests = []
-        self._requests_lock = threading.Lock()
-
-    @property
-    def requests(self):
-        with self._requests_lock:
-            return list(self._requests)
+        self.requests = []
 
     def process_request_thread(self, request, client_address):
         thread = threading.current_thread()
-        with self._requests_lock:
-            self._requests.append((thread, request))
+        self.requests.append((thread, request))
         SocketServer.ThreadingMixIn.process_request_thread(
             self, request, client_address)
-        with self._requests_lock:
-            self._requests.remove((thread, request))
+        self.requests.remove((thread, request))
 
     def handle_error(self, request, client_address):
         exc_type, exc_value, _ = sys.exc_info()
--- a/nengo_viz/swi.py
+++ b/nengo_viz/swi.py
@@ -419,7 +419,7 @@ class SimpleWebInterface(BaseHTTPServer.BaseHTTPRequestHandler):
                     if thread.is_alive():
                         # giving the first thread more time to close
                         # effectively gives all threads more time to close
-                        thread.join(0.05 if first else 0.01)
+                        thread.join(0.2 if first else 0.01)
                         first = False
 
                 n_zombie = sum(thread.is_alive()
--- a/nengo_viz/swi.py
+++ b/nengo_viz/swi.py
@@ -447,14 +447,22 @@ class AsyncHTTPServer(SocketServer.ThreadingMixIn, BaseHTTPServer.HTTPServer):
         BaseHTTPServer.HTTPServer.__init__(self, *args, **kwargs)
 
         # keep track of open threads, so we can close them when we exit
-        self.requests = []
+        self._requests = []
+        self._requests_lock = threading.Lock()
+
+    @property
+    def requests(self):
+        with self._requests_lock:
+            return list(self._requests)
 
     def process_request_thread(self, request, client_address):
         thread = threading.current_thread()
-        self.requests.append((thread, request))
+        with self._requests_lock:
+            self._requests.append((thread, request))
         SocketServer.ThreadingMixIn.process_request_thread(
             self, request, client_address)
-        self.requests.remove((thread, request))
+        with self._requests_lock:
+            self._requests.remove((thread, request))
 
     def handle_error(self, request, client_address):
         exc_type, exc_value, _ = sys.exc_info()
--- a/nengo_viz/swi.py
+++ b/nengo_viz/swi.py
@@ -49,6 +49,7 @@ import random
 import select
 import string
 import sys
+import errno
 import os
 try:
     import StringIO
@@ -223,6 +224,11 @@ class SimpleWebInterface(BaseHTTPServer.BaseHTTPRequestHandler):
             self.user = self.testing_user
         try:
             getattr(self, command)(*[client] + args[1:], **db)
+        except socket.error as err:
+            # Only print traceback if it isn't likely to be caused by the
+            # server shutting down.
+            if err.args[0] != errno.EPIPE and err.args[0] != errno.EBADF:
+                traceback.print_exc()
         except Exception:
             traceback.print_exc()
         client.socket.close()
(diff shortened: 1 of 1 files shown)


nengo_viz/swi.py
```diff
@@ -402,14 +408,18 @@ def start(cls, port=80, asynch=True, addr='', browser=False):
             # shut down any remaining threads
             if asynch and server.requests is not None:
                 for _, sock in server.requests:
-                    sock.close()
+                    try:
+                        sock.shutdown(socket.SHUT_RDWR)
```
hunse: Hah, I tried both `shutdown` and `close`, but I totally missed that line in the docs about calling `shutdown` *before* `close`.

nengo_viz/swi.py
```diff
@@ -49,6 +49,7 @@ def ws_data(self, client):
 import select
 import string
 import sys
+import errno
```
hunse: We should probably use this down below, too, in ClientSocket.

nengo_viz/swi.py
```diff
@@ -402,14 +408,18 @@ def start(cls, port=80, asynch=True, addr='', browser=False):
             # shut down any remaining threads
             if asynch and server.requests is not None:
                 for _, sock in server.requests:
```
hunse: It might be fine just to copy the list rather than doing all the lock stuff, but the lock stuff is probably the more proper way to do it.

hunse: Actually, looking at your implementation down below, if list copying is atomic (and I think it is), then all the lock stuff is unnecessary.

hunse:
This looks great, especially if it fixes that zombie thread problem! I'm not convinced that the lock stuff in the last commit is necessary, though the list certainly should be copied before iterating over it.

nengo_viz/swi.py
jgosmann: Yeah, I was wondering about list copying being atomic, but I couldn't find any usable information on it before I decided that writing the implementation with the lock will quicker than figuring out the answer.

nengo_viz/swi.py
tbekolay: Also, would be good to alphabetize these imports ;)

tbekolay:
This looks great to me too! No opinion on the necessity of the lock, but seems like it can't hurt!

tcstewar:
> No opinion on the necessity of the lock, but seems like it can't hurt!

My only worry is that it might slow down the initial opening of the page (since that's when all the threads are getting started up).  Maybe just put in a "TODO: do some profiling to see if this lock slows things down", just so we don't forget about it in the future?

nengo_viz/swi.py
jgosmann: They weren't alphabetized in the first place, so I didn't bother ... though I agree.

jgosmann:
I think, we can increase start up time much more by optimizing CSS and Javascript and the first step there would be to have just one CSS and one JS file to load which automatically reduces the number of requests.

tcstewar:
> I think, we can increase start up time much more by optimizing CSS and Javascript

I totally agree.  :)

tbekolay:
Are there objections to using a templating engine like jinja2? We could use
that to include all the js and css into one file. We could, if we wanted,
generate the big js and css files just once in setup.py, then it would only
be an optional dependency... Maybe that's more of a long term thought
though.

tcstewar:
> Are there objections to using a templating engine like jinja2? We could use
> that to include all the js and css into one file.

If we're wanting to generate one efficient js and css file, then we really should be doing something like a javascript minimization thing, rather than using a templating engine.

tcstewar:
Are people happy with this PR?  It looks like the conversation has reached consensus, but there are comments about fixes that should be made that haven't been done yet...

tbekolay:
PR looks good to me!

Minimization doesn't have that big an effect, especially for local stuff
like this. I'll make a separate issue / PR for the templating stuff anyhow!

jgosmann:
I take care of the fixes now.

tcstewar:
> PR looks good to me!

So does that mean we are ignoring these comments:

> We should probably use this down below, too, in ClientSocket.

> Also, would be good to alphabetize these imports ;)

> Maybe just put in a "TODO: do some profiling to see if this lock slows things down", just so we don't forget about it in the future?

I'm mostly happy to ignore these, or to do them myself, but I usually wait for the person whose PR it is to respond to them....   :)

jgosmann:
just fixed the second one, will do the first one

jgosmann:
and will take a quick look at the third one

jgosmann:
give me like 5 minutes :clock1:

hunse:
I'd be happier if we got rid of the locks, just because I'm quite sure they're not necessary and they make the code messier. I'm worried people will see them and think they have to use locks with lists everywhere, and start adding needless complication elsewhere.

That said, what we've got here works fine, so I'm not going to complain if it gets merged.

jgosmann:
I'm happy to remove the locks, if anyone has a trustworthy source that copying the list is safe.

I changed the error codes in ClientSocket to the appropriate constants, but couldn't find one of the constants in my `errno`.

jgosmann:
rebased to master

tcstewar:
> I'm happy to remove the locks, if anyone has a trustworthy source that copying the list is safe.

This http://effbot.org/pyfaq/what-kinds-of-global-value-mutation-are-thread-safe.htm says that ```extend``` is thread-safe, so I suppose we could do ```[].extend(server.requests)``` if we really wanted to...  But I can't yet find an explicit statement that ```list(server.requests)``` is thread-safe....  still looking...

jgosmann:
The variance in load time is way larger than whatever the influence of the locks is and I honestly don't think that they have a large affect. The lock in only held for the append/remove operation, so other threads don't have to wait long to acquire it.

jgosmann:
Also not sure if it's a good idea to rely on how the operations are represented in byte-code. This seems like something that could change in future versions.

tcstewar:
> I changed the error codes in ClientSocket to the appropriate constants, but couldn't find one of the constants in my errno.

Yikes, looks like this errno thing might require some rethinking....  the values you changed it to don't map on to the errnos for me.  For me, EDEADLK=36 and EAGAIN=11 and WSAEWOULDBLOCK=10035.

tcstewar:
> The variance in load time is way larger than whatever the influence of the locks is and I honestly don't think that they have a large affect.

I totally agree, but this is more about @hunse 's comment that "I'd be happier if we got rid of the locks, just because I'm quite sure they're not necessary and they make the code messier. I'm worried people will see them and think they have to use locks with lists everywhere, and start adding needless complication elsewhere."  Python never requires locks on append or remove -- it guarantees those actions are thread-safe and will always be.  That's the whole point of the GIL.

hunse:
That source says that anything that looks atomic, is, and `list(x)` certainly looks atomic.

> Also not sure if it's a good idea to rely on how the operations are represented in byte-code. This seems like something that could change in future versions.

But when the way they're represented in byte-code affords certain features, i.e. thread safety, I don't think that's something they're going to change. That would break a lot of people's code.

jgosmann:
On the lock: I now compared how long it takes to acquire the lock and add to the list vs. only adding to the list. I eyeballed it, but a difference of 10μs seems to be a conservative estimate. Thus, we would need to process 100 request to even be 1ms slower. No user will notice ...

But sure, the performance side was only part of the argument.

tcstewar:
> Yikes, looks like this errno thing might require some rethinking.... the values you changed it to don't map on to the errnos for me. For me, EDEADLK=36 and EAGAIN=11 and WSAEWOULDBLOCK=10035.

Note: the way I came up with those numbers in the original code was just running it on different operating systems and seeing what error number came back when there was no data to read.  It was 10035 on windows, and I don't remember which operating systems had 35 and 11.  

So I think we should leave the code with the original numbers and make an Issue to go find out what the errno mapping is on these different operating systems.

Sorry!

hunse:
Also, the source says that `L[i:j] = L2` is thread safe, which suggests `L2 = L[i:j]` is thread safe, which suggests `L2 = L[:]` is thread safe, so we could always just use `L[:]` to make our copy.

hunse:
> So I think we should leave the code with the original numbers and make an Issue to go find out what the errno mapping is on these different operating systems.

Agreed. I think the point of `errno` is that you can just use one statement to cover all OSes, but we'll see.

tcstewar:
> Agreed. I think the point of errno is that you can just use one statement to cover all OSes, but we'll see.

I hope so...  I just have this sneaking suspicion that the "WS" in "WSAEWOULDBLOCK" means "Windows Socket".....   <sigh>   But in any case, it'll be better than the bare numbers I left it with....

tcstewar:
> so we could always just use L[:] to make our copy.

Sweet!  I'm sold on that logic.  :)

jgosmann:
Removed the errno commit and implemented the `L[:]` syntax. But I'm not sold on it. Now everyone accessing `server.requests` has to know that it might be modified by other threads and has to remember to access it an a thread safe way. (Also it looks a little bit strange like it's not doing anything. So if someone misses the comment on top it might get removed again?)

jgosmann:
btw: `list(L)` produces a `CALL_FUNCTION` byte-code. So it's still not clear to me whether it would be thread-safe.

jgosmann:
How the heck did I type 'git' instead of 'rid'? :open_mouth: Should get squashed with 8abe2d8 anyways.

tcstewar:
Thank you!  This all looks great.  :)

hunse:
> Now everyone accessing `server.requests` has to know that it might be modified by other threads and has to remember to access it an a thread safe way.

I thought you might keep `server._requests` and then have the `server.requests` property return the copy. But on the other hand, now the copy is visible where it's used, and it might help people get used to seeing lists copied when they're iterated over for thread safety.

jgosmann:
I might change it to to the property.

Another thing that just occurred to me:
```python
# giving the first thread more time to close
# effectively gives all threads more time to close
thread.join(0.2 if first else 0.01)
```
This logic is not correct. The first thread might be finished very quickly in which case the overall timeout is a lot smaller than the 0.2s.

hunse:
> This logic is not correct. The first thread might be finished very quickly and in that case the overall timeout is a lot smaller than the 0.2s.

Yeah, it's just a heuristic. The idea is that 0.01 s should be enough time for any thread to join, but I gave the first one a bit more time just in case the machine was slow or something. But really, now that we have a much better thread closing procedure (i.e. zombies are much less likely), I'd be happy just picking a fixed time for all threads, maybe 0.05 s. I guess the risk is that if *all* threads have zombied, and there's 100 of them or something, then that could be 5 seconds.

We could do something like in your new PR, but I don't want to over-complicate things.

PR #229, opened 2015-05-29
Added hover border, swapped rounded corners

tcstewar:
This adds an outline around Ensembles and Nodes when mouseover,
and moves uses that outline for the resizing.  We also make
Networks have straight corners and Nodes have rounded corners.
This fixes the resizing problems (including the bottom corners)

### DIFF
--- a/nengo_viz/static/viz_netgraph.css
+++ b/nengo_viz/static/viz_netgraph.css
@@ -51,10 +51,10 @@ g.net rect{
 
 .netgraph rect.highlight {
     stroke-width: 1;
-    stroke: #888;
+    stroke: rgba(0,0,0,0.1);
 }
 
 .netgraph rect.dragging {
     stroke-width: 1;
-    stroke: #888;
+    stroke: rgba(0,0,0,0.1);
 }
--- a/nengo_viz/static/viz_netgraph.css
+++ b/nengo_viz/static/viz_netgraph.css
@@ -43,3 +43,18 @@ g.net rect{
     dominant-baseline: text-before-edge;
     }    
 
+.netgraph rect.border {
+    stroke-width: 0;
+    stroke: black;
+    fill: transparent;
+}
+
+.netgraph rect.highlight {
+    stroke-width: 1;
+    stroke: #888;
+}
+
+.netgraph rect.dragging {
+    stroke-width: 1;
+    stroke: #888;
+}
--- a/nengo_viz/static/viz_netgraph_item.js
+++ b/nengo_viz/static/viz_netgraph_item.js
@@ -136,7 +136,8 @@ VIZ.NetGraphItem = function(ng, info) {
         var tmp = this.border === undefined ? this.shape : this.border;
         interact(tmp)
             .resizable({
-                edges: { left: true, right: true, bottom: true, top: true }
+                edges: { left: true, right: true, bottom: true, top: true },
+                invert: 'reposition'
                 })
             .on('resizestart', function(event) {
                 self.menu.hide_any();
@@ -170,12 +171,6 @@ VIZ.NetGraphItem = function(ng, info) {
                     if (event.edges.bottom) {
                         h *= -1;
                     }
-                    if (w < 0) {
-                        w = 1;
-                    }
-                    if (h < 0) {
-                        h = 1;
-                    }
 
                     var screen_w = self.get_width();
                     var screen_h = self.get_height();
@@ -189,6 +184,11 @@ VIZ.NetGraphItem = function(ng, info) {
                         h = w / self.aspect;
                     } else {
                         w = h * self.aspect;
+                    }
+
+                    if (w < 0 && h < 0) {
+                        w = -w;
+                        h = -h;
                     }
 
                     var scaled_w = w / h_scale;
--- a/nengo_viz/static/viz_netgraph_item.js
+++ b/nengo_viz/static/viz_netgraph_item.js
@@ -182,7 +182,7 @@ VIZ.NetGraphItem = function(ng, info) {
 
                     if (horizontal_resize && vertical_resize) {
                         var p = (screen_w * w + screen_h * h) / Math.sqrt(
-                            screen_w * screen_w + screen_h * screen_h);
+                            2 * (screen_w * screen_w + screen_h * screen_h));
                         h = p / self.aspect;
                         w = p * self.aspect;
                     } else if (horizontal_resize) {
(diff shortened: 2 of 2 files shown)


tcstewar:
I definitely like how this fixes a lot of the resizing issues I was having.  How do people feel about the slight change to the look (rounded corners on Nodes instead of Networks)?

jgosmann:
Feels strange, but I suppose, I'll get used to it. (I don't think I had resizing issues beforehand?)

Issues:
* [ ] Ensembles are not centered in the frame (a lot of whitespace to the left, the right is extending outside of the frame, a little whitespace at the bottom, no whitespace at the top). The amount of whitespace padding should match that around the ensembles.
* [ ] It feels weird, that there's no hover effect for networks. (But adding a second border might be equally weird?)
* [ ] The corner grab handle for ensemble was on the enclosing circle before, wasn't it? Now it's on the corner of the enclosing square, screwing up the resize math (make the ensemble large, than try resizing it in one corner and observe how it suddenly enlarges). Will be an easy fix, but should definitely be fixed.

jgosmann:
More issues:
* [ ] The padding between the highlight frame and the node depends on resizing. I think, this is one of the few places where a fixed pixel padding is better. It looks especially weird if you resize to an aspect ratio far off 1.
* [ ] Because of the previous issue, the frame might either include or not include the label which is inconsistent. It's the worst if goes right through the label.

tcstewar:
Good points...  I'll see what I can do.  :)

One thing I'm also going to try as an option is just slightly changing the colour of the Item, rather than doing this border thing (but keeping the invisible border to fix the resizing issues).

jgosmann:
And just discovered another one:
* [ ] If dragging an item A across another item B and B is in front of A the frame will switch to B while the mouse is over it (but still dragging A).

tcstewar:
I'm liking the borders less and less after playing with this.

With that in mind, I've made PR #233, which just has the resizing issue resolved.  The resizing behaviour there is the same as here, but there's no visual changes.  I'll continue to fiddle with this PR to see if I can improve the look of the borders (and deal with the problems noted here by @jgosmann ).

tcstewar:
> The corner grab handle for ensemble was on the enclosing circle before, wasn't it? Now it's on the corner of the enclosing square, screwing up the resize math (make the ensemble large, than try resizing it in one corner and observe how it suddenly enlarges). Will be an easy fix, but should definitely be fixed.

Looks like this was just a typo in the original resize code, which was never noticed before because it was so hard to do a corner-resize on an Ensemble: https://github.com/nengo/nengo_viz/commit/c63e770cd7686a6ec3385b72d9c8eeb39049f371

tcstewar:
> Looks like this was just a typo in the original resize code, which was never noticed before because it was so hard to do a corner-resize on an Ensemble: c63e770

Never mind, @jgosmann set me straight on that and pointed out the correct fix.  Thank you!

tcstewar:
> Ensembles are not centered in the frame (a lot of whitespace to the left, the right is extending outside of the frame, a little whitespace at the bottom, no whitespace at the top). The amount of whitespace padding should match that around the ensembles.

I've shifted the Ensemble circles a bit to the left to compensate.

tcstewar:
> It feels weird, that there's no hover effect for networks. (But adding a second border might be equally weird?)

Yeah, I think it feels weird too....  I've now added them to the Networks, and I think I like it better with that consistency.

tcstewar:
> The padding between the highlight frame and the node depends on resizing. I think, this is one of the few places where a fixed pixel padding is better. It looks especially weird if you resize to an aspect ratio far off 1.
> Because of the previous issue, the frame might either include or not include the label which is inconsistent. It's the worst if goes right through the label.

I've switched things around so that the border is a fixed size *larger* than the shape (rather than shrinking the shape a bit).  These feels a lot better to me, and deals with those two issues.

tcstewar:
> If dragging an item A across another item B and B is in front of A the frame will switch to B while the mouse is over it (but still dragging A).

This looks like a duplicate of #141.  If we bring the item to the top when we drag it, then this frame problem should go away.

tcstewar:
With these fixes (thanks to @jgosmann !), I'm starting to like this border thing more.  The only thing I'm not sure about yet is whether the rounded corners should be on the Networks or on the Nodes.

tbekolay:
Hmm... I'm undecided about this change. It kind of feels like I'm using a debug version of the GUI, which the developers use to refine the resize behavior... I'm not sure I would really expect this in a finished version of a GUI. While I think it does help with certain resizing behavior, I'm not that convinced that users will actually resize Ensembles all that often, and so giving them this border hint feels like overkill since it feels like something that doesn't happen in very many GUIs.

This feeling might be alleviated a bit if networks don't get the border. That's probably the part that makes it feel the most weird to me. But, it does feel inconsistent, as @jgosmann noted earlier. So yeah, I'm not sure.

I think I like having nodes be rounded rectangles and networks sharp rectangles though. Feels more consistent to have round things in square holes rather than the opposite.

jgosmann:
* The border still touches the label which doesn't look good. Maybe just move the label a bit lower? If that doesn't make it look to far away without the border ...
* Ensemble are still not completely centered.

tcstewar:
New thing to try: I've changed the border to be much fainter.  Does this feel less debuggy?  Having this hint of a border seems to help me a bit resizing, but doesn't draw my attention anywhere near as much as the original version of this border thing.

jgosmann:
Not sure about the debugginess. I feel it's harder on the eyes because it is only slightly visible, but maybe it's just that it's late again.

celiasmith:
I actually wonder if this would look ok if we were doing the highlighting with graphs at the same time?  Although I'd probably think that highlighting the ensemble by outlining all the circles might look better than a square.  

But maybe if the highlighting was being used to indicate something useful to the user independent of grabbing, it would seem less like a debug feature and more like helping the user to organize the display (i.e. grouping all graphs/network items related to each other).

tcstewar:
> But maybe if the highlighting was being used to indicate something useful to the user independent of grabbing, it would seem less like a debug feature and more like helping the user to organize the display

Possibly...  One other thing that will be happening is that if they have the code editor open, it'll also be highlighting the line of code responsible for creating the item they're hovering over...  Not sure if that'll help or hinder this, though.

tbekolay:
I agree with @jgosmann, the faint borders are very tough to see; and with @celiasmith, if there's some other reason to do the highlighting then I'd be more okay with it.

This is making me wonder why have a hover border on graphs? I never really questioned the behavior there, but I don't like it for the netgraph items, so perhaps I would like the graphs to not have that behavior too. Especially if there's a transparent border handling the resize events on those too.

tcstewar:
> This is making me wonder why have a hover border on graphs?

Because without the hover border it's almost impossible to figure out where to place your mouse cursor to let you resize a graph.  There's so much white space around those graphs that moving/resizing gets very surprising without some indication of their edges.

tcstewar:
All the useful stuff from this PR has made it into master, and if we want a border on the NetGraphItem.area, then we can add that from scratch easier than using this PR.

PR #230, opened 2015-05-29
Closed the websocket of a component after it is removed. 

s72sue:
This caused a bug where multiple threads trying to update the client caused a crash for the semantic pointer plot.


jgosmann:
LGTM
I think, that function could use some refactoring. It has five levels of nesting and a lot of branching. But that does not need to be done for this PR.

tcstewar:
> I think, that function could use some refactoring. It has five levels of nesting and a lot of branching. But that does not need to be done for this PR.

Probably a good idea for a future PR.  :)

PR #231, opened 2015-05-30
Fixed filename display on Windows

tcstewar:
On Windows, the paths include backslashes, which
weren't being escaped in the string, so they were bring
interpretted as special escape characters

### DIFF
--- a/nengo_viz/components/sim_control.py
+++ b/nengo_viz/components/sim_control.py
@@ -102,7 +102,7 @@ class SimControl(Component):
         info = dict(uid=self.uid)
         fn = json.dumps(self.viz.viz.filename)
         js = self.javascript_config(info)
-        return ('sim = new VIZ.SimControl(control, %s);\n' \
+        return ('sim = new VIZ.SimControl(control, %s);\n'
                 'toolbar = new VIZ.Toolbar(%s); ' % (js, fn))
 
     def message(self, msg):
--- a/nengo_viz/components/sim_control.py
+++ b/nengo_viz/components/sim_control.py
@@ -100,11 +100,9 @@ class SimControl(Component):
 
     def javascript(self):
         info = dict(uid=self.uid)
-        json = self.javascript_config(info)
-        fn = repr(self.viz.viz.filename)
-        if fn.startswith('u'):
-            fn = fn[1:]
-        return 'sim = new VIZ.SimControl(control, %s); ' % json + \
+        fn = json.dumps(self.viz.viz.filename)
+        js = self.javascript_config(info)
+        return 'sim = new VIZ.SimControl(control, %s); ' % js + \
                'toolbar = new VIZ.Toolbar(%s); ' % fn
 
     def message(self, msg):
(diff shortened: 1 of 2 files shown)


jgosmann:
Might it be better to use JSON serialization for this?
Like proposed [here](https://stackoverflow.com/questions/17118714/escaping-javascript-strings-in-python), [here as the answer with more votes](https://stackoverflow.com/questions/3708152/how-to-escape-single-quotes-in-python-on-server-to-be-used-in-javascript-on-clie), [and here](https://stackoverflow.com/questions/1694939/generating-javascript-string-in-python).

While we're at it might there be other spots where escaping has to done (like the label strings)?

tcstewar:
> Might it be better to use JSON serialization for this?

Oooo....  that's way better.  I like that.  :)

> While we're at it might there be other spots where escaping has to done (like the label strings)?

Very good question.  I'll post that as a separate issue....

tcstewar:
> While we're at it might there be other spots where escaping has to done (like the label strings)?

Actually, it looks like those are all fine, since they are already being passed back and forth via JSON.

nengo_viz/components/sim_control.py
```diff
@@ -93,9 +93,10 @@ def get_status(self):
 
     def javascript(self):
         info = dict(uid=self.uid)
-        json = self.javascript_config(info)
-        return 'sim = new VIZ.SimControl(control, %s); ' % json + \
-               'toolbar = new VIZ.Toolbar("%s"); ' % self.viz.viz.filename 
+        fn = json.dumps(self.viz.viz.filename)
+        js = self.javascript_config(info)
+        return 'sim = new VIZ.SimControl(control, %s); ' % js + \
```
tbekolay: You don't need `+` and `\` here, and it probably makes sense to have a newline. So, this:
```python
return ('sim = new VIZ.SimControl(control, %s);\n'
        'toolbar = new VIZ.Toolbar(%s);' % (js, fn))
```
This works because the Python interpreter automatically concatenates strings next to each other. And since they're in a set of brackets, they don't need a continuation character.

tcstewar:
I just realized it was messed up in the webpage title as well, so I put in the same fix.

nengo_viz/components/sim_control.py
```diff
@@ -93,9 +93,10 @@ def get_status(self):
 
     def javascript(self):
         info = dict(uid=self.uid)
-        json = self.javascript_config(info)
-        return 'sim = new VIZ.SimControl(control, %s); ' % json + \
-               'toolbar = new VIZ.Toolbar("%s"); ' % self.viz.viz.filename 
+        fn = json.dumps(self.viz.viz.filename)
+        js = self.javascript_config(info)
+        return ('sim = new VIZ.SimControl(control, %s);\n' \
```
tbekolay: Again, the line continuation character `\` isn't necessary here :) There should be almost no places where it's necessary.

tcstewar: Argh, I can't type.  I meant to delete that.  <sigh>  And yay for getting rid of line continuations!  I really dislike how they look....

PR #232, opened 2015-05-30
Removed extraneous path from default labels

tcstewar:
If nengo_viz is inventing a label, it used to do something
like "model.net.subnet.input". All of the stuff before the
last "." is redundant, since that's already shown by
nesting inside the visual display.  So this change cuts it
down to just "input".

### DIFF
--- a/nengo_viz/viz.py
+++ b/nengo_viz/viz.py
@@ -363,6 +363,8 @@ class Viz(object):
         label = obj.label
         if label is None:
             label = default_labels.get(obj, None)
+            if '.' in label:
+                label = label.rsplit('.', 1)[1]
         if label is None:
             label = repr(obj)
         return label


jgosmann:
Do we care about users putting '.' in there labels?

tcstewar:
> Do we care about users putting '.' in there labels?

Users are totally free to do so, and if they manually put "."s in their labels they will still show up.  This fix only changes the label that nengo_viz makes up if you don't specify a label.

jgosmann:
Ah right, then this LGTM. :cake:

tbekolay:
There are no issues with uniqueness here? I.e., multiple objects can be named input? If so, LGTM!

tcstewar:
> There are no issues with uniqueness here? I.e., multiple objects can be named input? If so, LGTM!

Yup, labels are completely distinct from uids (the unique identifiers).  :)

tbekolay:
Cool cool, want me to merge?

tcstewar:
Oops, damnit, I merged to master instead of rebased.  One second and I'll fix it...

PR #233, opened 2015-05-30
Fixed resizing issues

tcstewar:
All NetGraphItems now have a transparent bounding box
around them, and that bounding box handles resize events.
For Nodes and Networks, the shape (a rect) also handles
resize events.  For Ensembles, the shape (6 circles) does not handle
resize, since it's already a small enough target.

### DIFF
--- a/nengo_viz/static/viz_netgraph_item.js
+++ b/nengo_viz/static/viz_netgraph_item.js
@@ -128,7 +128,7 @@ VIZ.NetGraphItem = function(ng, info) {
         interact(this.area)
             .resizable({
                 edges: { left: true, right: true, bottom: true, top: true },
-                invert: 'reposition'
+                invert: this.type == 'ens' ? 'reposition' : 'none'
                 })
             .on('resizestart', function(event) {
                 self.menu.hide_any();
--- a/nengo_viz/static/viz_netgraph_item.js
+++ b/nengo_viz/static/viz_netgraph_item.js
@@ -628,7 +628,8 @@ VIZ.NetGraphItem.prototype.redraw_size = function() {
         }
     }
 
-    var area_w = screen_w * 0.97; // the circle pattern isn't perfectly square
+    // the circle pattern isn't perfectly square, so make its area smaller
+    var area_w = this.type === 'ens' ? screen_w * 0.97 : screen_w;
     var area_h = screen_h;
     this.area.setAttribute('transform', 
             'translate(-' + (area_w / 2) + ', -' + (area_h / 2) + ')');
--- a/nengo_viz/static/viz_netgraph_item.js
+++ b/nengo_viz/static/viz_netgraph_item.js
@@ -127,7 +127,8 @@ VIZ.NetGraphItem = function(ng, info) {
         /** dragging the edge of item to change its size */
         interact(this.area)
             .resizable({
-                edges: { left: true, right: true, bottom: true, top: true }
+                edges: { left: true, right: true, bottom: true, top: true },
+                invert: 'reposition'
                 })
             .on('resizestart', function(event) {
                 self.menu.hide_any();
@@ -158,12 +159,6 @@ VIZ.NetGraphItem = function(ng, info) {
                     if (event.edges.bottom) {
                         h *= -1;
                     }
-                    if (w < 0) {
-                        w = 1;
-                    }
-                    if (h < 0) {
-                        h = 1;
-                    }
 
                     var screen_w = self.get_width();
                     var screen_h = self.get_height();
@@ -177,6 +172,11 @@ VIZ.NetGraphItem = function(ng, info) {
                         h = w / self.aspect;
                     } else {
                         w = h * self.aspect;
+                    }
+
+                    if (w < 0 && h < 0) {
+                        w = -w;
+                        h = -h;
                     }
 
                     var scaled_w = w / h_scale;
--- a/nengo_viz/static/viz_netgraph_item.js
+++ b/nengo_viz/static/viz_netgraph_item.js
@@ -176,7 +176,7 @@ VIZ.NetGraphItem = function(ng, info) {
 
                     if (horizontal_resize && vertical_resize) {
                         var p = (screen_w * w + screen_h * h) / Math.sqrt(
-                            screen_w * screen_w + screen_h * screen_h);
+                            2 * (screen_w * screen_w + screen_h * screen_h));
                         h = p / self.aspect;
                         w = p * self.aspect;
                     } else if (horizontal_resize) {
--- a/nengo_viz/static/viz_netgraph_item.js
+++ b/nengo_viz/static/viz_netgraph_item.js
@@ -57,7 +57,6 @@ VIZ.NetGraphItem = function(ng, info) {
 
     this.area = this.ng.createSVGElement('rect');
     this.area.style.fill = 'transparent';
-    g.appendChild(this.area);
 
     this.menu = new VIZ.Menu(this.ng.parent);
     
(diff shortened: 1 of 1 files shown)


tcstewar:
I find this resizing behaviour a huge improvement for me.  Two things in particular:

- dealing with somewhat-small Ensembles (in the old version they'd always resize on me when I just want to move them, even if they're not all that small)
- all 4 corners work for resizing (the bottom corners for Nodes and Networks didn't use to work well)

nengo_viz/static/viz_netgraph_item.js
```diff
@@ -120,11 +124,13 @@ VIZ.NetGraphItem = function(ng, info) {
...
-        var tmp = this.shape
-        if(info.type === 'ens') {
-            tmp = $(this.shape.getElementsByClassName('mainCircle'))[0];
+        var items = [this.area];
+        if (info.type !== 'ens') {
+            items.push(this.shape);
         }
-        interact(tmp)
+        for (var i = 0; i < items.length; i++) {
+          var tmp = items[i];
```
tbekolay: suddenly 2 line indents...?! I don't think the inconsistency is worth the smaller diff.

tbekolay:
I won't pretend to know what's going on with any of the resizing behaviors; if this makes things more consistent then it's cool with me.

I found it surprising, though, when I tried out this branch that I could never resize when the pointer is on one of the circles in the Ensemble icon. But, moving the pointer to the invisible space beside the circle makes it resizable. That feels super weird to me, especially since the actual resizable area is inside that circle (if that makes any sense; kind of hard to explain). Basically, I think the z-index of the transparent bounding box should be above the ensemble circles so that I can select anywhere on that bounding box, not just anywhere that there aren't circles.

tcstewar:
> I found it surprising, though, when I tried out this branch that I could never resize when the pointer is on one of the circles in the Ensemble icon.

I'll take a stab at moving that around.  :)

Also, if you could try out the ```borders``` branch again and see whether that feel is any better -- I find the borders made a huge difference in how the resizing of Ensembles felt for me.

tbekolay:
Trying out the borders now!

nengo_viz/static/viz_netgraph_item.js
tcstewar: :)  I agree -- that was just there because I was feeling lazy and was just trying out that for loop to see if it works.  Should definitely be fixed before merging, if we decide to go this route.

tcstewar:
> Basically, I think the z-index of the transparent bounding box should be above the ensemble circles so that I can select anywhere on that bounding box, not just anywhere that there aren't circles.

I switched that around and that seems much nicer (and gets rid of that annoying for loop that screwed up the indentation).

I did find that with the Ensembles being easier to resize like this, I was still resizing them too much, so I also reduced the resize margin a bit.

jgosmann:
* The ensemble corner resize can be a bit hard to find because it's quite a bit away from all circles.
* The resize border is not correctly aligned. Probably you should cherry-pick 6d06f4b429836bd4ef5ff3a870a52bba18d6d074 (or parts of it).
* An corner resize item on minimum size will always grap the bottom right corner. Even though it might look like the cursor is above the top right corner. I was thinking at first resizing wasn't working anymore because I tried to resize in the wrong direction.

celiasmith:
I think I'm liking this more than #229 ... but I *think* I like the square networks more (aesthetically not, but practically yes, because of thing fitting together better).
* I don't mind point 3 from @jgosmann, since the mouse cursor tells you what's happening, and it's so small people won't be too surprised it's weird.
* point 1 from @jgosmann is true, in some circumstances (large ensembles) but i'm not sure what you can do about it...
* I often have to hover around a corner for a while before it gives me the right interaction.  That's probably some subtle annoying thing.

All in all, a lot better than before!

tcstewar:
> The ensemble corner resize can be a bit hard to find because it's quite a bit away from all circles.

I agree.  It's a bit better now with the better centring of the icon, but I still sometimes miss.

> I don't mind point 3 from @jgosmann, since the mouse cursor tells you what's happening, and it's so small people won't be too surprised it's weird.

I actually really mind that one, so I just fixed it.  The mouse cursor tells you that you can resize in the northwest/southeast direction, but with the old behaviour it would only resize if you went down or right, so the mouse cursor was lying to me.  Now it'll resize in all directions.

celiasmith:
> I actually really mind that one, so I just fixed it.

Hmmm... well, the previous behaviour made more sense to me (i.e. you could only resize along the axis indicated by the cursor).  But I don't have a problem with this either.

tcstewar:
> Hmmm... well, the previous behaviour made more sense to me (i.e. you could only resize along the axis indicated by the cursor)

Hmm, maybe I wasn't being clear.  The previous behaviour was that you could only resize *in one direction* along the axis indicated by the cursor.  I.e. the cursor would indicate northwest<->southeast, but only southeast would work.  This fix was to allow either direction along that axis to work.

celiasmith:
Ah, that's odd... I wonder if there's a difference in behaviour we're seeing?  For me, the fix also has the resize stopping when you move northwest (in this example).  So the intent was for it to 'flip' during the nw movement? ... oops, just noticed my silly gui didn't grab the latest comment when i switched to the branch.  Yes, this is definitely better.

jgosmann:
Not sure what I think of the resize flipping over. Also, because there is a weird interaction with the minimum size. Once the node gets that small it will be moved around until the cursor is on the other side where the resizing flips over.

tcstewar:
> Not sure what I think of the resize flipping over. Also, because there is a weird interaction with the minimum size. Once the node gets that small it will be moved around until the cursor is on the other side where the resizing flips over.

I'm not sure either, but I don't see any other way to deal with the "only one resize option" problem.  :(

jgosmann:
Can we decide whether we allow to flip when the resize is started or is this a global setting? In the former case we could figure the correct setting out ...

tcstewar:
> Can we decide whether we allow to flip when the resize is started or is this a global setting?

It's mostly a global setting, but we can also do logic inside the resize handler.  No idea what that would be in this case, though.

tbekolay:
I prefer this to the other resizing PRs for sure!

I find that I really like the ability to do the flipped resizing (NW or SE when grabbing either corner) on ensembles, but not so much on nodes. I don't know why, but the fact that it's a simple rectangle makes it seem like it shouldn't flip, but the ensemble is a different beast so it feels natural there. I dunno!

One thing that surprised me in this was that I couldn't mouse over the left or right corner of a network to resize it. Apparently the transparent bounding box is actually like this:

bounding

Seems to me like it should be the same width as the network. This might be a moot point when the networks are sharp corners and the nodes round corners, though.

tcstewar:
> I prefer this to the other resizing PRs for sure!

Yay!  :)  I'm somewhat feeling like we should merge this one as it improves on the current behaviour, and worry about whether we want a visual indication of the mouseover separately.

> I find that I really like the ability to do the flipped resizing (NW or SE when grabbing either corner) on ensembles, but not so much on nodes.

I was feeling the exact same way.  Fortunately, that's an easy thing to change, so I've set it so that one Ensembles get this ability

> Apparently the transparent bounding box is actually like this:

Oops, that was a dumb bug in my code.  Fixed now.  :)

jgosmann:
> I find that I really like the ability to do the flipped resizing (NW or SE when grabbing either corner) on ensembles, but not so much on nodes.

This might be because ensembles have a fixed aspect ratio implying resizing around the center.

> > Can we decide whether we allow to flip when the resize is started or is this a global setting?

> It's mostly a global setting, but we can also do logic inside the resize handler. No idea what that would be in this case, though.

Should be something like this:
1. Determine the area where the resize borders overlap.
2. If the initial mouse down is in this area, allow resize in either direction (i.e., allow flipping over).
3. Disable resize in either direction after the resize borders stop overlapping or the mouse leaves the initially calculated area.

This way it should seem to the user that resizing is only possible in the direction intended. It's not even necessary to more click precisely to get the right corner.

But it's definitely a bunch of math and some logic.

jgosmann:
LGTM :cake: 
The remaining issues can be left for later ...

Issue #222, opened 2015-05-28
Netgraph takes super long in Firefox to load


jgosmann:
no idea why ...

tcstewar:
I see this every now and then, and then it magically goes away.  Do you have a situation where it's reliably reproducible?

jgosmann:
When I created the issue it was ... now it isn't. :/

tcstewar:
> When I created the issue it was ... now it isn't. :/

:(  I have no idea what causes this....

jgosmann:
[This might be the problem.](https://bugzilla.mozilla.org/show_bug.cgi?id=782704)

jgosmann:
[or this?](https://bugzilla.mozilla.org/show_bug.cgi?id=711793)

jgosmann:
This is still not completely solved: When we shutdown the server, all sockets are just closed without doing a proper closing handshake on the websockets.

hunse:
Do we need a handshake [here](https://github.com/nengo/nengo_viz/blob/c6b3f8997ed0d51e9de67c7e7fdac8f11ad1a7d9/nengo_viz/swi.py#L549) too?

jgosmann:
No, there the socket is already closed. That's the reason we get he exception. If the server dies because of an exception we could still try to close everything properly. but I'm not that concerned about it because something already went wrong, so the connections might as well terminate abnormally too.

PR #237, opened 2015-05-31
Perform a proper closing handshake for websockets.

jgosmann:
Addresses #222 and hopefully fixes it.

Without properly closing the connection it seems that Firefox treats it as an unexpected close due to an error which increases a delay before new websocket connections are allowed.


jgosmann:
[Documentation](https://tools.ietf.org/html/draft-ietf-hybi-thewebsocketprotocol-17#section-1.4)

tcstewar:
Looks great to me... I'll rebase and merge.  :)

PR #238, opened 2015-05-31
Shaded borders

tcstewar:
Here's a third version of the ongoing fun with resizing.

Here, instead of having borders, the item gets a faint shading applied to it.  This is a lot more subtle, but still draws attention and gives you a sense of where the object boundary is for the resizing.

### DIFF
--- a/nengo_viz/static/viz_netgraph.css
+++ b/nengo_viz/static/viz_netgraph.css
@@ -45,16 +45,13 @@ g.net rect{
 
 .netgraph rect.border {
     stroke-width: 0;
-    stroke: black;
     fill: transparent;
 }
 
 .netgraph rect.highlight {
-    stroke-width: 1;
-    stroke: #888;
+    fill: rgba(0,0,0,0.05);
 }
 
 .netgraph rect.dragging {
-    stroke-width: 1;
-    stroke: #888;
+    fill: rgba(0,0,0,0.05);
 }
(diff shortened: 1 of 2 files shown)


jgosmann:
If the frames didn't look like a debug version, this definitely does!

A bit different styling might help ... but not sure ...

tcstewar:
> If the frames didn't look like a debug version, this definitely does!

This is why I'm trying lots of different possibilities...   this looks less debuggy to me, but I'm weird. :)

celiasmith:
This works well on the networks (except it flashes when you drag over connections, as did the outlining... but that's ignorable for now).  But, i agree it looks weird on nodes and ensembles.  Making it a circle around ensembles and a rounded rectangle on nodes, unfortunately won't help the issue.  Sorry, I don't have anything more constructive :disappointed:

tbekolay:
Initially I actually thought that it was just the fill color of the node rectangle or ensemble circles changing, which I kind of liked (and would work with highlighting the right line in the source code editor) but yeah, having it be a full rectangle looks weird/debuggy to me, especially within expanded networks. I think if we have intuitive resizing behavior, we won't need to have this kind of visual help...

tcstewar:
People seem happy without any visual indication for now, and if we want to implement it it'll be easier to do it from scratch than to use this PR.

PR #240, opened 2015-05-31
Mousewheel sensitivity more appropriate for mac

celiasmith:
(and unchanged for other oss)  The results of discussion in #226

### DIFF
--- a/nengo_viz/static/viz_viewport.js
+++ b/nengo_viz/static/viz_viewport.js
@@ -2,9 +2,11 @@ VIZ.Viewport = function() {
     this.x = 0;
     this.y = 0;
     this.scale = 1.0;
-    
-    this.scale_step_size = 1.1;
-    
+
+    this.scale.step_size = 1.1;
+    if (navigator.appVersion.indexOf("Mac")!=-1) {
+        this.scale.step_size = 1.05;
+    }
     this.w = $("#main").width();
     this.h = $("#main").height();
     var self = this;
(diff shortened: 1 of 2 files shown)


jgosmann:
Nope. :no_good: 
I slightly move my fingers and the trackpad and the network graph disappears into nirvana.

jgosmann:
Same for the mouse wheel and both Chrome and Firefox.
This is like the worst version yet. :fire:

nengo_viz/static/viz_viewport.js
```diff
@@ -2,9 +2,11 @@ VIZ.Viewport = function() {
     this.x = 0;
     this.y = 0;
     this.scale = 1.0;
-    
-    this.scale_step_size = 1.1;
-    
+
+    this.scale.step_size = 1.1;
```
jgosmann: This should probably be `this.scale_step_size`.

nengo_viz/static/viz_viewport.js
```diff
@@ -2,9 +2,11 @@ VIZ.Viewport = function() {
...
+    if (navigator.appVersion.indexOf("Mac")!=-1) {
+        this.scale.step_size = 1.05;
```
jgosmann: Same here.

celiasmith:
You'll notice it was actually broken before i changed anything... I have no idea what's going on.  If I comment those three changes out, it's still broken...?  But, it seems fine in master!  I'll probably have to try again, as this is baffling.

nengo_viz/static/viz_viewport.js
celiasmith: yes

tcstewar:
This has been replaced by #247 which has now been merged.

PR #242, opened 2015-05-31
Minimap

studywolf:
All the code online seemed intentionally obfuscated, so I've been messing around not using any other libraries. So this is a homebrew minimap that just displays the whole model. It's pretty hacky at the moment but I thought it would be worthwhile to get some feedback on it just to see if people would be interested having something along these lines at all. 

It shows items, networks, and connections, but not any graphs at the moment (also was thinking it would probably be best to just have a placeholder for graphs if they're shown at all). It's just displayed always at the moment but it might make sense to have it only pop up if the whole model isn't on the main screen, and to possibly add a ctrl+M to get it to pop up or something. Would be nice to also be able to resize it and maybe drag it around to place it wherever you want it. Also this map isn't interactive at all, it's just for lookin'. 

Thoughts?

### DIFF
--- a/nengo_viz/static/viz_menu.js
+++ b/nengo_viz/static/viz_menu.js
@@ -42,7 +42,7 @@ VIZ.Menu.prototype.show = function (x, y, items) {
     this.menu_div.appendChild(this.menu);
     this.div.appendChild(this.menu_div);
 
-    VIZ.set_transform(this.menu_div, x - 20, y - 20);
+    VIZ.set_transform(this.menu_div, x - 20, y - $('.minimap').height());
 
     this.actions = {}
 
--- a/nengo_viz/static/viz_menu.js
+++ b/nengo_viz/static/viz_menu.js
@@ -42,7 +42,7 @@ VIZ.Menu.prototype.show = function (x, y, items) {
     this.menu_div.appendChild(this.menu);
     this.div.appendChild(this.menu_div);
 
-    VIZ.set_transform(this.menu_div, x - 20, y - $('.minimap').height());
+    VIZ.set_transform(this.menu_div, x - 20, y - 60 - $('.minimap').height());
 
     this.actions = {}
 
--- a/nengo_viz/static/main.css
+++ b/nengo_viz/static/main.css
@@ -37,7 +37,7 @@ div.minimap {
     left: 84.5%;
     top: 82.5%;
     position: relative;
-    z-index: 1000000000000;
+    z-index: 99999999;
 }
 
 .graph {
--- a/nengo_viz/static/main.css
+++ b/nengo_viz/static/main.css
@@ -21,7 +21,7 @@ div.minimap {
     border: 2px solid #787878;
     height: 15%;
     width: 15%;
-    left: 82.5%;
+    left: 84.5%;
     top: 82.5%;
     position: relative;
     z-index: 1000000000000;
--- a/nengo_viz/static/main.css
+++ b/nengo_viz/static/main.css
@@ -7,13 +7,26 @@ body {
     height: 100%;
     width: 100%;
     cursor: inherit;
+    display: -webkit-flex;
+    display: flex;
+    -webkit-flex-direction: column;
+    flex-direction: column;
 }
+
 
 #main {
     background: white;
     width: 100%;
-    height: calc(100% - 80px);
+    height: 100%;
+    position: absolute;
+    top: 0;
+    left: 0;
+}
+
+#vmiddle {
     position: relative;
+    flex: 1 1 auto;
+    -webkit-flex: 1 1 auto;
 }
 
 div.minimap {
--- a/nengo_viz/static/main.css
+++ b/nengo_viz/static/main.css
@@ -14,7 +14,21 @@ body {
     width: 100%;
     height: calc(100% - 80px);
     position: relative;
+}
 
+div.minimap {
+    background: white;
+    border: 2px solid #787878;
+    height: 20%;
+    width: 20%;
+    left: 75%;
+    top: 75%;
+    /* height: 15%; */
+    /* width: 15%; */
+    /* left: 82.5%; */
+    /* top: 82.5%; */
+    position: relative;
+    z-index: 1000000000000;
 }
 
 .graph {
--- a/nengo_viz/static/main.css
+++ b/nengo_viz/static/main.css
@@ -19,14 +19,10 @@ body {
 div.minimap {
     background: white;
     border: 2px solid #787878;
-    height: 20%;
-    width: 20%;
-    left: 75%;
-    top: 75%;
-    /* height: 15%; */
-    /* width: 15%; */
-    /* left: 82.5%; */
-    /* top: 82.5%; */
+    height: 15%;
+    width: 15%;
+    left: 82.5%;
+    top: 82.5%;
     position: relative;
     z-index: 1000000000000;
 }
--- a/nengo_viz/static/viz_hotkeys.js
+++ b/nengo_viz/static/viz_hotkeys.js
@@ -43,6 +43,11 @@ VIZ.Hotkeys = function () {
             self.callMenu();
             ev.preventDefault();
         }
+        // bring up minimap with ctrl-m
+        if (ctrl && key == 'm') {
+            VIZ.netgraph.toggleMiniMap();
+            ev.preventDefault();
+        }
     });
 }
 
--- a/nengo_viz/static/viz_hotkeys.js
+++ b/nengo_viz/static/viz_hotkeys.js
@@ -52,9 +52,9 @@ VIZ.Hotkeys = function () {
 }
 
 VIZ.Hotkeys.prototype.callMenu = function () {
-        VIZ.modal.title("Hotkeys list");
-        VIZ.modal.help_body();
-        VIZ.modal.show();
+    VIZ.modal.title("Hotkeys list");
+    VIZ.modal.help_body();
+    VIZ.modal.show();
 }
 
 VIZ.hotkeys = new VIZ.Hotkeys();
--- a/nengo_viz/static/viz_modal.css
+++ b/nengo_viz/static/viz_modal.css
@@ -127,3 +127,7 @@ div.modal {
 .input-group.col-xs-2 {
     left:20px;
 }
+
+.modal-dialog.modal-sm td {
+    padding: 4px;
+}
--- a/nengo_viz/static/viz_modal.js
+++ b/nengo_viz/static/viz_modal.js
@@ -79,10 +79,17 @@ VIZ.Modal.prototype.help_body = function() {
     this.clear_body();
 
     this.$div.find('.modal-dialog').addClass('modal-sm');
-    var $body = $('<table width=100%>');
-    $body.append('<tr> <td>Play / pause</td> <td align="right">Spacebar</td></tr>');
-    $body.append('<tr> <td>Undo</td> <td align="right">Ctrl-z</td></tr>');
-    $body.append('<tr> <td>Redo</td> <td align="right">Ctrl-Shift-z || Ctrl-y</td></tr>');
+    var $body = $('<table class="table-striped" width=100%>');
+    $body.append('<tr><td>Play / pause</td>' +
+                 '<td align="right">Spacebar</td></tr>');
+    $body.append('<tr><td>Undo</td>' +
+                 '<td align="right">Ctrl-z</td></tr>');
+    $body.append('<tr><td>Redo</td>'+
+                 '<td align="right">Ctrl-Shift-z, Ctrl-y</td></tr>');
+    $body.append('<tr><td>Toggle minimap</td>'+
+                 '<td align="right">Ctrl-m</td></tr>');
+    $body.append('<tr><td>Show hotkeys</td>'+
+                 '<td align="right">?</td></tr>');
     $body.append('</table>');
     $body.appendTo(this.$body);
 }
--- a/nengo_viz/static/viz_netgraph.css
+++ b/nengo_viz/static/viz_netgraph.css
@@ -42,10 +42,10 @@ g.net rect{
 }
 
 rect.view{
-    fill: none;
+    fill: black;
     stroke: black;
-    stroke-width: 5px;
-    stroke-dasharray: 10 3;
+    /* stroke-width: 5px; */
+    opacity:.2;
 }
 
 .recur {
--- a/nengo_viz/static/viz_netgraph.css
+++ b/nengo_viz/static/viz_netgraph.css
@@ -42,10 +42,8 @@ g.net rect{
 }
 
 rect.view{
-    fill: black;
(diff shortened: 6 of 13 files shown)


celiasmith:
Just looking at the functionality, this is super cool!  I'll try to open spaun in it sometime soon :smile: 

I could nitpick feedback, but if you're just looking for a general impression :+1:

jgosmann:
Cool :+1:

tbekolay:
I like this a lot! Seems like a pretty useful little thing to have, even without the graphs. I'm kind of undecided whether or not having the graphs in there would be helpful, but yeah, I'd have to see it.

I think that if you added a little rectangle showing where the current viewport is (which should hopefully be pretty easy to get at through `viz_viewport.js`) this would be a good addition as is. Being able to click on the minimap to move/rescale the viewport would be great too of course ;) And yeah, maybe having a top toolbar button to show / hide the minimap?

studywolf:
I'm having some trouble getting the scaling working properly for the viewbox, I'm not sure if it's because there are some weird scaling (*2?) values somewhere that I'm not aware of. @jgosmann could you take a look at the scaleMiniMapViewBox function in viz_netgraph.js and see if you can spot where things are going wrong?

jgosmann:
can do when I'm in the office

bjkomer:
I really like this! Well done :)

I think it makes more sense for graphs to not show up in the minimap, and possibly be shown tiled in a separate 'graph view' instead, but its hard to say without seeing it.

One minor thing I noticed, using auto-layout can zoom the minimap in or out to the wrong area. This might be related to the scaling stuff you mentioned.

jgosmann:
Fixed the viewbox. I'm not using `minimap_scale_x` and `minimap_scale_y` because they seem always to be 0.95... not sure what they are for. In fact, I'm not sure what the whole `scaleMiniMap` function is needed for as all calculated values seem to be constant and not to be used?

studywolf:
Hurray! ```The scaleMiniMap``` function is only needed when the components of the model are moved around, it scales the minimap so that the whole model remains on display.

jgosmann:
ah, yeah, moving stuff around can break the viewbox.

studywolf:
I think i've got it now...

jgosmann:
nope

studywolf:
dang. where is it breaking?

jgosmann:
Everywhere except for dragging items outside of the top.
I think, the width and height of the viewbox might need adjustment depending on the `minItem` and `maxItem` values.

jgosmann:
There we go

jgosmann:
btw: How about making the viewbox the same blue as in the timeline?

studywolf:
hey this I can do. had to approximate because of the opacity, but I think it's a pretty close match.

jgosmann:
Test whether it works now. I also moved the viewbox to the background so the color can be matched exactly.

studywolf:
huzzah! this looks like the viewbox works for real now! @bjkomer could you test this on yours too and make sure?

bjkomer:
Looks nice, the viewbox works really well for me!

The auto-layout bug still exists though. It seems to set the minimap's reference point to whatever area you were looking at when you clicked auto-layout. For example, if you pan so that the network is off screen, and hit auto-layout, the minimap disappears. As soon as you move any components it fixes itself, so there is probably just one update call missing somewhere.

bjkomer:
Just noticed that moving around components is really slow on firefox (works fine on chrome though)

bjkomer:
Looks like the slow components in firefox is not due to the minimap, it happens on master too. It seems to affect expanded networks the most.

studywolf:
oh awesome! ok maybe this can make it in before the school then, i'll work on the auto-layout thing

studywolf:
what about _now_, @bjkomer??

studywolf:
something i did notice that's a bug is that when loading from store layout configs the networks are all collapsed...which is weird...not sure where to look for that

studywolf:
alright that was some code i accidentally deleted on a rebase, that should be fixed now too!

studywolf:
changed from WIP to pull request!

celiasmith:
Minimap is cool...!  Couple things: In the 2D network, this cuts off the ensemble.  It also shows a resized node, depending on where the elements sit wrt one another.  

In net.py it seems that in general, the objects look different on the minimap than they do on the screen (they change shape on the map as i drag them around the screen).  Recurrent connections look especially odd (that's probably my fault :open_mouth:).

In spa.py, the networks get chopped in half when i drag the c network to the left (at one point, then it fixes again)....?

jgosmann:
Yes, objects in the minimap might have a different aspect ratio because the whole netgraph is taken and fitted into a box with the screen's aspect ratio. Those two might not be the same and thus the distortion.

studywolf:
for the recurrent connection, i'm not sure we can do anything to fix that, since it's made with an ellipse the line can't get any thinner...we could remake those connections using bezier curves instead...? we can fix the arrow size easy though.

celiasmith:
you can change the stroke-width style on the path element to make it however thin you want... also it's not an ellipse, it's an svg path.

studywolf:
wait no it's a path, what's going on. also, i'm getting that the ```transform, translate``` command for the ```recurrent_ellipse``` doesn't actually do anything for me... is that the same on anyone else's browser?

studywolf:
i'm seeing no difference when i change stroke-width though for recur, on minimap or in regular

celiasmith:
works for me on firefox and chrome... it's the class=recur path

studywolf:
i found the class, but it just doesn't make a difference when i change the stroke-width...i have it set to 20px and it looks exactly the same...

celiasmith:
don't know what to say, it works both on the main netgraph and minimap for me.

studywolf:
can you push your code?

celiasmith:
i'm just editing the webpage directly... but this works:
$('.minimap path.recur').attr('style','stroke-width:3px')

studywolf:
hmm, yeah inspecting the element there's a big strike through ```stroke-width: 2px```, so, that's something to investigate. @bjkomer are you getting this too?

bjkomer:
When I edit the stroke-width to 2px, it gets a lot smaller for me (editing with inspect element, haven't tried changing the code yet) The original stroke-width was set to 8px

studywolf:
Hmm, thanks! just looking at it more (thanks terry!) the stroke-width is being set for recurrent connections on line 236, so that's what's causing the problem for me. I'll mess around with that.

studywolf:
OK can you pull and see how it looks for you now, @celiasmith? 4px is what looks right on mine for minimap-recur.

bjkomer:
I'm seeing the recurrent line connection change thickness in the opposite way to the other lines when you zoom in/out. When you zoom out, the other lines get thicker, but the recurrent gets thinner and vice versa. (this is all not on the minimap, but the regular view) On the minimap the recur line gets thicker as you expand the network, while the other lines don't, which looks weird

bjkomer:
Another weird bug, if you resize the subnet to take up more space than the rest of the model, the minimap zooms in to a weird place, where you can't see the whole model.
Also, in my previous comment by 'expand' I mean resize the network to make it bigger. Just realized that could be ambiguous.

celiasmith:
Sorry, had git problems.  Yup this looks ok.

celiasmith:
Oh yeah, good in the minimap, but messed up now in the regular view the way brent describes.

studywolf:
alright, think i fixed that, at least.

studywolf:
OK, think that I got all of the things that you guys mentioned, @celiasmith and @bjkomer, could you check again?

bjkomer:
Looks great! All of those previous bugs are now fixed.

One very minor thing I just noticed, is that you need to press Ctrl-M twice to toggle the minimap the first time. Is that because the minimap is starting as shown by default, but its flag starts as not shown?

studywolf:
hmm this is very strange, when i console.log this.display it comes up as undefined the first time... i have no idea why.

studywolf:
thanks for checking brent! the ctrl-m is probably not a blocking thing, i would think...? might resolve itself when merged with the hotkeys branch anyway

celiasmith:
I'm still a bit worried by the fact that the minimap doesn't look like the actual network... that seems like a fairly significant concern.  I guess it's not blocking if everyone's fine with them not being the same?

studywolf:
hmmm. if we wanted to force the aspect ratio to be the same then we could just choose the larger of the two aspect ratios and have some whitespace in it.

i just pushed a change thath shows what that would look like, what do you think?

bjkomer:
I like this version a lot more!

celiasmith:
very nice!

celiasmith:
If you want to be *really* picky, the blue area is a bit big because the top bar cuts some of the screen off (that's included in the blue)...  :guitar:

studywolf:
i don't even know what you mean!

celiasmith:
Basically the blue thing is the size of your screen including under the top bar.  but what you can actually see doesn't include under the top bar...

nengo_viz/static/=
```diff
@@ -0,0 +1,601 @@
+/**
```
tbekolay: This file probably isn't supposed to exist? I think you vimmed too hard ;)

jgosmann:
The connections in the minimap do not disappear in Firefox when disabling the minimap.

jgosmann:
> Basically the blue thing is the size of your screen including under the top bar. but what you can actually see doesn't include under the top bar...

I've got a fix for that, but it will break nengo_viz on Safari ... (more than it is right now and with no easy fix)

jgosmann:
I believe, I can get it so that it does not break Safari :)

jgosmann:
> Basically the blue thing is the size of your screen including under the top bar. but what you can actually see doesn't include under the top bar...

Fixed it by changing the layout to a [flex layout](https://developer.mozilla.org/en-US/docs/Web/Guide/CSS/Flexible_boxes). Works in Chrome/Firefox/Safari. :)

celiasmith:
Seems good to me.  I just noticed the arrow on the recurrent connection is off the line and kinda huge-ish in the minimap.

celiasmith:
K, I fixed the recurrent arrows so they're all beautiful and stuff :fireworks: :nail_care:

studywolf:
yay this looks great! i also removed that file @tbekolay pointed out
are there any other things left?

tbekolay:
Looks good, I think! Needs a rebase though and should make a note to add Ctrl+M to that help modal.

jgosmann:
:warning: In Firefox connections are still shown when disabling the minimap.

studywolf:
OK i rebased to master and i'll look at the connections thing in firefox when i get in.

studywolf:
OK, for whatever reason needed to set the opacity of the ```g_conns_mini = 0``` for connections to disappear in firefox, so that works now! Anything else?

studywolf:
Alright I think this is good to go @tbekolay!

tbekolay:
Two things, should hopefully be quick fixes:

1. The minimap displays on top of the modal dialog backdrop for me. Can you drop the z-index to below the modal backdrop?
2. Should be the minimap be displayed by default? It's definitely cool, and useful for big models. But most people will start off with small models that don't require the minimap. I think it makes more sense to have it off by default, then people can enable it with Ctrl+M (or a top toolbar button?).

studywolf:
on it!

studywolf:
OK rebased again, set up with the hotkeys system, added a button to the left of the config button in the top toolbar. I think this is ready to go!

tbekolay:
Awesome! I like the icon you chose, makes sense to me.

I noticed one weird thing, that hopefully isn't tough to fix... if I click on the play button, it'll highlight the play button with blue. If I then press Ctrl+m to bring up / hide the minimap, then the simulation will start playing or pause, as though I've clicked the play button. This doesn't happen if you click on the icon in the toolbar; it also stops happening if you click on that icon; it only happens if the play button is selected (i.e., highlighted in blue).

Any ideas?

studywolf:
Hmm i'm not seeing these things happen on my linux box in chrome or
firefox...

studywolf:
does it happen when you hit any of the other hotkey combos?

tbekolay:
Nope, just Ctrl+m....

tbekolay:
I think you might have missed the `preventDefault` for Ctrl+m.

nengo_viz/static/viz_menu.js
```diff
@@ -42,7 +42,8 @@ VIZ.Menu.prototype.show = function (x, y, items) {
     this.menu_div.appendChild(this.menu);
     this.div.appendChild(this.menu_div);
 
-    VIZ.set_transform(this.menu_div, x - 20, y - 20);
+    VIZ.set_transform(this.menu_div, x - 20, y - 60 - $('.minimap').height());
+    console.log($('.minimap').height());
```
tbekolay: Probably remove this?

studywolf:
hmm no couldn't be the preventDefault...I would never miss that. Try again maybe...

tbekolay:
Hm, now it's working for some reason! Must have been my imagination... or ghosts or something...

studywolf:
definitely ghosts! definitely.

tbekolay:
How do you feel about the branch history...? Lots of commits in here... can I squash stuff or just leave it?

studywolf:
follow your heart! i trust you

tbekolay:
Oh! I... uh... I think those ghosts are spraying tears in my eyes or something... :sob:

studywolf:
Rainstorms happen on sunny days too <3

tbekolay:
Sweet :D Got it down to a mean, lean 26 commits. Then added one to add the minimap toggle to the hotkeys list. Merging, if no one has any issues with minimaps in master!

jgosmann:
For some reason changes I made so that the top toolbar is not considered as visible space disappeared

tbekolay:
The code is missing or just the commit? I might have squashed it in...

jgosmann:
The actual code, checked the diff and tried it out

jgosmann:
It's at least this commit 016c4aa947aac600ebf836dbe2d2748fb71ea411

jgosmann:
(git reflog ftw)

tbekolay:
Dag, I thought I rebased so nicely. Will do a full diff to make sure I didn't miss anything.

tbekolay:
Hmm.. my diff before and after rebase are the same, but whatevs, cherrypicking  016c4aa947aac600ebf836dbe2d2748fb71ea411! Notice anything else missing?

tbekolay:
I even put it in the right place in the history. Please give it a test @jgosmann!

jgosmann:
> my diff before and after rebase are the same

Did you diff the right things? Maybe someone force pushed?
But seems to be ok now.

tbekolay:
> Did you diff the right things?

Maybe! Before I did any squashing, my version of `homebrewmm` was at f417b91, so I dunno.

In any case, I'll merge this in.

Issue #244, opened 2015-05-31
Dragging items through one another is ugly


celiasmith:
This is especially the case when you're dragging something through a nested network.  Some populations go on top of the item being dragged, some below, etc.  

I guess the right thing to do is have the dragged item come to the top while being dragged.

What happens when you let go is not obvious.  Should it stay on top of where it was released? (Will this cause possible stacking problems with z_indexes?)  Probably don't want it to disappear behind where it is released... (or return to it's previous level, which could cause that).

tcstewar:
> I guess the right thing to do is have the dragged item come to the top while being dragged.

The problem is when you're dragging an expanded Network.  If you make the Network come to the top (and all of the things inside it), any of the Connections to something outside the Network will now be behind the network.  And if you make those Connections come to the front as well, then they'll be on top of the things they are connected to.  So I don't know what this should look like.  :(

> What happens when you let go is not obvious. Should it stay on top of where it was released? (Will this cause possible stacking problems with z_indexes?) Probably don't want it to disappear behind where it is released... (or return to it's previous level, which could cause that).

That's not a problem, since SVG doesn't do z-indexes.  It's just drawing things in order, and you move things to the top by putting them at the end of the draw order.  So just leaving things how they lokoed while you were dragging should be fine.

celiasmith:
Yeah this is a tough one.  Here's a possible, hopefully low-overhead option: Turn the links coming into the network to grey or dashed while it's being dragged.  This would pop them on top of their externally connected objects, but the visual change would make clear that this was temporary, and to be sorted out on click release.  On release, you could hide any connections where the middle of the external object was under the dragged item.

tcstewar:
This is definitely better than it was before (thanks to #245), but there's definitely still situations where it looks weird.  However, there isn't a clear thing to do to improve it.  If someone wants to take a stab at this, it'd be useful to identify a few specific situations where things look wrong and exactly what it ought to look like.  Then we might be able to sort out how to adjust the algorithm.

PR #245, opened 2015-05-31
Move items to front on drag

tcstewar:
This is meant to address #244 and #141 .  It moves NetGraphItems to the front whenever they are dragged.

However, it's not yet a perfect solution.  It currently does not move expanded Networks to the front, since that would put them on top of any child networks that are currently expanded.  I need to add some logic for that.

### DIFF
--- a/nengo_viz/static/viz_netgraph_item.js
+++ b/nengo_viz/static/viz_netgraph_item.js
@@ -98,6 +98,9 @@ VIZ.NetGraphItem = function(ng, info) {
         .draggable({
             onstart: function () {
                 self.menu.hide_any();
+                if (!self.expanded) {
+                    self.g.parentNode.appendChild(self.g); // move to front
+                }
             },
             onmove: function(event) {
                 var w = ng.get_scaled_width();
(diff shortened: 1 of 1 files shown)


tcstewar:
Argh, it's actually worse than I thought.  I actually don't think there is a solution to this.

The current commit handles the problem I noted above: now networks move to the front as well, pushing anything inside them even more in front.

The problem is what to do with connection lines.  If I'm dragging an expanded Network around and I want it to be in front of all the other Items, what happens to a Connection that is going from the Network I'm dragging to some other Item that's outside of the Network?  I need to draw that Connection above the Network but behind the Item, and I also need to draw the Network above the Item.  So there's no draw order that satisfies this.

Still, what's in this PR is an improvement.

jgosmann:
> The problem is what to do with connection lines. If I'm dragging an expanded Network around and I want it to be in front of all the other Items, what happens to a Connection that is going from the Network I'm dragging to some other Item that's outside of the Network? I need to draw that Connection above the Network but behind the Item, and I also need to draw the Network above the Item. So there's no draw order that satisfies this.

That's where we have to introduce 3D rendering. ;)

tcstewar:
> That's where we have to introduce 3D rendering. ;)

I think it'd have to be 4D to deal with this problem...  ;)

tbekolay:
I checked this out, looks good to me! The connections not being in front of things seemed a little weird, but only because I was explicitly looking for it; I doubt a normal user would ever notice.

As a sidenote (not for this PR) I'm not a fan of our current strategy for picking z-indices. We should figure out a better strategy for that at some point.

PR #246, opened 2015-05-31
Added 2-second delay before server shutdown

### LINKED ISSUE #241: Reset shuts down the server ...
... because the page is reloaded and thus all open connections are closed.

tcstewar:
This might fix #241 

@jgosmann could you check if this works for you?


jgosmann:
Seems to work.

tcstewar:
> Seems to work.

Cool.  I can't see any big problem with this approach, so I'm going to merge it in so it stops annoying you.  :)

PR #247, opened 2015-05-31
Use delta value in zooming.

jgosmann:
Unsatisfied with all attempts to fix the zooming, I did some research on my own. The first things on SO you find on this are actually outdated or the people don't know what they are talking about. The wheel event delta values are actually not as widely inconsistent as it may appear in the first place. Though there are some inconsistencies across browserse, platforms, and devices, they are in a sensible range and we don't have to achieve exactly the same result for any combination as long as it feels ok for any combination.

I tested this PR with
* Chrome, Firefox, and Safari on OS X
* Chrome, Firefox in a Windows VM

Trackpad zooming on OS X feels super smooth. :sparkles: Mouse wheel works well (a bit better in Firefox I think). On Windows it didn't feel as nice, but still ok/usable. The main issue there is probably that I don't have proper mouse drivers installed and one scroll wheel click is counted as scrolling by 3 lines (even scrolling a normal text page in the browser doesn't feel any better). Trackpad feels the same because it is treated as a normal mouse wheel in the VM.

So test this and see whether it works on your systems and devices! There is still range for sensitivity adjustments.

This would make PR #240 obsolete.

### DIFF
--- a/nengo_viz/static/viz_netgraph.js
+++ b/nengo_viz/static/viz_netgraph.js
@@ -116,7 +116,7 @@ VIZ.NetGraph = function(parent, args) {
             switch (event.deltaMode) {
                 case 1:  // DOM_DELTA_LINE
                     if (event.deltaY != 0) {
-                        var delta = Math.log(1. + Math.abs(event.deltaY)) * 30;
+                        var delta = Math.log(1. + Math.abs(event.deltaY)) * 60;
                         if (event.deltaY < 0) {
                             delta *= -1;
                         }
@@ -134,7 +134,7 @@ VIZ.NetGraph = function(parent, args) {
                     break;
             }
 
-            var scale = 1. + Math.abs(delta) / 400.;
+            var scale = 1. + Math.abs(delta) / 600.;
             if (delta < 0) {
                 scale = 1. / scale;
             }
--- a/nengo_viz/static/viz_netgraph.js
+++ b/nengo_viz/static/viz_netgraph.js
@@ -107,19 +107,37 @@ VIZ.NetGraph = function(parent, args) {
     interact(document.getElementById('main'))
         .on('wheel', function(event) {
             event.preventDefault();
-            if (self.in_zoom_delay) {
-                return;
-            }
-            self.in_zoom_delay = true;
-            setTimeout(function () { self.in_zoom_delay = false; }, 50);
 
             self.menu.hide_any();
 
             var x = (event.clientX / $(self.svg).width())
             var y = (event.clientY / $(self.svg).height());
 
-            var delta = event.wheelDeltaY || -event.deltaY
-            var scale = delta > 0 ? viewport.scale_step_size : 1.0 / viewport.scale_step_size; // will either be 1.1 or ~0.9
+            switch (event.deltaMode) {
+                case 1:  // DOM_DELTA_LINE
+                    if (event.deltaY != 0) {
+                        var delta = Math.log(1. + Math.abs(event.deltaY)) * 30;
+                        if (event.deltaY < 0) {
+                            delta *= -1;
+                        }
+                    } else {
+                        var delta = 0;
+                    }
+                    break;
+                case 2:  // DOM_DELTA_PAGE
+                    // No idea what device would generate scrolling by a page
+                    var delta = 0;
+                    break;
+                case 0:  // DOM_DELTA_PIXEL
+                default:  // Assume pixel if unknown
+                    var delta = event.deltaY;
+                    break;
+            }
+
+            var scale = 1. + Math.abs(delta) / 400.;
+            if (delta < 0) {
+                scale = 1. / scale;
+            }
 
             VIZ.Component.save_components();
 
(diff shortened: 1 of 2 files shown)


tcstewar:
Nice!  This works fairly well on my Windows machine...  the trackpad is perfect, but a single click of the mouse wheel jumps things by more than I'd like.  Still, way better than before!

jgosmann:
Which browsers on Windows? If you could provide some sample values of `deltaY` and `deltaMode` that could also help to fine tune it.

tcstewar:
On Chrome I get deltaMode = 0 and deltaY = 100 or -100.
On Firefox I get deltaMode = 1 and deltaY = 3 or -3.

After playing with it a little bit, I like these for constants (good behaviour on both browsers, with both scrollwheel and trackpad):

 ```var delta = Math.log(1. + Math.abs(event.deltaY)) * 60;```

``` var scale = 1. + Math.abs(delta) / 600.;```

**Comment by jgosmann** [comment]
Works for well on OS X too.

**Comment by tcstewar** [comment]
> Works for me well on OS X too.

Nice.  Anyone else want to chime in as to whether it works well for them?  :)

File: nengo_viz/static/viz_netgraph.js [hunk 1]
```diff
@@ -107,19 +107,37 @@ VIZ.NetGraph = function(parent, args) {
...
+                        }
+                    } else {
+                        var delta = 0;
+                    }
+                    break;
+                case 2:  // DOM_DELTA_PAGE
+                    // No idea what device would generate scrolling by a page
+                    var delta = 0;
+                    break;
+                case 0:  // DOM_DELTA_PIXEL
```
celiasmith: The way this is written, case 0 and default do the same thing.  Is the case 0 here for completeness and possible future changes?  Is this way of future proofing preferable to writing a comment?

nengo_viz/static/viz_netgraph.js
```diff
@@ -107,19 +107,37 @@ VIZ.NetGraph = function(parent, args) {
...
+                default:  // Assume pixel if unknown
+                    var delta = event.deltaY;
+                    break;
+            }
```
celiasmith: Another way to do the previous is 
```
            var delta = event.deltaY;
            switch (event.deltaMode) { //DOM_DELTA_PIXEL or other has no effect
                case 1:  // DOM_DELTA_LINE
                    if (delta != 0) {
                        delta = Math.log(1. + Math.abs(event.deltaY)) * 60
                            * Math.sign(delta);
                    }
                    break;
                case 2:  // DOM_DELTA_PAGE
                    // No idea what device would generate scrolling by a page
                    var delta = 0;
                    break;
            }
```
It's a bit shorter and hopefully pretty clear.  Question is: Is this the kind of thing I should just not mention because it's bad etiquette (and maybe a waste of people's time)?  I could definitely see such a thing as put down to 'coding style' which obviously is different for different people, and which no one has an interest in arguing about really.  I honestly have no idea.

celiasmith:
Yay, this is awesome :fireworks:!  Soooooo much better than before!  I was just going to merge, but then had some questions (above), so thought I'd take the opportunity to get educated :smiley:.

nengo_viz/static/viz_netgraph.js
tcstewar: I prefer the original way of writing this.  Partly, this is a style thing that we haven't run into because Python doesn't have switch statements.  But in your (shorter) way of writing things, it isn't clear to me that we expect event.deltaMode==0 as a pretty common case.  So your way is kinda like doing this:

```python
value = 1
if condition:
   value = 2
```

which I find confusing, and I think we've generally been avoiding as a style thing.  So I'd prefer to have the handling for case 0 be explicit inside the switch statement.

As for having both ```default``` and ```case 0```, I liked having both of them in the switch statement, with the comment indicating that we don't really expect ```default``` to happen, but if it does, we'll use the same logic as ```case 0```.

nengo_viz/static/viz_netgraph.js
jgosmann: That is intended, but I'm open to other solutions (like setting the delta to 0 as for `DOM_DELTA_PAGE`. The `default` case should never occur, but if it does we don't really know what the delta value represents.

nengo_viz/static/viz_netgraph.js
jgosmann: The reason to include all conditions in the statement is to make for other readers clear that they haven't been forgotton.

celiasmith: Thanks, that makes sense.  So explicitness/clarity is a pretty strong constraint (which makes a lot of sense for maintenance).  Would the same reasoning apply to the "if x<0..." vs multiplying by Math.sign?

tcstewar: > Would the same reasoning apply to the "if x<0..." vs multiplying by Math.sign?

Good question.  I'd say that's a common enough thing to do that it's still clear what's going on, so I wouldn't force a multiply by Math.sign.

jgosmann: Not sure ... `Math.sign` makes the line and formula a bit longer, but would be only one assignment.

PR #248, opened 2015-05-31
More plot colors

jgosmann:
This makes the plotted line in the XY plot blue and the spike raster plots colorful.

Not sure whether this makes the spike raster plots too colorful. Maybe gradual changing colors would look better. (But will be a bit more work to implement and match to the other colors.)

### DIFF
--- a/nengo_viz/static/viz_raster.js
+++ b/nengo_viz/static/viz_raster.js
@@ -32,9 +32,10 @@ VIZ.Raster = function(parent, sim, args) {
 
     this.path.enter().append('path')
              .attr('class', 'line');
-    this.colors = VIZ.make_colors(this.n_neurons);
 
-    this.spikes = this.axes2d.svg.append("g").attr('class', 'spikes');
+    this.spikes = this.axes2d.svg.append("g")
+        .attr('class', 'spikes')
+        .style('stroke', VIZ.make_colors(1));
 
     this.update();
     this.on_resize(this.get_screen_width(), this.get_screen_height());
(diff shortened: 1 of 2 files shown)


tcstewar:
Looks great to me!  I like the blue on the xy plot, and I like the rainbow of colours for the spike plot.  But, I'm not allowed to make colour choices without supervision, so I'll wait for someone else to +1 it...  :)

celiasmith:
I like the blue on the 2d graph.  Not a fan of the rainbow... (also wouldn't like a rainbow in an activity grid).  The best I can think of is to keep them all the same blue as well.  Mostly because I can't think of a good, other, more subtle way to distinguish neurons.

celiasmith:
>  Maybe gradual changing colors would look better.
I was trying to think of something like this, but we also want to be careful not to make people think similar colours mean similar properties of the neurons (the neurons aren't ordered in any way).

tcstewar:
> Mostly because I can't think of a good, other, more subtle way to distinguish neurons.

This would be much harder to do, but in some of the NEF videos I chose the colour based on the encoder of the neurons.  In 1D it was blue for +1 and red for -1, and then a brightness change based on the intercept.  For 2D it was chosen from the colour wheel based on the encoder, and for >2D it just used the first 2 dimensions.   Not sure whether this was a good idea, but just thought I'd mention it.

tbekolay:
I definitely like the blue in the XY plots! I like the rainbow raster plots, personally, but I can see the argument against them. I'd prefer spikes to be all blue rather than all black though; all data should be colourful in some way.

Is setting the color a costly operation? This branch seemed really slow and jerky when a spike raster was showing. But, I just checked out master, and it's slow and jerky there too, so probably just a spike raster thing.

tcstewar:
> But, I just checked out master, and it's slow and jerky there too, so probably just a spike raster thing.

Yeah, spike rasters suck for rendering.  It'd be good to try implementing them via the canvas approach that @hunse used for the sparkleplots, if we could figure out how to make those not blur.....

jgosmann:
Yeah, the spike raster was awful in Firefox for me too. Chrome didn't have any trouble.
Setting the color might make it a bit worse.

jgosmann:
changed the spike raster to purely blue spikes

tcstewar:
Look great!  merging...  :)

PR #250, opened 2015-05-31
Made autolayout use full available space again

tcstewar:
This used to happen, but was lost in the merge with the undo stack.
There's also a commit here to change the logic so we don't run the layout algorithm three times in immediate succession.


celiasmith:
Functionally tested, works well.  LGTM

PR #252, opened 2015-05-31
Add option to make expanded networks transparent

tbekolay:
This is a bit of an experiment to see what it looks like when expanded networks are transparent, with just the border shown. I've been thinking for some time that this is the right way to render networks, but it's hard to convince even myself without actually seeing it. Having seen it now, I do like it, but will leave it to the consensus.

Note that this is still considered a WIP because I just did the minimal implementation in order to get it to work; the code needs to be looked at by someone more knowledgable with this part of the code base. Also, if the consensus thinks this is good, then we can remove some additional existing code; for example, we could remove `compute_fill` because we don't have to worry about the infinite depth issue.
A network is either collapsed (grey), or expanded (transparent).

Some example screenshots from `examples/net.py`.

First, with everything collapsed (same as before):

collapsed

With subnet expanded:

subnet

With subsubnet expanded

subsubnet

### DIFF
--- a/nengo_gui/static/modal.js
+++ b/nengo_gui/static/modal.js
@@ -160,6 +160,15 @@ Nengo.Modal.prototype.main_config = function(options) {
             '<input type="checkbox" id="fixed-resize">' +
             'Fix aspect ratio of elements on canvas resize' +
           '</label>' +
+          '<div class="help-block with-errors"></div>' +
+        '</div>' +
+      '</div>' +
+      '<div class="form-group">' +
+        '<div class="checkbox">' +
+          '<label for="transparent-nets" class="control-label">' +
+            '<input type="checkbox" id="transparent-nets">' +
+            'Expanded networks are transparent' +
+          '</label>' +
         '<div class="help-block with-errors"></div>' +
       '</div>' +
       '<div class="select">' +
@@ -183,6 +192,12 @@ Nengo.Modal.prototype.main_config = function(options) {
     $('#fixed-resize').prop('checked', options["aspect_resize"]);
 
     $('#config-fontsize').val(options["font_size"]);
+
+    $('#transparent-nets').prop('checked', options["transparent_nets"]);
+    $('#transparent-nets').change(function () {
+        Nengo.netgraph.transparent_nets = $('#transparent-nets').prop('checked');
+    });
+
     $('#config-fontsize').bind('keyup input', function () {
         Nengo.netgraph.set_font_size(parseInt($('#config-fontsize').val()));
     });
--- a/nengo_gui/static/netgraph.js
+++ b/nengo_gui/static/netgraph.js
@@ -17,6 +17,24 @@ Nengo.NetGraph = function(parent, args) {
     this.font_size = 100;       // font size as a percent of base
     this.aspect_resize = false;  //preserve aspect ratios on window resize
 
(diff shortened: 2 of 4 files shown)


tcstewar:
Interesting...  I do like that it sort of hints that Networks are just there for organization.  And I like getting rid of the ```compute_fill```.  That said, it's a very different visual aesthetic....

tbekolay:
I think this will look nicer with sharp-cornered networks and round-cornered nodes also.

jgosmann:
Doesn't look worse than the filled networks to me. Maybe a bit better.
Are the expanded networks completely transparent or do they have a white filling? In the former case items not included in the network might appear as they were included. But in the latter case it might not be obvious that they are hidden below a network?

tcstewar:
> But in the latter case it might not be obvious that they are hidden below a network?

No Node or Ensemble (or collapsed Network) is ever "below" an expanded Network.  So this PR doesn't do anything to affect that visibility issue.

tbekolay:
It sets the svg attribute fill-opacity to 0, so there would be no z-index
issues.

celiasmith:
Must admit I like the filled networks more.  Mostly because it makes expanded networks seem very different from collapsed ones...

tbekolay:
Yeah... I guess I think that there shouldn't be a big difference between expanded networks and collapsed ones. It's just grouping things. So, if you care about what's inside, you expand it and it looks like everything else; if you don't care about what's inside, and you sort of treat it like a black box, then you collapse it. At least, when I've made network diagrams like this in the past for papers, this is basically how I've done it.

bjkomer:
Another option could be to have them semi-transparent. I tried setting the opacity to 0.1 and I really like how it turned out. This way you can still visually see the nestedness of the networks but it doesn't get too dark when you have many of them, and you still have no z-index overlap issues. You also don't have to worry about calculating different fills for different levels, that comes for free because of how opacity works.

Of course this still breaks when you have very deep nested networks. I'm not sure the best way to get around this, but one option could be to have varying transparency as you go down. Start off with opacity 0.2 for the first layer, and then 0.1 for the second, 0.05 for the third, etc. In the limit you'll have an opacity of 0.4 (or whatever value is picked) and I think it will still look okay.

tbekolay:
FYI this is what it looks like with opacity = 0.1. Not too different to my eyes; I still like it fully transparent, personally.

opacity

studywolf:
Hmm, I'd vote for leaving as is or opacity = .5ish, when the networks to be filled (not white) I find they stand out from the background better.

tbekolay:
I've now changed this PR to add an option, rather than just doing it. Seems to be all the rage nowadays.

jgosmann:
If I have a network `A` with a subnetwork `B` and `A` is expanded (while `B` is not), activating the option makes both networks transparent.

tbekolay:
Right you are! I was just blindly setting all networks to transparent, even if they were collapsed. That should be [fixed](https://github.com/nengo/nengo_gui/commit/fd704650e84a2cdde2a870f009b6fe6d2ac17240#diff-4da79125e75575ec27bcee13e060ba46R31) now; also rebased this since #324 changed the options to a hash. Ready for another review!

jgosmann:
LGTM :cake: 

Might be nice to have a hover effect for transparent networks ...

PR #253, opened 2015-05-31
Handle any sort of error on backend import during monkeypatch

tcstewar:
When it tries to import nengo modules for the monkey-patching,
it might get something other than an ImportError if that module
is installed wrong.  For example, nengo_ocl on my machine
currently raises an AttributeError.  This change causes it to
handle any error raised.

### DIFF
--- a/nengo_viz/monkey.py
+++ b/nengo_viz/monkey.py
@@ -57,10 +57,10 @@ def patch():
     for name in known_modules:
         try:
             mod = importlib.import_module(name)
-            simulators[mod] = mod.Simulator
-            mod.Simulator = make_dummy(mod.Simulator)
-        except ImportError:
-            pass
+        except:
+            continue
+        simulators[mod] = mod.Simulator
+        mod.Simulator = make_dummy(mod.Simulator)
     yield
     for mod, cls in simulators.items():
         mod.Simulator = cls


tcstewar:
This is a pretty clear fix, and is needed on any machine with a broken nengo_ocl install (like mine), so I'll probably merge this PR in master pretty shortly, if there's no objections.  :)

drasmuss:
Does the user get an error message somewhere else telling them what went wrong?  I don't know this code base at all so can't really comment.  I just know in 1.4 the GUI had a tendency to silently catch legitimate import errors in user scripts, which made it really hard to debug those errors.

tcstewar:
> Does the user get an error message somewhere else telling them what went wrong?

This isn't anything to do with user code.  This is nengo_viz looking around to see what backends you have installed.  Right now, it's doing this by trying to import ```nengo_ocl``` and ```nengo_distilled``` and whatever else is on the list, and if it gets an ImportError then it assumes it's not installed.

The problem is if someone has manually tried to install nengo_ocl and has screwed things up in a weird way, then I get an AttributeError rather than an ImportError.  This crashes nengo_viz since it wasn't expecting any problems other than an ImportError, making it unusable.

In the future when we have some sort of user option to change the backend, then yes, we'd definitely want to give some sort of message as to what went wrong.  But for now, the only thing this code is doing is making sure someone hasn't put a ```nengo_ocl.Simulator(model).run(1000)``` in the middle of their script.  (and it is printing a message to the console when it sees something like that).

PR #256, opened 2015-06-01
Timely thread shutdown 2

jgosmann:
Few changes and fixing the timeout logic.

### DIFF
--- a/nengo_viz/swi.py
+++ b/nengo_viz/swi.py
@@ -69,6 +69,7 @@ except ImportError:
     import io as StringIO
 import sys
 import threading
+import time
 import traceback
 import webbrowser
 
@@ -416,13 +417,9 @@ class SimpleWebInterface(BaseHTTPServer.BaseHTTPRequestHandler):
                     except socket.error:
                         pass
 
-                first = True
                 for thread, _ in server.requests[:]:
                     if thread.is_alive():
-                        # giving the first thread more time to close
-                        # effectively gives all threads more time to close
-                        thread.join(0.2 if first else 0.01)
-                        first = False
+                        thread.join(0.05)
 
                 n_zombie = sum(thread.is_alive()
                         for thread, _ in server.requests[:])
--- a/nengo_viz/swi.py
+++ b/nengo_viz/swi.py
@@ -410,19 +410,19 @@ class SimpleWebInterface(BaseHTTPServer.BaseHTTPRequestHandler):
             # server.requests might be modified from other threads, so we need
             # a copy which we get in a thread-safe way be slicing it with [:].
             if asynch and server.requests is not None:
-                for _, sock in server.requests[:]:
+                for _, sock in server.requests:
                     try:
                         sock.shutdown(socket.SHUT_RDWR)
                         sock.close()
                     except socket.error:
                         pass
 
-                for thread, _ in server.requests[:]:
+                for thread, _ in server.requests:
                     if thread.is_alive():
                         thread.join(0.05)
 
                 n_zombie = sum(thread.is_alive()
-                        for thread, _ in server.requests[:])
+                        for thread, _ in server.requests)
                 if n_zombie > 0:
                     print("%d zombie threads will close abruptly" % n_zombie)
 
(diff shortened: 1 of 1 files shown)


nengo_viz/swi.py
```diff
@@ -409,23 +410,20 @@ def start(cls, port=80, asynch=True, addr='', browser=False):
...
-                first = True
-                for thread, _ in server.requests[:]:
+                start = time.clock()
+                for thread, _ in server.requests:
                     if thread.is_alive():
-                        # giving the first thread more time to close
-                        # effectively gives all threads more time to close
-                        thread.join(0.2 if first else 0.01)
-                        first = False
+                        thread.join(max(0.01, 0.2 - (time.clock - start)))
```
jgosmann: In theory `time.clock` might overflow which could produce an extremely long timeout ... but not sure if that's a reason to complicate the code. You'd have to terminate the server in exactly the right moment and the threads would have to block for some reason ...

hunse: Don't you need to call `time.clock` here? Also, is `time.clock` preferred over `time.time`? I always use the latter, but I'm always on Linux.

hunse:
This looks good. As I said, I'd also be fine just giving the threads a fixed timeout of 0.05 seconds or something.

nengo_viz/swi.py
jgosmann: How does that even type check?! ... ah right, dynamic typing, but it should have thrown an exception!

`time.clock` is preferred because it has a higher resolution. `time.time` only guarantees seconds (though it might be more on some systems).

nengo_viz/swi.py
```diff
@@ -409,23 +410,20 @@ def start(cls, port=80, asynch=True, addr='', browser=False):
...
-                first = True
-                for thread, _ in server.requests[:]:
+                start = time.clock()
+                for thread, _ in server.requests:
                     if thread.is_alive():
-                        # giving the first thread more time to close
-                        # effectively gives all threads more time to close
-                        thread.join(0.2 if first else 0.01)
-                        first = False
+                        thread.join(max(0.01, 0.2 - (time.clock() - start)))
```
hunse: Looking at the `time.clock` vs `time.time` issue again, it seems that `time.clock` gets confused by multithreading on my machine:
```python
import time
import timeit
import numpy as np

n = 1000
a = np.random.normal(size=(n, n))
f = lambda: np.dot(a, a.T)

t0 = np.mean(timeit.repeat(f, repeat=3, number=1))

t1 = time.time()
f()
t1 = time.time() - t1

t2 = time.clock()
f()
t2 = time.clock() - t2

print "timeit: %0.3f" % t0
print "time.time: %0.3f" % t1
print "time.clock: %0.3f" % t2
```
produces 
```
timeit: 0.015
time.time: 0.013
time.clock: 0.050
```
since my Numpy BLAS uses multithreading. If I substitute `f = lambda: sum(xrange(10000000))`,
then the results I get are comparable. Also, time.clock is considerably less accurate than time.time on my machine (clock only gives two decimal places, whereas time.time can measure down to the microsecond it seems).

Reading the docs, it seems that `time.clock` does completely different things depending whether you're on Windows (wall time) versus Unix (process time). Furthermore, the resolution of `time.time` is highly variable, and apparently some systems can't guarantee better resolution than 1 second.

Given all these complications, I'd vote just to have a constant timeout for threads of 0.05 seconds. I had only put in that extra time hack to help solve the zombie thread problem, which it didn't really help with, so I don't think this variable time thing is really necessary either. If we start running into problems where a thread is getting killed and it really needed more time, it will show up as a zombie thread, and we can look into it further.

jgosmann: Arg, I should get into a habit of reading the most recent docs and not always the 2.7. `time.perf_counter` would probably be the right function to use, but does not exist before Python 3.3 which would require even more code.
I agree, let's just do a constant amount.

jgosmann:
Simplified the timeout logic and rebased to master.

tcstewar:
Looks good to me.

hunse:
Cool!

PR #264, opened 2015-06-01
Fix Firefox cursor

jgosmann:
3rd attempt to fix the Firefox cursor.
Hopefully, this time it will work and stay in working condition.

### DIFF
--- a/nengo_viz/static/main.css
+++ b/nengo_viz/static/main.css
@@ -1,11 +1,12 @@
-html * {
-    cursor: inherit !important;
+html {
+    height: 100%;
 }
 
 body {
     margin: 0px;
     height: 100%;
     width: 100%;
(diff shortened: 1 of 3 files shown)


tbekolay:
Just tried out this branch on firefox. I haven't used FF much, so I'm not sure what the cursor issues are on it, but everything seemed to be the same as on Chrome except for the text in SimControl (the Time and Speed in the very bottom left). This gives the default pointer on Chrome, but the I-beam text cursor on FF. Presumably this is still an improvement though :)

jgosmann:
The text cursor in SimControl was not related to the problems fixed in this PR, but I see @celiasmith added a commit to take care of it.

tbekolay:
That commit fixed that issue ;) So this seems good to me!

PR #265, opened 2015-06-01
Restyle toolbar shadows.

jgosmann:
Someone commented that the new toolbar style was hard on the eyes because the light grey bleeds into the backgound. I think, this was mainly due to the light shadow which I made stronger now and which gives a better separation.


studywolf:
It was me! Thanks! This looks a liiiiitttlee dark to me, how would you feel about something in the middle, like #989898?

jgosmann:
Made it #999 which saves us three bytes. (Memory is expensive :moneybag: :stuck_out_tongue_winking_eye: )

studywolf:
LGTM!

PR #274, opened 2015-06-02
Invert scroll/zoom direction.

### LINKED ISSUE #258: Scrolling is backwards?
I just checked out master and was surprised that the scrollwheel behavior is the opposite of what I would have expected. Scrolling up zooms out, while scrolling down zooms in. I know that this behavior makes more sense with trackpads in Mac OS X, but it's the opposite of what linux / windows people expect. Could this be specific to the platform?

jgosmann:
Should make it the same as on Google Maps.

Fixes #258.

### DIFF
--- a/nengo_viz/static/viz_netgraph.js
+++ b/nengo_viz/static/viz_netgraph.js
@@ -135,7 +135,7 @@ VIZ.NetGraph = function(parent, args) {
             }
 
             var scale = 1. + Math.abs(delta) / 600.;
-            if (delta < 0) {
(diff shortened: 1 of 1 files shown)


jgosmann:
Please test.

tcstewar:
This is now consistent with Google Maps for me on Windows (Chrome and Firefox): scrollwheel up and trackpad down both zoom in.

celiasmith:
Works on my mac like google maps too.  Good to merge i think.

PR #277, opened 2015-06-02
Increase time slider min width.

### LINKED ISSUE #209: Adjusting the shown_time to min amount at the right hand side causes the time_slider to break
image

From this point the slider cannot be moved or resized anymore

jgosmann:
Fixes #209.


tbekolay:
This seems like a good minimum size; I can move and resize the slider as expected. I can't see the use case for having a smaller time slider ;)

Issue #269, opened 2015-06-02
Add keyboard shortcuts


tbekolay:
- [x] Undo (C-z) / redo (C-Z or C-y)
- [ ] Pop up / close code editor
- [x] Play / pause (spacebar)
- [x] List keyboard shortcuts (?)

Any more ideas?

tbekolay:
This is done except for the code editor; adding a note to that PR and closing!

PR #281, opened 2015-06-02
Added hotkey functionality and hotkey modal

studywolf:
Not much to this, added the functionality discussed in #269. The help modal could be made to look prettier maybe? Aside from that the only weird thing is that the hotkey to pull up the info box is ```ctrl-/``` because ```ctrl-h``` is reserved for history. I'm not sure that's a great hotkey for it, thoughts?

### DIFF
--- a/nengo_viz/disposable_js.py
+++ b/nengo_viz/disposable_js.py
@@ -35,12 +35,6 @@ def configmodal():
     js.append(show_modal_js())
     return '\n'.join(js)
 
-def helpmodal():
-    js = [add_modal_title_js('Hotkeys list')]
-    js.append('VIZ.modal.help_body();')
-    js.append(show_modal_js())
-    return '\n'.join(js)
-
 class PlotInfo(object):
     def __init__(self, title, plot="none"):
         self.title = title
--- a/nengo_viz/disposable_js.py
+++ b/nengo_viz/disposable_js.py
@@ -35,6 +35,12 @@ def configmodal():
     js.append(show_modal_js())
     return '\n'.join(js)
 
+def helpmodal():
+    js = [add_modal_title_js('Hotkeys list')]
+    js.append('VIZ.modal.help_body();')
+    js.append(show_modal_js())
+    return '\n'.join(js)
+
 class PlotInfo(object):
     def __init__(self, title, plot="none"):
         self.title = title
--- a/nengo_viz/static/viz_hotkeys.js
+++ b/nengo_viz/static/viz_hotkeys.js
@@ -1,0 +1,27 @@
+VIZ.Hotkeys = function () { 
+
+    document.addEventListener('keydown', function(ev) {
+        // bring up help sheet with ctrl-/ (which is the ? button)
+        if (ev.ctrlKey == true && ev.keyCode == 191) {
+            sim.ws.send('help');
+        }
+        // undoo with ctrl-z
+        if (ev.ctrlKey == true && ev.keyCode == 90) {
+            VIZ.netgraph.notify({ undo: "1" });
+        }
+        // redo with shift-ctrl-z
+        if (ev.ctrlKey == true && ev.shiftKey == true && ev.keyCode == 90) {
+            VIZ.netgraph.notify({ undo: "0" });
+        }
+        // redo with ctrl-y
+        if (ev.ctrlKey == true && ev.keyCode == 89) {
+            VIZ.netgraph.notify({ undo: "0" });
+        }
+        // run model with spacebar
+        if (ev.keyCode == 32) {
+            sim.on_pause_click();
+        }
+    });
+}
+
+VIZ.Hotkeys = new VIZ.Hotkeys();
--- a/nengo_viz/static/viz_hotkeys.js
+++ b/nengo_viz/static/viz_hotkeys.js
@@ -1,28 +1,48 @@
 VIZ.Hotkeys = function () { 
     var self = this;
 
-    document.addEventListener('keypress', function(ev) {
+    document.addEventListener('keydown', function(ev) {
+        if (typeof ev.key != 'undefined') {
+            var key = ev.key;
+        } else {
+            switch (ev.keyCode) {
+                case 191:
+                    var key = '?';
+                    break;
+                default:
+                    var key = String.fromCharCode(ev.keyCode)
+            }
+        }
+        var key = key.toLowerCase();
+        var ctrl = ev.ctrlKey || ev.metaKey;
+
         // undo with ctrl-z
-        if (ev.ctrlKey == true && ev.keyCode == 90) {
+        if (ctrl && key == 'z') {
             VIZ.netgraph.notify({ undo: "1" });
+            ev.preventDefault();
         }
         // redo with shift-ctrl-z
-        if (ev.ctrlKey == true && ev.shiftKey == true && ev.keyCode == 90) {
+        if (ctrl && ev.shiftKey && key == 'z') {
             VIZ.netgraph.notify({ undo: "0" });
+            ev.preventDefault();
         }
         // redo with ctrl-y
-        if (ev.ctrlKey == true && ev.keyCode == 89) {
+        if (ctrl && key == 'y') {
             VIZ.netgraph.notify({ undo: "0" });
+            ev.preventDefault();
         }
         // run model with spacebar
-        if (ev.keyCode == 32) {
-            sim.on_pause_click();
+        if (key == ' ') {
+            if (!ev.repeat) {
+                sim.on_pause_click();
+            }
+            ev.preventDefault();
         }
         // bring up help menu with ? 
-        if (ev.shiftKey && ev.keyCode == 63) {
+        if (key == '?') {
             self.callMenu();
+            ev.preventDefault();
         }
-        console.log(ev.keyCode);
     });
 }
 
--- a/nengo_viz/static/viz_hotkeys.js
+++ b/nengo_viz/static/viz_hotkeys.js
@@ -20,4 +20,4 @@ VIZ.Hotkeys = function () {
     });
 }
 
-VIZ.Hotkeys = new VIZ.Hotkeys();
+VIZ.hotkeys = new VIZ.Hotkeys();
--- a/nengo_viz/static/viz_hotkeys.js
+++ b/nengo_viz/static/viz_hotkeys.js
@@ -1,7 +1,8 @@
 VIZ.Hotkeys = function () { 
+    var self = this;
 
-    document.addEventListener('keydown', function(ev) {
-        // undoo with ctrl-z
+    document.addEventListener('keypress', function(ev) {
+        // undo with ctrl-z
         if (ev.ctrlKey == true && ev.keyCode == 90) {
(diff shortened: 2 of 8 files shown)


drasmuss:
In a lot of editors `ctrl-/` is the hotkey for commenting out a line of
code.  Might cause problems when the code editor gets added to the gui?

tcstewar:
> Aside from that the only weird thing is that the hotkey to pull up the info box is ctrl-/ because ctrl-h is reserved for history.

I'm fine not having a hot-key to pop it up, as long as there's a button on the toolbar.

studywolf:
oh that's a good idea! i'll do that now

studywolf:
done!

celiasmith:
Just made it a bit smaller, which seems appropriate for the little list.... should i bold the keys or something?

tbekolay:
In web stuff, bringing up a list of hotkeys is always done with a question mark. Works in Gmail, works in Github (like, literally this page, outside of a text box. Try it!)

studywolf:
I like the thing being smaller! I just put everything into a table so it can be aligned a bit better, I'm not crazy about it but it's a little cleaner than having no spacing...

bjkomer:
Whoa, never knew about that question mark thing. That is so useful!

drasmuss:
>In web stuff, bringing up a list of hotkeys is always done with a question mark. Works in Gmail, works in Github (like, literally this page, outside of a text box. Try it!)

:scream_cat:

studywolf:
Samesies!

xchoo:
TIL.

tbekolay:
You're welcome.

nengo_viz/disposable_js.py
```diff
@@ -35,6 +35,12 @@ def configmodal():
     js.append(show_modal_js())
     return '\n'.join(js)
 
+def helpmodal():
+    js = [add_modal_title_js('Hotkeys list')]
```
tbekolay: This function isn't actually using any information from the Python server, so there's no need to send work back to Python. Just do this all client-side (i.e., in Javascript).

nengo_viz/static/viz_hotkeys.js
```diff
@@ -0,0 +1,23 @@
...
+            VIZ.netgraph.notify({ undo: "0" });
+        }
+        // run model with spacebar
+        if (ev.keyCode == 32) {
+            sim.on_pause_click();
+        }
+    });
+}
+
+VIZ.Hotkeys = new VIZ.Hotkeys();
```
tbekolay: This should be `VIZ.hotkeys`? Or something to make it different from the `VIZ.Hotkeys` function, which is no longer accessible after this line (since you're overwriting the `VIZ.Hotkeys` function with the results of this call).

nengo_viz/disposable_js.py
studywolf: yeah i was just thinking about this last night, meant to do it then. i'll do it now!

nengo_viz/static/viz_modal.js
```diff
@@ -61,6 +61,20 @@ VIZ.Modal.prototype.text_body = function(text, type) {
     $p.append(document.createTextNode(text));
 }
 
+VIZ.Modal.prototype.help_body = function() {
+    this.clear_body();
+
+    $('.modal-dialog').addClass('modal-sm');
```
tbekolay: This class will persist on other modals. Can you add something like `$('.modal-dialog').removeClass('modal-sm')` to `VIZ.Modal.prototype.clear_body`? Also, it would be better to be specific (i.e., `this.$div.find('.modal-dialog')`, or capture this as `this.$dialog = this.$div.find('.modal-dialog').first()` in the constructor).

nengo_viz/static/viz_hotkeys.js
```diff
@@ -0,0 +1,23 @@
+VIZ.Hotkeys = function () { 
+
+    document.addEventListener('keydown', function(ev) {
+        // undoo with ctrl-z
```
tbekolay: `undoo` intentional or typo?

nengo_viz/static/viz_hotkeys.js
```diff
@@ -0,0 +1,23 @@
...
+        }
+        // redo with ctrl-y
+        if (ev.ctrlKey == true && ev.keyCode == 89) {
+            VIZ.netgraph.notify({ undo: "0" });
+        }
+        // run model with spacebar
+        if (ev.keyCode == 32) {
+            sim.on_pause_click();
+        }
+    });
```
tbekolay: Should we add the question mark for popping up the help here too?

nengo_viz/static/viz_hotkeys.js
```diff
@@ -0,0 +1,23 @@
+VIZ.Hotkeys = function () { 
+
+    document.addEventListener('keydown', function(ev) {
```
tbekolay: A bit of googling tells me that we should be listening to the `'keypress'` event rather than `'keydown'`. Apparently that plays much nicer with international keyboards (see, e.g., [this SO question](http://stackoverflow.com/questions/24995316/how-to-make-difference-between-question-mark-and-for-azerty-keyboard)).

nengo_viz/static/viz_modal.js
studywolf: done! just curious though what the benefit of having ```this.$div.find``` is!

studywolf:
OK! Think I got all of the comments from @tbekolay !

nengo_viz/static/viz_modal.js
tbekolay: `$('.modal-dialog')` selects every element on the page that has `modal-dialog` in the class list. If in the future we add multiple modals, then this is going to affect all of those modals. It can also be a bit slow to search the entire page if we have a lot of elements. By using `find` we're only searching for elements with the `modal-dialog` class that are a descendant of `this.$div`.

studywolf: Cool thanks!

tbekolay:
I just tried this out functionally. The question mark is cool, works out of the box! Had some issues with the others though...

- C-z and C-Z don't work for me, for whatever reason... I'm on a Mac, so that might be part of it? Mac people typically expect it to be Command-z and Command-Z also... ⌘-z and ⌘-Z. I have remapped my keyboard a bit, but a control should be a control regardless of where it is on the physical keyboard... I can help debug and tell you what my keycodes are if you like?
- Spacebar works for me, but fires multiple times for every time I press spacebar, so it ends up playing it momentarily and then pausing it. Does anyone else get this behavior? Perhaps we need to [debounce](http://davidwalsh.name/javascript-debounce-function) that function?

studywolf:
hmm i don't get either of those problems on linux in chrome and in
firefox...oh my...nothing works at all!

jgosmann:
* `?` to pull up the overview is neither working in Firefox or Chrome for me on OS X. But I also have the weirdest keyboard mapping of all (to type `?` I have to press `CapsLock + h` wherein `CapsLock` is actually `Mod3` which is roughly equivalent to `Alt`). But that shortcut works on Github ...
* Apart from that same problems as @tbekolay 
* Instead of having to entries for redo in the overview list, I would put only one there and list both key combinations in the same row.

studywolf:
so if i change it back to ```keydown``` everything but ? works for me in firefox

jgosmann:
> The onkeypress event is not fired for all keys (e.g. ALT, CTRL, SHIFT, ESC) in all browsers. To detect only whether the user has pressed a key, use the onkeydown event instead, because it works for all keys. —[Source](http://www.w3schools.com/jsref/event_onkeypress.asp)

:arrow_right: Use `keydown`

[`keyCode` is deprecated](https://developer.mozilla.org/en-US/docs/Web/Events/keydown) and turns out to be always (?) `0` in my Firefox. :arrow_right: Use `key` instead.

jgosmann:
Instead of magic numbers, it would be better to use [these constant strings](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key).

studywolf:
hmm key is not defined for keyboard events from what i can see in chrome...

jgosmann:
Then we can still use `keyCode` as fallback.

nengo_viz/static/viz_hotkeys.js
```diff
@@ -0,0 +1,35 @@
...
+        // redo with ctrl-y
+        if (ev.ctrlKey == true && ev.keyCode == 89) {
+            VIZ.netgraph.notify({ undo: "0" });
+        }
+        // run model with spacebar
+        if (ev.keyCode == 32) {
+            sim.on_pause_click();
+        }
+        // bring up help menu with ? 
+        if (ev.shiftKey && ev.keyCode == 63) {
```
jgosmann: Why is this checked for the `shift` key? Not all keyboard layouts (e.g., mine) require the `shift` key to be pressed to produce `?`.

studywolf:
how would that look? like 

```
if (typeof ev.key === 'undefined') {
    code = ev.key;
} else {
    code = ev.keyCode;
}
```
ish?

nengo_viz/static/viz_hotkeys.js
studywolf: international keyboards will be the death of me

jgosmann:
I never can remember how to correctly check for `undefined` is JS. I think, it might work with this oneliner?
`var code = ev.key || ev.keyCode;`

jgosmann:
We probably should allow either `Ctrl` or `Meta` for the hotkeys (Mac `Cmd` is treated as `Meta`).

bjkomer:
If that doesn't work, this one-liner should also work for checking undefined
```
var code = typeof ev.key !== 'undefined' ? ev.key : ev.keyCode;
```

jgosmann:
Fixed the hotkey functions. Should work at least in Firefox and Chrome on OS X and Windows (latter one untested). Space bar is "debounced" (but not other key combinations).

Key press events in JS are :hankey:

jgosmann:
Oh, and you can use now either `Ctrl` or `Cmd` (= `Meta`)

nengo_viz/static/viz_modal.js
```diff
@@ -61,6 +62,19 @@ VIZ.Modal.prototype.text_body = function(text, type) {
...
 }
 
+VIZ.Modal.prototype.help_body = function() {
+    this.clear_body();
+
+    this.$div.find('.modal-dialog').addClass('modal-sm');
+    var $body = $('<table width=100%>');
+    $body.append('<tr> <td>Play / pause</td> <td align="right">Spacebar</td></tr>');
+    $body.append('<tr> <td>Undo</td> <td align="right">Ctrl-z</td></tr>');
+    $body.append('<tr> <td>Redo</td> <td align="right">Ctrl-Shift-z || Ctrl-y</td></tr>');
```
tbekolay: Only huge nerds will know what `||` means. How about the word or? If you want it to stand out, you could put or in italics or bold.

tbekolay:
This is looking good now! Not on a mac so I can't test the command stuff, but everything works for me now except redoing with Ctrl+shift+z. Ctrl+y works, though.

jgosmann:
`C-S-z` is working for me on all systems and browsers. I will go ahead and merge this now.

Issue #113, opened 2015-04-28
Decide on text behviour when components/netgraph are shrunk


cchan987:

tcstewar:
My ideal, I think, is that the text gradually disappears and the graph takes up a larger and larger proportion of the space (i.e. the margins shrink) until the graph itself takes up all the area of the display.

tbekolay:
@celiasmith will add in an option to the options menu to toggle this behavior, to see if it's desired.

tbekolay:
This was done in #282. We'll ask people for their opinions once they try the options out!

PR #282, opened 2015-06-02
Font scaling options

celiasmith:
This allows the user to turn on/off scaling of fonts with zoom.  It also allows defining a % of base for the font size.  This addresses #113

### DIFF
--- a/nengo_viz/components/sim_control.py
+++ b/nengo_viz/components/sim_control.py
@@ -86,11 +86,8 @@ class SimControl(Component):
             client.write('reload')
             self.reload = False
         if self.send_config_options == True:
-            options = ["zoom"] #This is ignored by the client for now
-                               #But will be useful when saving/loading options
-            js_options = json.dumps(options)
             client.write('config' +
-                'VIZ.Toolbar.prototype.config_modal_show(%s);' % js_options)
+                'VIZ.Toolbar.prototype.config_modal_show();')
             self.send_config_options = False
 
     def get_status(self):
--- a/nengo_viz/components/sim_control.py
+++ b/nengo_viz/components/sim_control.py
@@ -86,9 +86,11 @@ class SimControl(Component):
             client.write('reload')
             self.reload = False
         if self.send_config_options == True:
-            from ..disposable_js import configmodal
-            javascript = configmodal()
-            client.write('config' + javascript)
+            options = ["zoom"] #This is ignored by the client for now
+                               #But will be useful when saving/loading options
+            js_options = json.dumps(options)
+            client.write('config' +
+                'VIZ.Toolbar.prototype.config_modal_show(%s);' % js_options)
             self.send_config_options = False
 
     def get_status(self):
--- a/nengo_viz/static/viz_netgraph.js
+++ b/nengo_viz/static/viz_netgraph.js
@@ -10,7 +10,8 @@ VIZ.NetGraph = function(parent, args) {
     this.scale = 1.0;          // global scaling factor
     this.offsetX = 0;          // global x,y pan offset 
     this.offsetY = 0;
-    this.zoomFonts = false;    // scale fonts when zooming
+    this.zoom_fonts = false;    // scale fonts when zooming
+    this.font_size = 100;       // font size as a percent of base
 
     this.svg_objects = {};     // dict of all VIZ.NetGraphItems, by uid
     this.svg_conns = {};       // dict of all VIZ.NetGraphConnections, by uid
@@ -299,20 +300,29 @@ VIZ.NetGraph.prototype.set_scale = function(scale) {
 
 
 VIZ.NetGraph.prototype.update_font_size = function(scale) {
-    if (this.zoomFonts) {
-        $('#main').css('font-size', 3 * this.scale + 'em');
+    if (this.zoom_fonts) {
+        $('#main').css('font-size', 3 * this.scale * this.font_size/100 + 'em');
     } else {
-        $('#main').css('font-size', '');
+        $('#main').css('font-size', this.font_size/100 + 'em');
     }
 }
 
-VIZ.NetGraph.prototype.set_zoomFonts = function(value) {
-    this.zoomFonts = value;
+VIZ.NetGraph.prototype.set_zoom_fonts = function(value) {
+    this.zoom_fonts = value;
     this.update_font_size();
 }
 
-VIZ.NetGraph.prototype.get_zoomFonts = function() {
-    return this.zoomFonts;
+VIZ.NetGraph.prototype.get_zoom_fonts = function() {
+    return this.zoom_fonts;
+}
+
+VIZ.NetGraph.prototype.set_font_size = function(value) {
+    this.font_size = value;
+    this.update_font_size();
+}
+
+VIZ.NetGraph.prototype.get_font_size = function() {
+    return this.font_size;
 }
 
 /** redraw all elements */
--- a/nengo_viz/static/viz_netgraph.js
+++ b/nengo_viz/static/viz_netgraph.js
@@ -10,7 +10,8 @@ VIZ.NetGraph = function(parent, args) {
     this.scale = 1.0;          // global scaling factor
     this.offsetX = 0;          // global x,y pan offset 
     this.offsetY = 0;
-    
+    this.zoomFonts = false;    // scale fonts when zooming
+
     this.svg_objects = {};     // dict of all VIZ.NetGraphItems, by uid
     this.svg_conns = {};       // dict of all VIZ.NetGraphConnections, by uid
 
@@ -298,12 +299,21 @@ VIZ.NetGraph.prototype.set_scale = function(scale) {
 
 
 VIZ.NetGraph.prototype.update_font_size = function(scale) {
-    // TODO: text resizing is disabled for now until we can figure out
-    // how best to have it work with large models
-
-    //$('#main').css('font-size', 3 * this.scale + 'em');
+    if (this.zoomFonts) {
+        $('#main').css('font-size', 3 * this.scale + 'em');
+    } else {
+        $('#main').css('font-size', '');
+    }
 }
 
+VIZ.NetGraph.prototype.set_zoomFonts = function(value) {
+    this.zoomFonts = value;
+    this.update_font_size();
+}
+
+VIZ.NetGraph.prototype.get_zoomFonts = function() {
+    return this.zoomFonts;
+}
 
 /** redraw all elements */
 VIZ.NetGraph.prototype.redraw = function() {
--- a/nengo_viz/static/viz_top_toolbar.css
+++ b/nengo_viz/static/viz_top_toolbar.css
@@ -73,23 +73,24 @@
     position: absolute;
 }
 
-#config-fontsize-group {
-    margin-left:20;
-}
-
 #config-fontsize.form-control {
     padding: 6px 6px;
     height: 35px;
 }
 
-#config-fontsize-group .control-label {
-    margin-left: -20px;
-}
-
-
-.input-group .input-group-addon{
+.input-group .input-group-addon {
     text-indent:0px;
-
 }
 
+#myModalForm .form-group {
+    margin-bottom:0px;
+}
 
+#myModalForm .checkbox {
+    padding-top:0px;
+    text-indent:10px;
+}
+
+.input-group.col-xs-2 {
+    left:20px;
+}
--- a/nengo_viz/static/viz_top_toolbar.css
+++ b/nengo_viz/static/viz_top_toolbar.css
@@ -72,3 +72,24 @@
     margin-top: 3em;
     position: absolute;
 }
+
+#config-fontsize-group {
+    margin-left:20;
+}
+
+#config-fontsize.form-control {
+    padding: 6px 6px;
+    height: 35px;
+}
+
+#config-fontsize-group .control-label {
+    margin-left: -20px;
+}
+
+
+.input-group .input-group-addon{
+    text-indent:0px;
+
+}
+
+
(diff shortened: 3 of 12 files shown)


jgosmann:
* [ ] Can't select text in the font size input box in Firefox.
* [ ] I think, it could be more useful to have 10% or 20% steps when clicking the arrows
* [ ] If I put the cursor into the text box and press enter, the state of 'allow fonts to zoom' is lost.

Do we need that input box at all? The browser should provide the same functionality. The only difference is that the browser scales also the top and bottom toolbar, but we could set a fixed font-size for those.

nengo_viz/components/sim_control.py
```diff
@@ -86,9 +86,11 @@ def update_client(self, client):
             client.write('reload')
             self.reload = False
         if self.send_config_options == True:
-            from ..disposable_js import configmodal
-            javascript = configmodal()
-            client.write('config' + javascript)
+            options = ["zoom"] #This is ignored by the client for now
+                               #But will be useful when saving/loading options
+            js_options = json.dumps(options)
```
jgosmann: I would prefer to exclude such code in the merge (we can keep it in a separate branch).

jgosmann:
Also wondering, would it be better if one can set the base font size in pt because it's a more common unit?

nengo_viz/static/viz_netgraph.js
```diff
@@ -8,9 +8,11 @@
  */
 VIZ.NetGraph = function(parent, args) {
     this.scale = 1.0;          // global scaling factor
-    this.offsetX = 0;          // global x,y pan offset 
+    this.offsetX = 0;          // global x,y pan offset
     this.offsetY = 0;
-    
+    this.zoom_fonts = false;    // scale fonts when zooming
+    this.font_size = 100;       // font size as a percent of base
```
jgosmann: I would prefer to store the actual fraction, i.e. `font_size` divided by 100. Then it is not necessary to remember the scaling factor whenever this value is used (only for the input form it has to be converted to a percentage).

nengo_viz/static/viz_netgraph.js
```diff
@@ -298,22 +300,40 @@ VIZ.NetGraph.prototype.set_scale = function(scale) {
 
 
 VIZ.NetGraph.prototype.update_font_size = function(scale) {
-    // TODO: text resizing is disabled for now until we can figure out
-    // how best to have it work with large models
+    if (this.zoom_fonts) {
+        $('#main').css('font-size', 3 * this.scale * this.font_size/100 + 'em');
```
jgosmann: Do we want to keep that factor of 3 or should we make the font-size match independent of the zoom mode on scale = 1.

nengo_viz/static/viz_top_toolbar.css
```diff
@@ -68,3 +68,24 @@
     margin-top: 3em;
     position: absolute;
 }
+
+#config-fontsize-group {
+    margin-left:20;
```
jgosmann: Missing a unit ...

nengo_viz/static/viz_top_toolbar.css
```diff
@@ -68,3 +68,24 @@
...
+}
+
+#config-fontsize.form-control {
+    padding: 6px 6px;
+    height: 35px;
```
jgosmann: It might be better to use `em` here.

nengo_viz/static/viz_top_toolbar.css
```diff
@@ -68,3 +68,24 @@
...
+}
+
+#config-fontsize-group .control-label {
+    margin-left: -20px;
```
jgosmann: Is this undoing the effect of line 73? Maybe both should be deleted?

nengo_viz/static/viz_top_toolbar.js
```diff
@@ -82,8 +82,33 @@ VIZ.Toolbar.prototype.reset_model_layout = function () {
...
-VIZ.Toolbar.prototype.start_modal = function () {
-    sim.ws.send('config')
+ *  call to server to call config_modal_show with config data. */
+VIZ.Toolbar.prototype.config_modal = function () {
+    sim.ws.send('config');  //Doing it this way in case we need to save options to a file later
+}
+
+VIZ.Toolbar.prototype.config_modal_show = function(options) {
+    var self = this;
+    console.log(options); //Options are ignored for now
```
jgosmann: I think, this should be removed because it's just debug output.

jgosmann:
* [ ] The clickable arrows are pretty small (hard to hit).
* [ ] The spacing between the input box and the 'allow font zoom' checkbox feels too large. These are related options, but they space makes them look unrelated.

jgosmann:
We might no need to fix all of that before the summer school, though.

nengo_viz/static/viz_top_toolbar.css
celiasmith: bootstrap defines these in px

nengo_viz/components/sim_control.py
celiasmith: done

nengo_viz/static/viz_netgraph.js
celiasmith: Seems to me this adds unnecessary complexity (stripping decimals, etc.), and it's a documented unit.

nengo_viz/static/viz_netgraph.js
celiasmith: I'm fine either way... in both cases we'll be putting parameters in somewhere (either here or various .css files).

nengo_viz/static/viz_top_toolbar.css
celiasmith: fixed

nengo_viz/static/viz_top_toolbar.css
celiasmith: No it isn't. This is positioning the label and the text input box separately.

celiasmith:
> The spacing between the input box and the 'allow font zoom' checkbox feels too large. These are related options, but they space makes them look unrelated.

I agree.  I was a bit torn about playing with spacing too much, since right now it's consistent between all modals.  These are defined largely by the bootstrap defaults atm.

tcstewar:
> Do we need that input box at all? The browser should provide the same functionality. The only difference is that the browser scales also the top and bottom toolbar, but we could set a fixed font-size for those.

There'll also generally be a difference between the font size for the NetGraph and the font size for the code editor....  In presentations I often want to make the code font bigger, but not the NetGraph.

celiasmith:
> The clickable arrows are pretty small (hard to hit).

This can only be changed on chrome it seems.  So, i'll just leave it to be consistent across browsers.

celiasmith:
> can we get the 'as percentage' text right of the input field (instead of weirdly aligned below)?

This is the bootstrap/validator default alignment for such text (and for errors).  So, i'll leave it to be consistent with all the other modals.

nengo_viz/static/viz_netgraph.js
jgosmann: [Here the unit was also documented and it still went wrong.](https://en.wikipedia.org/wiki/Mars_Climate_Orbiter) Good thing we're not building space crafts. :rocket: :boom:

nengo_viz/static/viz_top_toolbar.css
jgosmann: If bootstrap defines them, why do we have to redefine them?

celiasmith:
> Can't select text in the font size input box in Firefox.

This seems to be some weird firefox bug, and i can't track it down.  You can't select anything in the modals with the mouse (the keyboard works fine).

tbekolay:
I don't think this is a firefox bug, I think we deliberately set everything to be unselectable? There's something in the CSS about not being able to select text... yeah. [Here it is](https://github.com/nengo/nengo_viz/blob/cd3edbb381b4b8d1a444de1743b3523e837549db/nengo_viz/static/main.css#L29). This should probably be more specific.

nengo_viz/static/viz_netgraph.js
celiasmith: guess we shouldn't have units... no, wait a sec :stuck_out_tongue_winking_eye:

celiasmith:
> I think we deliberately set everything to be unselectable?

I guess chrome is ignoring us then... and so is firefox when we use the keyboard?  Soooo confused :confused: ... I'll reset the css for this element and see.

celiasmith:
No luck with the css... styling any id, class, or tag seems to style the entire document, so either it's full of annoying text selecting mouse cursors in netgraph, or you can't select text with the mouse (in firefox and safari).  I think the latter is the best of these bad options.

I think this is ready to go unless there are other outstanding issues i'm forgetting.

jgosmann:
[That CSS property seems to be non-standard with a big warning banner.](https://developer.mozilla.org/en-US/docs/Web/CSS/user-select) Makes me wonder whether we should be using it at all? And it seems that Firefox can re-enable selection only with `-moz-user-select:text`.

jgosmann:
Moreover, `user-select` does not change the cursor in Firefox, so we actually should set the cursor to `default` wherever we have non-selectable text, but in a way which does not break the resize cursors.

celiasmith:
I created another branch (text-selecting) to look at fixing the selection stuff, as that's beyond what this is about.

celiasmith:
PR #284 takes care of the text selection problem in the modal.

nengo_viz/static/viz.css
```diff
@@ -45,4 +45,4 @@
 
 .graph .label:active {
     cursor: ns-resize;
-}
\ No newline at end of file
+}
```
tbekolay: Looks like this is only whitespace changes... amend this out?

nengo_viz/static/viz_netgraph.css
```diff
@@ -41,5 +41,5 @@ g.net rect{
 .netgraph g.passthrough text {
     text-anchor: middle;
     dominant-baseline: text-before-edge;
-    }    
+    }
```
tbekolay: Also just whitespace changes in this file.

nengo_viz/static/viz_top_toolbar.css
```diff
@@ -68,3 +68,25 @@
...
+#config-fontsize.form-control {
+    padding: 6px 6px;
+    height: 35px;
+}
+
+.input-group .input-group-addon {
+    text-indent:0px;
+}
+
+#myModalForm .form-group {
```
tbekolay: These should probably be in `viz_modal.css`?

celiasmith: done!

tbekolay:
Tried this out. Issues I had:

- I tried to make the font size 125%, but that failed validation. It would only let me put in multiples of 10. Feels weird, should allow any number greater than 0...
- There's a text input box here that's not connected to anything? I think you're missing a call to `VIZ.modal.clear_body` in the config body function.

The actual text scaling checkbox works perfectly though ;) It would be kind of nice if it updated the model when you click the checkbox, but definitely not essential.

celiasmith:
> I tried to make the font size 125%, but that failed validation

It's not clear how to do this, I've submitted an issue to the validator github.  I can't override the default validator.  So, either I turn off the clickable by 10, or you have to do steps of 10.  I've left it for now.

The text input was just a dumb mistake :unamused: ... instant update for another day... (it should update immediately on exit from the modal).

celiasmith:
BTW, the developer responded: 

"This isn't really possible with the way the validator is currently written. The only viable solution at the moment is to add an option like data-disable-native=true, but since the native validators are all bundled together, you'd lose them all (required/min/max/etc.). I guess that's not a huge issue since you'll be adding a custom validator and can check whatever you need to in there, but it's kinda coupling and not ideal."

I actually figured out how to disable the native validator, but then you have to write validators for *all* classes.  I thought about doing this because I only needed to add the checkbox, but this will be crazy if the options expand.

tbekolay:
> data-disable-native=true

Yeah... I think it's maybe fine to just have it update in steps on 1? Can't people just click and hold? Can we change the rate at which it gets bigger when you hold?

celiasmith:
So I implemented your other suggest (immediate apply changes)... good idea!  

Yes, if you hold, it will spin (I don't think you can change the speed).  I'm thinking steps of 5 maybe?

celiasmith:
Unfortunately the immediate update doesn't work if you hold down the arrows... so this makes steps of 5 seem good, since you can see the change on each click (can't really with 1).

celiasmith:
OK, I fixed the immediate update to now work when you hold the button down.  So, I also changed the step size to one, so you can watch it grow as you hold the button down.

jgosmann:
If we keep the immediate update, the `Cancel` button should be removed, because it doesn't undo the changes.

jgosmann:
I lied ... if there's a validation error only `Cancel` allows to close the window. But that doesn't change that he button does not have the expected effect.

Otherwise, this looks pretty good now. :sparkles:

tbekolay:
This LGTM. Any objections to me cleaning up the history and merging?

celiasmith:
I just made the change jan suggested of actually resetting all properties on cancel (despite immediate update).  So... a quick review of that, and this should be ready to go.

celiasmith:
I was going to say, it won't work with !==, but does with !=, but you're right, being more explicit 'undefined' is better....

tcstewar:
> I was going to say, it won't work with !==, but does with !=, but you're right, being more explicit 'undefined' is better....

I think that's because it's supposed to be 

```
if (cancel_function !== undefined)
```

```undefined``` is a keyword in javascript.

jgosmann:
[How to check for undefined.](https://stackoverflow.com/questions/3390396/how-to-check-for-undefined-in-javascript)

tbekolay:
Okay cool; I'm going to rebase and fix up the history, then we can do a final lookover before merging.

tbekolay:
OK, rebased and stripped out whitespace changes. If there are no weird unforeseen interactions with new stuff in master, this should be good to go!

PR #283, opened 2015-06-03
Time bar now redraws on resizing

### LINKED ISSUE #174: Bottom time bar does not handle horizontal resize correctly
The left end stays fixed, but both ends should stay fixed and the time bar should rescale.

tcstewar:
Fixes #174


tbekolay:
Just tested this out; fixes the issue for me! And the diff is pretty minimal, so looks ready to merge IMO.

PR #284, opened 2015-06-03
Fixing css styles for cursors

celiasmith:
Seems to work on firefox, chrome and safari as expected.  No selectable text on netgraph/components, text cursor in modals input boxes.  No interference on resizing. I found the magic 'pointer-events:none' css which makes this easier.

### DIFF
--- a/nengo_viz/static/main.css
+++ b/nengo_viz/static/main.css
@@ -19,18 +19,9 @@ body {
 .graph {
     background: #eeeeee;
 }
-    
+
 .line {
     fill: none;
     stroke: black;
-    stroke-width: 1.5px;    
-}
-
-* {
-    -webkit-touch-callout: none;
-    -webkit-user-select: none;
-    -khtml-user-select: none;
-    -moz-user-select: none;
-    -ms-user-select: none;
-    user-select: none;
+    stroke-width: 1.5px;
 }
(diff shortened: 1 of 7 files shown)


jgosmann:
In Firefox:
* [ ] Getting text select cursor in menus
* [ ] No text select cursor when typing in slider value

celiasmith:
> No text select cursor when typing in slider value

I don't think you can do this one because you want to be able to grab the slider handle.  Presumably you don't want a text cursor every time you go over the handle.

jgosmann:
I think, it should be possible because the HTML changes from normal text to `input` while you can type into it.

celiasmith:
So you want it to change after you've double clicked, but not before?

jgosmann:
I only do a single click ... but essentially yes and I think it can be done with CSS (no JS)

celiasmith:
> I think, it should be possible because the HTML changes from normal text to input while you can type into it.

I got it to work on chrome and safari, but firefox refuses to cooperate.  Any ideas more than welcome.

jgosmann:
I'll see what I can do

jgosmann:
Figured it out.

celiasmith:
Yup, works for me.  You want to merge?

PR #286, opened 2015-06-03
Removing the old slider when making a new one

s72sue:
This change has been made on the server side, because it wasn't possible to implement it on the client side since we do not have access to the parent 'node' on the client side.

### DIFF
--- a/nengo_viz/components/action.py
+++ b/nengo_viz/components/action.py
@@ -126,6 +126,11 @@ class CreateGraph(Action):
         cls = getattr(nengo_viz.components, self.type + 'Template')
         self.template = cls(self.obj, **kwargs)
 
+        # Remove any existing sliders associated with the same node
+        for component in self.net_graph.viz.components:
+            if isinstance(component, nengo_viz.components.slider.Slider) and component.node == self.obj: 
+                self.send('delete_graph', uid=component.uid)
+
         self.act_create_graph()
 
     def act_create_graph(self):
--- a/nengo_viz/components/action.py
+++ b/nengo_viz/components/action.py
@@ -128,7 +128,8 @@ class CreateGraph(Action):
 
         # Remove any existing sliders associated with the same node
         for component in self.net_graph.viz.components:
-            if isinstance(component, nengo_viz.components.slider.Slider) and component.node == self.obj: 
+            if (isinstance(component, nengo_viz.components.slider.Slider)
+                    and component.node is self.obj):
                 self.send('delete_graph', uid=component.uid)
 
         self.act_create_graph()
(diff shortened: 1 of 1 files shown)


nengo_viz/components/action.py
```diff
@@ -126,6 +126,11 @@ def __init__(self, net_graph, uid, type, x, y, width, height, **kwargs):
         cls = getattr(nengo_viz.components, self.type + 'Template')
         self.template = cls(self.obj, **kwargs)
 
+        # Remove any existing sliders associated with the same node
+        for component in self.net_graph.viz.components:
+            if isinstance(component, nengo_viz.components.slider.Slider) and component.node == self.obj:
```
jgosmann: I think, the equals operator `==` should probably be the `is` operator here?

I dislike it, that we have to test for a specific class here. I think, it would be better to introduce an attribute on the components which indicates whether multiple instances are allowed and that we check for that attribute here.

s72sue:
You are right, I have replaced  ==  with the 'is' operator.  

I agree that introducing an attribute would be better, but this is just a temporary fix. According to @tcstewar (correct me if I am wrong), ideally we wanted to handle this on the client side but that will be a big change since we do not have access to the parent 'node' on the client side right now and we will probably need to pass that info from the server side. This can be implemented after the summer school.

nengo_viz/components/action.py
```diff
@@ -126,6 +126,11 @@ def __init__(self, net_graph, uid, type, x, y, width, height, **kwargs):
         cls = getattr(nengo_viz.components, self.type + 'Template')
         self.template = cls(self.obj, **kwargs)
 
+        # Remove any existing sliders associated with the same node
+        for component in self.net_graph.viz.components:
+            if isinstance(component, nengo_viz.components.slider.Slider) and component.node is self.obj:
```
tcstewar: This line should be split to stay within the 80 character limit:

```python
if (isinstance(component, nengo_viz.components.slider.Slider) and 
     component.node is self.obj):
```

tcstewar:
Looks good to me, other than the 80 char thing I noted above...

tbekolay:
Just checked this out; looks good to me! Pushed a style fix commit. Would recommend that the style commit be squashed into the previous commit... @tcstewar if you think this is ready, I can squash and merge.

tcstewar:
> if you think this is ready, I can squash and merge.

Yup, I think it's ready.  :)  Thanks!

tbekolay:
OK, doing so. I noticed some weird behavior with the undo / redo stuff... It's possible to add a slider a few times, hit undo/redo a bunch of times, and end up with two sliders. I think this is pretty much only behavior people will do if they're jerks, so I think we should merge this in and I'll file a bug for that.

s72sue:
Thanks.

I'll take a look at the undo/redo stuff.

PR #289, opened 2015-06-03
Fixed Python3 errors

tcstewar:


nengo_viz/swi.py
```diff
@@ -465,7 +465,7 @@ def handle_error(self, request, client_address):
                 exc_value.args[0] == errno.EBADF):
             return  # Probably caused by a server shutdown
         else:
-            print exc_type, exc_value, dir(exc_value)
+            print(exc_type, exc_value, dir(exc_value))
```
jgosmann: Maybe `from __future__ import print_function` should be added?

nengo_viz/swi.py
```diff
@@ -169,7 +169,7 @@ def swi_favicon_ico(self):
 
     def do_POST(self):
         self.clear_headers()
-        data = self.rfile.read(int(self.headers['Content-Length']))
+        data = self.rfile.read(int(self.headers['Content-Length'])).decode('ascii')
```
jgosmann: Should we really assume 'ascii' here?

tcstewar: Possibly not, but the whole rest of the code assumes it. 

(and it might even be correct to assume, because that's parsing the POST request, which has some rules about its format.  But I think it's a separate issue to actually check all the unicode issues.)

nengo_viz/swi.py
tcstewar: We don't do that anywhere else, and it's only needed if someone is using Python 2.5 or earlier.

jgosmann: Maybe we should do it elsewhere too. Or actually I think we should get rid of all print statements and replace them with something which gives us more control about when and where to stuff is printed (but not before the summer school).

The print statement here will behave differently depending on the existence of the `__future__` import (and without depending on the Python version, I suppose):
```python
In [1]: print("a", "b")
('a', 'b')

In [2]: from __future__ import print_function

In [3]: print("a", "b")
a b
```

tcstewar: > Maybe we should do it elsewhere too. Or actually I think we should get rid of all print statements and replace them with something which gives us more control about when and where to stuff is printed (but not before the summer school).

I totally agree. :)

tbekolay:
Can confirm, this branch works for me on Python 3! Code seems good to me for now (but yeah, we should use logging in the future; will make an issue for that that we can deal with after release).

PR #292, opened 2015-06-04
No Spikes shown for an ensemble in direct mode

s72sue:
This is a very temporary fix on the server side to prevent the exception being thrown right now. This code should later be removed from the server side and the fix should be made on the client side (this would require passing the information regarding the neuron type of an ensemble to the client).

The spike plot will still show up , but there will be no spikes displayed in direct mode

### DIFF
--- a/nengo_viz/components/raster.py
+++ b/nengo_viz/components/raster.py
@@ -14,14 +14,14 @@ class Raster(Component):
         self.obj = obj.neurons
         self.data = collections.deque()
         self.label = viz.viz.get_label(obj)
-        self.max_neurons = self.obj.size_out
+        self.max_neurons = obj.n_neurons
         if n_neurons is None:
-            n_neurons = min(self.obj.size_out, 10)
+            n_neurons = min(self.max_neurons, 10)
         self.n_neurons = n_neurons
 
     def add_nengo_objects(self, viz):
         with viz.model:
-            self.node = nengo.Node(self.gather_data, size_in=self.obj.size_out)
+            self.node = nengo.Node(self.gather_data, size_in=self.max_neurons)
             if 'spikes' in self.neuron_type.probeable:
                 self.conn = nengo.Connection(self.obj, self.node, synapse=None)
 
@@ -30,11 +30,8 @@ class Raster(Component):
         if 'spikes' in self.neuron_type.probeable:
             viz.model.connections.remove(self.conn)
 
-    def gather_data(self, t, x=None):
-        if x is None:
-            indices = []
-        else:
-            indices = np.nonzero(x[:self.n_neurons])[0]
+    def gather_data(self, t, x):
+        indices = np.nonzero(x[:self.n_neurons])[0]
         data = struct.pack('<f%dH' % len(indices), t, *indices)
         self.data.append(data)
 
--- a/nengo_viz/components/raster.py
+++ b/nengo_viz/components/raster.py
@@ -10,9 +10,7 @@ from nengo_viz.components.component import Component, Template
 class Raster(Component):
     def __init__(self, viz, config, uid, obj, n_neurons=None):
         super(Raster, self).__init__(viz, config, uid)
-        if getattr(obj, 'neuron_type') is not None:
-            type = (getattr(obj, 'neuron_type'))
-        self.neuron_type = type
+        self.neuron_type = obj.neuron_type
         self.obj = obj.neurons
         self.data = collections.deque()
         self.label = viz.viz.get_label(obj)
--- a/nengo_viz/components/raster.py
+++ b/nengo_viz/components/raster.py
@@ -10,6 +10,9 @@ from nengo_viz.components.component import Component, Template
 class Raster(Component):
     def __init__(self, viz, config, uid, obj, n_neurons=None):
         super(Raster, self).__init__(viz, config, uid)
+        if getattr(obj, 'neuron_type') is not None:
+            type = str(getattr(obj, 'neuron_type'))
+        self.neuron_type = type
         self.obj = obj.neurons
         self.data = collections.deque()
         self.label = viz.viz.get_label(obj)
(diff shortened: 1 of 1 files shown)


nengo_viz/components/raster.py
```diff
@@ -10,6 +10,9 @@
 class Raster(Component):
     def __init__(self, viz, config, uid, obj, n_neurons=None):
         super(Raster, self).__init__(viz, config, uid)
+        if getattr(obj, 'neuron_type') is not None:
```
tbekolay: I think it's impossible for `neuron_type` to be `None`, assuming that you can only make rasters for ensembles. So you should just be able to do `self.neuron_type = obj.neuron_type`. Then, below, rather than comparing strings, just check `if isinstance(self.neuron_type, nengo.Direct)`. In case we end up changing the string representation of `nengo.Direct` somewhere down the line!

tbekolay:
Tried this out, works great! I actually think this is a good candidate for a long-term fix, since it is nice to not force users to change their nengo_viz config when they change the neuron type. No spikes showing up is actually exactly what's happening, so I think this gives people the right mental model.

s72sue:
I agree that it gives the right mental model. However, the only problem is
that we are checking for a specific type here (so its not scalable).  But
we might not need it to be scalable in this case since direct mode would
probably ever be the only mode for not showing the spikes?

jgosmann:
Does `NeuronType.probable` contain the information we want?

tbekolay:
Yeah, you could check `'spikes' in obj.neuron_type.probeable`. Though, I tried the raster with LIFRate, and it does do something sorta correct... but you're right, it's misleading, so I think that's the right check.

s72sue:
Ohhh I see, I will change that to use obj.neuron_type.probeable and check
for 'spikes' in the list.

s72sue:
Okay changed it. This looks better to me now.

I had created another issue #293 to later move this to the client side, but if there is consensus that we want to keep it this way, I can get rid of that issue.

nengo_viz/components/raster.py
s72sue: What should happen if the neuron_type is None? Do we still want to show spikes?

tcstewar: > What should happen if the neuron_type is None? Do we still want to show spikes?

I'm pretty sure it's never None.  The validation system won't let that happen.

s72sue: If it won't be none, then I think checking the probeable list generalizes better than checking for a specific type (Direct()) . 

Any thoughts?

tbekolay: Yeah, we validate so that it can never be None. If it is none, the simulator will never build anyway, so we'll get a ton of issues, not just this one. I say check the probeable list.

s72sue: In that case, I think checking for None here is redundant. I will get rid of it.

tbekolay:
This looks great to me! I'm going to merge.

tbekolay:
OK, rebased, but encountered a weird bug when testing it out. Here's what I did:

1. Opened `examples/net.py` and added a value plot and spike raster. Runs great.
2. Changed line 9 to make `result` use `nengo.Direct`.
3. When I press play, the simulation rebuilds itself (very cool).
4. The value plot is correct, looks like direct mode, but the spike raster stays at its previous state, instead of clearing itself.

Screenshot:

screen shot 2015-06-04 at 12 43 40 pm

Is there an easy fix for this, or will it take a while? If there's an easy fix I'll hold off on merging; if not, I'll merge now and make a separate issue for that.

tcstewar:
> Is there an easy fix for this, or will it take a while? If there's an easy fix I'll hold off on merging; if not, I'll merge now and make a separate issue for that.

Merge now and make a separate issue, I think....

s72sue:
I will have to look into what might be causing it. I think its better to
create a separate issue.

tbekolay:
I figured out a fix; basically, we still make the node when there are no spikes to be recorded, but we don't make the connection. @s72sue, let me know what you think!

s72sue:
The same behavior is also reproducible on switching to LIFRate(). So I am not sure whether its related to this change.

tbekolay:
Can you pull the latest version of this branch and test it out? Would probably be easiest to delete your local `direct_mode_spikes` branch, pull, and then checkout `direct_mode_spikes` :)

s72sue:
Sure :)

jgosmann:
> Can you pull the latest version of this branch and test it out? Would probably be easiest to delete your local direct_mode_spikes branch, pull, and then checkout direct_mode_spikes :)

Or if you're already on the local `direct_mode_spikes` branch just do:
```
git fetch
git reset --hard origin/direct_mode_spikes
```

s72sue:
Okay, it looks great!

tbekolay:
OK cool :D will merge.

s72sue:
Oh Wait

s72sue:
Just noticed that if I start with direct mode and switch to LIF later, the
spikes do not show up

tbekolay:
Will investigate!

s72sue:
Actually, never mind, I had accidently switched to LIFRate instead of
LIF().

It looks good, you can go ahead with the merge

tbekolay:
OK, so it's okay if you start off with a `LIFRate`, but not `Direct`. I believe this is because the y-axis range of the spike raster is not being updated. I'll hack in a fix.

tbekolay:
Fixed. Now, instead of using `size_out` we use `n_neurons`, so the node always gets created with the same `size_in`, fixing the y-axis issue in the raster. Also makes the `gather_data` function simpler.

FYI, I initially tried to index into the neurons (e.g., `self.conn = nengo.Connection(self.obj[:self.n_neurons], self.node)`) but this gave me an error... Might be a Nengo bug. Would be good if it worked though, as right now we're transmitting way more data than we need to.

Anyhow, check this out @s72sue and let me know if it works for you!

s72sue:
Looks good.
I am able to switch from Direct mode to LIF and LIFRate. In case of LIFRate the spikes do not show up which is expected (since it doesn't have spikes in its probeable list)

tbekolay:
Awesome, I think this is ready to go then. Merging!

PR #300, opened 2015-06-04
Fixed shutting down with open connections

### LINKED ISSUE #299: ipynb integration breaks server reloading
Something in the IPython system means that the server shuts down immediately on a page reload or a browser window close (even if there are other browser windows open).

To replicate:
- start nengo_viz
- hit reload

or 
- start nengo_viz
- open a second tab to it
- close one of the tabs

tcstewar:
Fixes #299


tcstewar:
@jgosmann do you think this solves the problem?  Or is it deeper than that?

PR #301, opened 2015-06-04
Use signal handlers for shutdown confirmation.

### LINKED ISSUE #234: Shutdown with double Ctrl+C
An IPython notebook server can be shut down by pressing Ctrl+C twice (in addition to Ctrl+C, y, enter). Doing this with nengo_viz just kills the server without shutting it down properly.

jgosmann:
Closes #234.

### DIFF
--- a/nengo_viz/swi.py
+++ b/nengo_viz/swi.py
@@ -56,6 +56,7 @@ import os
 import random
 import re
 import select
+import signal
 import socket
 try:
     import SocketServer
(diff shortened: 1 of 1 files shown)


tbekolay:
Just tried this; works great on Mac OS X! Obviously the real test is Windows though, which I don't have access to. Any Windows users able to test this branch?

jgosmann:
I tested it in my VM and worked. (Doesn't ask for confirmation on Windows because it does not support select on anything that is not a socket. This is in line with what ipython notebook does.)

tcstewar:
I'm happy with it not asking for confirmation on windows.

tbekolay:
Kay, ready to merge then?

PR #303, opened 2015-06-04
Fixed shutting down with open connections

### LINKED ISSUE #299: ipynb integration breaks server reloading

jgosmann:
Fixes #299.

### DIFF
--- a/nengo_viz/server.py
+++ b/nengo_viz/server.py
@@ -151,7 +151,8 @@ class Server(swi.SimpleWebInterface):
                 time.sleep(2)
 
                 # if there are no simulations left, stop the server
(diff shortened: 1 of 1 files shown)


jgosmann:
Alternative proposal to #300

tcstewar:
Just to check, does this mean that you also want it to be checking for ```count_sims()==0``` in the IPython case?

PR #306, opened 2015-06-04
Make sure all data is sent.

jgosmann:
Am I the only one reading documentation?

[socket.send](https://docs.python.org/3/library/socket.html#socket.socket.send):
> Applications are responsible for checking that all data has been sent; if only some of the data was transmitted, the application needs to attempt delivery of the remaining data.

Addresses the problems experienced in #152.

### DIFF
--- a/nengo_viz/swi.py
+++ b/nengo_viz/swi.py
@@ -227,7 +227,7 @@ class SimpleWebInterface(BaseHTTPServer.BaseHTTPRequestHandler):
         key = key.encode('ascii')
         resp_data = (HSHAKE_RESP %
                      base64.b64encode(hashlib.sha1(key).digest()).decode('ascii'))
-        client.socket.send(resp_data.encode('ascii'))
+        client.socket.sendall(resp_data.encode('ascii'))
         client.set_blocking(False)
 
         self.user = self.get_user_from_cookie()
@@ -693,8 +693,8 @@ class ClientSocket(object):
         if not binary:
             data = data.encode('ascii')
         try:
-            self.socket.send(header)
-            self.socket.send(data)
+            self._sendall(header)
+            self._sendall(data)
         except socket.error as e:
             if e.errno == 32:  # Broken pipe
                 raise SocketClosedError("Cannot write to socket.")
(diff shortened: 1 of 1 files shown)


jgosmann:
This might have a minor performance impact. At some point (not before the summer school) we should probably introduce a mechanism to drop data if it's produce too quickly. But we definitely should not allow partial messages to be transmitted.

Another future performance optimization would be, to make the select call on all sockets, so that we can write pending data to whichever socket is ready first.

tbekolay:
This LGTM but I still think we should leave sparkle plots out of the release. This still seems like it'll make things more robust, so probably worth merging ATM... are there any other places where we call `socket.send`?

jgosmann:
There was one other place ...

jgosmann:
ah no, there are more

jgosmann:
I hope, that's all now.

nengo_viz/swi.py
```diff
@@ -220,7 +220,7 @@ def handle_ws_request(self, args, db):
         key = key.encode('ascii')
         resp_data = (HSHAKE_RESP %
                      base64.b64encode(hashlib.sha1(key).digest()).decode('ascii'))
-        client.socket.send(resp_data.encode('ascii'))
+        client.socket.sendall(resp_data.encode('ascii'))
```
jgosmann: socket is still in nonblocking mode here, so we can use `socket.sendall` instead of `ClientSocket._sendall`

tbekolay:
OK. I'm going to merge this to get it in this initial release, as hopefully it will fix some subtle bugs...

tbekolay:
This does have a performance impact for me; `2d.py` runs about 2.5x on master, about 1.8x on this branch. That's a simple model though, so I'm not sure if it's that important?

jgosmann:
I don't know, right now it can randomly break in theory.

tbekolay:
Fair enough! In it goes.

jgosmann:
Ah just changed it

tbekolay:
Ah whoops. Make another PR I guess

tcstewar:
> Am I the only one reading documentation?

Argh!  Thank you again, and I apologize...  My only defence is that I wrote that code almost 10 years ago....  Back in the days that I used 2-space tabs and horrible horrible spacing....

PR #312, opened 2015-06-04
Fix the spacebar problem with the editor

tcstewar:
Here's a quick fix for the spacebar problem.

It's not perfect because you can't click on the NetGraph to change focus (the text editor keeps the focus).  But if you click on the top or bottom bar, hotkeys will now work as expected.


tcstewar:
Oops, I realized my first version of this fix stopped all the Ctrl hotkeys from doing anything while you are on the editor too....  I just changed it so only ' ' and '?' do this check.

jgosmann:
Added a commit so that you can even click the netgraph to remove the focus from the editor

PR #321, opened 2015-06-04
Use correct frame of reference for zooming.

### LINKED ISSUE #313: Adjust zoom mouse-cursor position to account for top toolbar
When scrolling, we compute where the mouse cursor is and use that to make it so you are zoomin in and out of that point.  However, that point seems to be a bit off.  My best guess right now is that we're not taking into account the top toolbar.

Also, we need to make sure that ```#netgraph``` is correctly the full visual height available.

jgosmann:
Fixes #313.


tcstewar:
Nice!

PR #324, opened 2015-06-05
Resize options

celiasmith:
Added an option in the configure modal to choose between having fixed aspect ratio objects on resizing of the canvas or not (e.g. when you open/close the editor).

### DIFF
--- a/nengo_gui/static/netgraph.js
+++ b/nengo_gui/static/netgraph.js
@@ -15,7 +15,7 @@ Nengo.NetGraph = function(parent, args) {
     this.offsetY = 0;
     this.zoom_fonts = false;    // scale fonts when zooming
     this.font_size = 100;       // font size as a percent of base
-    this.aspect_resize = true;  //preserve aspect ratios on window resize
+    this.aspect_resize = false;  //preserve aspect ratios on window resize
 
     this.svg_objects = {};     // dict of all Nengo.NetGraphItems, by uid
     this.svg_conns = {};       // dict of all Nengo.NetGraphConnections, by uid
--- a/nengo_gui/static/netgraph.js
+++ b/nengo_gui/static/netgraph.js
@@ -15,6 +15,7 @@ Nengo.NetGraph = function(parent, args) {
     this.offsetY = 0;
     this.zoom_fonts = false;    // scale fonts when zooming
     this.font_size = 100;       // font size as a percent of base
+    this.aspect_resize = true;  //preserve aspect ratios on window resize
 
     this.svg_objects = {};     // dict of all Nengo.NetGraphItems, by uid
     this.svg_conns = {};       // dict of all Nengo.NetGraphConnections, by uid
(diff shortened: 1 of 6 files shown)


tcstewar:
With @jgosmann 's comments addressed, this looks fairly good to me.  I do think the default should be to have this be off, though, as I find it is much slower to update than the old method, leading to jerky resizing.

celiasmith:
@tcstewar actually, I intended to leave it off by default (thanks for the catch).  However, I don't think I get a difference in jerkiness. (There is little difference in the number of calculations; basically, three multiplies per item, which is tiny compared to what's already going on).  That said, the graphs in particular do move differently in a way that looks less continuous on smooth resizes, I'd agree.  Switching between edit and non edit seems equally smooth.  Anyway, we can see what folks think!

jgosmann:
I don't get a difference in jerkiness either.

tcstewar:
> There is little difference in the number of calculations; basically, three multiplies per item, which is tiny compared to what's already going on)

You're calling ```NetGraphItem.set_size()``` inside a loop in ```NetGraph.on_resize()```.  That's extremely expensive, since ```set_size``` has to reposition both the item and all of the connections into and out of the item, and do the same thing to any children inside the item.  And since it's inside a loop, this means that in any large network a whole bunch of things are going to end up getting repositioned multiple times.

celiasmith:
Sorry, just pushed, thought i did that yesterday...

nengo_gui/static/netgraph.js
```diff
@@ -625,3 +640,11 @@ Nengo.NetGraph.prototype.scaleMiniMapViewBox = function () {
     this.view.setAttribute('width', w / this.scale);
     this.view.setAttribute('height', h / this.scale);
 }
+
+Nengo.NetGraph.prototype.get_aspect_resize = function() {
```
tbekolay: Is the getter / setter necessary here? Not using them might be more clear.

tcstewar: Hmm..  we should make a general style decision here.  I could definitely see generally making setters/getters even when we're not doing anything, for the standard reasons of making it easier to make them complex in the future.

tbekolay: These are apparently a builtin feature of Javascript. See [MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Working_with_Objects#Defining_getters_and_setters). So, similar to Python, we should use raw attributes until we need something complex, at which point we can write this style of getter / setter without changing the code that's using those attributes.

tcstewar: Sweet!  I had no idea Janascript supports that... I totally agree.

tbekolay:
I rebased this and added a commit to fix up the `calc_axes_geometry` stuff (hopefully?) This can use some testing to make sure it's working post-rebase.

tbekolay:
One issue I noticed with this: resizing no longer moves graphs, only netgraph items. Seems to be due to removing the getters and setters; fixing now.

tbekolay:
Kay, fixed (we missed a call to `get_aspect_resize()`). If someone could test this branch and give a +1, I'll merge!

jgosmann:
Only gave it a short test, but seems to work.

tcstewar:
That seems to work for me, too.  It's still noticeably slower for complex displays for me, but that can be resolved in the future.

tcstewar:
That fixes the speed for me.

celiasmith:
cool, could someone merge this?

tbekolay:
Right, doing so now!

PR #327, opened 2015-06-05
only update minimap if displayed

studywolf:
realized that even when hidden the minimap was doing all the calculations etc, so set it such that the rescaling and resizing only happens when minimap is turned displayed.


tcstewar:
Looks good to me

Issue #332, opened 2015-06-05
Menus are being pushed out of position by minimap


cchan987:
By changing minimap div to relative, the menus no longer are pushed.

cchan987:
I think #339 would be the right approach since it maintains the cursor above the menu once the menu is opened

cchan987:
Seems there are 3 options:
#335 make minimap position relative through css
#337 set the menu position through js on each menu open
#339 Remove the surplus height subtraction from the menu transform

PR #335, opened 2015-06-05
made minimap position relative, doesn't push menu out of position anymore

cchan987:
Menus were previously being pushed out of position, changing minimap div to position:relative fixes this issue.

### DIFF
--- a/nengo_gui/static/main.css
+++ b/nengo_gui/static/main.css
@@ -36,7 +36,7 @@ div.minimap {
     width: 15%;
     left: 84.5%;
     top: 82.5%;
-    position: absolute;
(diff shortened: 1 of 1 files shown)


cchan987:
Fixes #332

jgosmann:
I don't think, that this is the correct fix. The minimap should be positioned `absolute`. As should be the menu so that they don't influence each other. Working on the, what I think, would be the correct fix. But I'm getting the menu now too low and it is not clear to me why.

jgosmann:
See #337

tcstewar:
Out of the three options, I went with #337

PR #337, opened 2015-06-05
Fix the position of the menu.

### LINKED ISSUE #332: Menus are being pushed out of position by minimap

jgosmann:
Fixes #332.

Better fix than #335, I think.

### DIFF
--- a/nengo_gui/static/menu.js
+++ b/nengo_gui/static/menu.js
@@ -30,6 +30,8 @@ Nengo.Menu.prototype.show = function (x, y, items) {
     // TODO: move this to the constructor
     this.menu_div = document.createElement('div');
     this.menu_div.style.position = 'fixed';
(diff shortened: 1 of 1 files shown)


tcstewar:
I like keeping it as ```absolute``` for positioning, and I prefer using standard ```style``` commands over the ```set_transform``` thing, so I think this solution is the best.

However, this raises the question of whether we want the menu to pop up such that the mouse cursor is at the top-left corner of the menu or overtop of the first item on the menu.  I strongly prefer the latter, but I'll put that in as a separate PR so there can at least be some discussion on it.

jgosmann:
Totally forgot about that the mouse cursor used to be over the first element (which explain the horizontal shift). I don't have a strong opinion. Usually context menus open with the corner at the mouse pointer, but whatever.

PR #339, opened 2015-06-05
remove height subtraction from transform to fix menu position

### LINKED ISSUE #332: Menus are being pushed out of position by minimap

cchan987:
Simple Fix for #332


jgosmann:
The  transform is is relative to the parent element, thus it behaves like `position: absolute`. The mouse position, however, is in browser coordinates (`layerX` and `layerY` are not available on the context menu event). To actually set the position of the element in browser coordinates `top` and `left` have to be set like in #335. This also gets rid of those magic constants adjusting for the top toolbar.

tcstewar:
Out of the three options, I went with #337

PR #342, opened 2015-06-05
Placed menu so cursor is on first item

tcstewar:
I usually like menus to appear positioned such that the mouse cursor is over the first item.  That's what nengo_gui had been doing up until yesterday, but the current master does not.  Instead, it positions the menu such that the top-left corner is under the mouse cursor.

This PR adds in that offset again.  However, I do want to get feedback on whether we want that offset.  One argument against the offset is that without the offset if you right-click accidentally, you can immediately left-click and the menu goes away (instead of selecting whatever the first item is). 

Thoughts?

### DIFF
--- a/nengo_gui/static/menu.js
+++ b/nengo_gui/static/menu.js
@@ -30,8 +30,8 @@ Nengo.Menu.prototype.show = function (x, y, items) {
     // TODO: move this to the constructor
     this.menu_div = document.createElement('div');
     this.menu_div.style.position = 'fixed';
-    this.menu_div.style.left = x;
-    this.menu_div.style.top = y;
+    this.menu_div.style.left = x - 20; // offset menu so that it appears
+    this.menu_div.style.top = y - 20;  //  with the cursor on the first item
     this.menu_div.style.zIndex = Nengo.next_zindex();
 
     this.menu = document.createElement('ul');


jgosmann:
+/- 0 whatever

jgosmann:
Related question: If you open the context menu on an element which at the bottom of the screen, part of the context menu can get hidden. Should we detect that case and shift it upwards? (Then you are not guaranteed to always have the same element under the cursor.)

tcstewar:
> If you open the context menu on an element which at the bottom of the screen, part of the context menu can get hidden. Should we detect that case and shift it upwards? (Then you are not guaranteed to always have the same element under the cursor.)

gah, good point.  Whereas we can always guarantee we're at a corner...  hmm...  that may convince ne to go with corners....

tbekolay:
The related question is now #373!

As for the discussion in this thread, my impression is that it's more important to avoid the accidental left-click than it is to have easy access to a quick right-click left-click combo. Especially since we have the ability to interact with double-clicks, I'm keen to keep the current top-left corner behavior, which seems to be the standard that window managers use.

cchan987:
For cursor on menu item:
It seems to me that the worst that can be done by a quick accidental right-left click would be an auto layout which would mess up the user's layout entirely. However we have undo built in to counter this problem easily. This means I wouldn't be worried about the accidental menu selection and having the cursor on top seems to save time in the general case.

For cursor on corner:
This does seem like the standard behaviour of browsers and keeping to the standards as much as possible has been the theme of this GUI. This allows the menu to be mirrored/inverted if you are opening it near the edge. For this reason, cursor on corner seems like the better option to me.

Pick: cursor on corner

tcstewar:
I'm switched over to the "cursor on corner" idea, mostly for the standardness and the allowing it to be shifted around if you're near an edge (#373)

tcstewar:
This issue got all sorted out with #373 so I'm closing this.

PR #343, opened 2015-06-06
Added config option to select backend to use

tcstewar:


tcstewar:
This PR also has some things that help a backend that needs to be explicitly close (like nengo_spinnaker) or backends that prefer being run for a specific number of steps at a time (like nengo_spinnaker)

PR #346, opened 2015-06-07
Allowed user to change the filename by editting it

tcstewar:
This was a handy feature in old nengo_gui

### DIFF
--- a/nengo_gui/components/netgraph.py
+++ b/nengo_gui/components/netgraph.py
@@ -262,6 +262,8 @@ class NetGraph(Component):
             del info['act']
             if action in ('auto_expand', 'auto_collapse'):
                 getattr(self, 'act_' + action[5:])(**info)
+            elif action == 'rename':
+                self.viz.viz.rename(info['filename'])
             elif action in ('pan', 'zoom', 'create_modal'):
                 # These should not use the undo stack
                 getattr(self, 'act_' + action)(**info)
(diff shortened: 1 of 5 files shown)


jgosmann:
* [ ] Should be disabled in IPython notebook.
* [ ] Can't type `/` because it opens the hotkey help window in Chrome (They don't provide `event.key` on keyboard events, so that `event.keyCode` has to be used. `event.keyCode` basically takes the printed character, which `event.key` would be, and converts it to a key code assuming the US layout where `?` and `/` are the same key. On my and many other keyboard layouts, these keys are different, but those symbols still produce the same keycode.)
* [ ] There is no indication of what happens, so I don't know whether the file was actually saved under the new name. I don't get any warning when overwriting a file. At first I even assumed it would open existing files.
* [ ] In Firefox there is no good indication that he input field has focus (in Chrome I get a standard OS X blue border around it).
* [ ] The cursor stays at `default`, but should change to the text selection bar.
* [ ] I'd assume pressing enter would accept the new name and remove the focus from the input, but it stays.
* [ ] When entering an invalid filename, the text should probably reverted to the original filename.

tcstewar:
Those are all extremely good points, and really should get addressed.  They won't happen for 0.1.1, though, and I think this feature is confusing enough if those aren't addressed that I don't think this PR should get considered for 0.1.1.

tcstewar:
This needs to be radically rethought, so I'm closing the PR.

PR #347, opened 2015-06-07
Refactor reload to fix reloading problems

tcstewar:
With the old system, we would occasionally get weird duplication bugs where things mostly don't work and there are two copies of the NetGraph (and other components).  These were accompanied by lots of "Error finding label" bugs.

This PR changes the load system so that instead of a two-part process where the internal Viz is updated and then the client does a reload, now the reloading is done in the actual swi page request itself (i.e. after the reload).  This means that we no longer have a period of time where the old page is connected to the server with the new content.  I believe the bug were due to information being passed to the client about the new page before the reload happens.

### DIFF
--- a/nengo_gui/components/sim_control.py
+++ b/nengo_gui/components/sim_control.py
@@ -28,7 +28,6 @@ class SimControl(Component):
         self.time = 0.0
         self.last_status = None
         self.next_ping_time = None
-        self.reload = False
         self.send_config_options = False
 
     def add_nengo_objects(self, viz):
@@ -84,9 +83,6 @@ class SimControl(Component):
         if status != self.last_status:
             client.write('status:%s' % status)
             self.last_status = status
-        if self.reload == True:
-            client.write('reload')
-            self.reload = False
         if self.send_config_options == True:
             client.write('sims:' + self.backend_options_html())
             client.write('config' +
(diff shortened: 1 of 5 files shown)


tcstewar:
If anyone who has been regularly getting that NetGraph duplication bug (or any "ERROR finding label" bugs) could test this PR, that'd be wonderful!  I haven't seen it happen for me since I've made the change, but it was pretty rare for me anyway.

jgosmann:
The reset button worked before in IPython and it would be nice to keep it that way.

tcstewar:
> The reset button worked before in IPython and it would be nice to keep it that way.

Hmm, it doesn't work for me in master at the moment....

In any case, I'll  get rid of that commit and we'll look at it separately -- I thought it was just something that got forgotten.

tcstewar:
> This would do a reset even when reset == False which might be confusing.

It's never False because it's a string, but it would be clearer as reset == "True".  Changed.  :)

jgosmann:
>> The reset button worked before in IPython and it would be nice to keep it that way.
> 
> Hmm, it doesn't work for me in master at the moment....

It does for me on bc519b0b5d

tcstewar:
> It does for me on bc519b0

I just tried that same commit, and it doesn't reload for me.  It just ends up hanging.   (both Chrome and Firefox).

jgosmann:
Reset shuts down the server for some reason.

jgosmann:
Turns out [we want this check](https://github.com/nengo/nengo_gui/pull/350/files#diff-6b004fd2ea83a7abcc0443f852b14d6fR155) to happen in either case.

tcstewar:
The reloading in IPython now works on this branch too, thanks to #350

PR #348, opened 2015-06-07
Created graphs now appear with a specific size

tcstewar:
They used to be sized based on the size of the component
that the graph is about.  Now they are generated at a fixed
200px by 200px size. (Sliders are at half that width).


celiasmith:
I had changed sliders to be dependent on the number of sliders being created (as it looks weird if you try to cram 8 sliders in to 100 px).  This seemed to work well (I've not checked to see exactly what this does, but just thought i'd mention it).

PR #350, opened 2015-06-07
Do not shutdown server too early in IPy notebook.

jgosmann:
See discussion in #347.


tcstewar:
That fixes it for me and looks good.  Merging...

Issue #355, opened 2015-06-08
Editting a read-only file breaks the editor


tcstewar:
If you try to save changes when you don't have permissions to edit that file, the editor breaks (it stops updating the NetGraph and showing syntax errors)

arvoelke:
This happens whenever you launch `nengo_gui` without write-access to the default example. Can be fixed by use of `sudo`. Proper fix proposed below:

The default behaviour of setuptools seems to be to preserve file modes. And when I pull the source from git, I get `-rw-r--r--`, which means only the user can write. Now if one were to sudo the pip install, this mode gets copied under the user `root`, which then of course is write-only unless `nengo_gui` is sudo'd as well.

Therefore, the best fix that I can see would be to simply apply:

`git update-index --add --chmod=a+w [example-files]`

To give all write-access to the example files and mark it that way in git. Then future releases should pip install with the same file mode, and thus have writeable examples without requiring sudo. A follow-up PR should then be made to give better behaviour in the case that a user attempts to open a script without write access.

tbekolay:
In my opinion, we shouldn't have the built-in examples modifiable. If we allow that, then it means that people's notions of what each example is can diverge. I would say the ideal situation would be that if a user goes to edit the built-in example, it should prompt them to save the modified file to a different location. Really, we shouldn't be writing `.cfg` files to these directories either.

arvoelke:
That also brings up the point that my proposed fix won't work, because the directory won't have write-access, and so the `.cfg` writing will break. Could also give the directory write access, but I agree that this is a little unconventional. I don't yet know what would give the better user experience though, because it would be nice if it just worked in an expected way with minimal interaction.

tbekolay:
FYI the `.cfg` writing issue just came up for a few people at the summer school. I've been recommending that people copy the examples to a local directory. If you run into anyone with this problem, you can send them [nengo_gui_examples.zip](https://dl.dropboxusercontent.com/u/6182929/nengo_gui_examples.zip) and have them extract it to a writeable directory.

tbekolay:
I've also added this to https://github.com/ctn-waterloo/summerschool2015/wiki/Resources.

tcstewar:
As a temporary solution, I'll also modify nengo_gui such that it doesn't crash if it tries to save the .cfg file.  That should just mean that everything works, but we won't get the layout file back when you close nengo_gui and restart it.  I'll put that fix in #356 as well.

tbekolay:
[...]

PR #356, opened 2015-06-08
Handle IOErrors when saving file

tcstewar:
This at least stops the editor from dying when you try to save a read-only file.  But there should also be some sort of visual feedback that it didn't save.


tcstewar:
(this is meant to address #355)

tcstewar:
This now also handles failure to save .cfg files

jgosmann:
LGTM :cake:

PR #364, opened 2015-06-09
Made left border on code editor obvious, cursor now changes on hover …

cchan987:
…over border

### DIFF
--- a/nengo_gui/static/ace.css
+++ b/nengo_gui/static/ace.css
@@ -3,5 +3,5 @@
     float:right;
     right: 0;
     z-index: 99998;
-    border-left:2px solid #606060;
+    border-left:5px solid #606060;
 }
--- a/nengo_gui/static/ace.css
+++ b/nengo_gui/static/ace.css
@@ -3,4 +3,5 @@
     float:right;
     right: 0;
     z-index: 99998;
+    border-left:2px solid #606060;
 }


tbekolay:
Great, this is a big improvement! For those who don't want to check this out, here's what it looks like:

screen shot 2015-06-09 at 3 25 30 pm

Hovering over the black line changes the cursor, as you'd expect, on both Chrome and Firefox.

One thing that's worth discussing, I think, is what color we want this to be? Right now it's black, which makes it more obvious where you have to hover to resize. On the other hand, I think it might be a bit nicer on the eyes to be the same grey color as the editor. If we do that, it might be good to make the border a bit thicker, even...

Thoughts?

cchan987:
> Right now it's black,

Actually a dark grey #606060.
Black is a little more harsh XD

tbekolay:
Hah, fair, I stand corrected ;)

cchan987:
> I think it might be a bit nicer on the eyes to be the same grey color as the editor. If we do that, it might be good to make the border a bit thicker, even...

I like the color as it stands as it shows nicely that the editor is distinct from other things. However I could totally go along with bumping up the border width from 2px to maybe 5px (do we want to use em?).

jgosmann:
> do we want to use em?

Usually not for borders, I would say. (Only for stuff like font-sizes, margins, paddings, and related stuff.)

jgosmann:
I think, the color is good, but I would make the border thinner. 1 or 2px, max 3px.

cchan987:
It seems like this should be a vote.
Is there a poll option for github?

jgosmann:
There's http://poll.gitrun.com/

tcstewar:
I'd stick with 5px, but that's because if we make it 2px that also reduces the size of the target for the resize, making it much harder to grab.  I'm somewhat surprised that interact.js does that, but it seems to on my machine (windows, chrome)

cchan987:
I agree with 5 px  on this one, I found I would overshoot/undershoot when going to grab the 2px border

tcstewar:
(the other option, of course, would be to have it visually look like 2px, but actually having a 5 or 10px grab area, if we can figure out how to do that).  But I'd say 5px is better for now.

tbekolay:
+1 5px

celiasmith:
+1

Trevor Bekolay wrote:

tcstewar:
Looks like 5px we're happy with..  if we want something visually thinner but with the same effective grab width, it should be a separate PR.  Merging!

PR #384, opened 2015-06-10
Add load file directory to sys.path

### LINKED ISSUE #379: Cannot import file from same folder as script
My script imports a helper file from the same folder. This doesn't work.

hunse:
This allows scripts in the directory with the file being loaded
to be imported. Fixes #379.

### DIFF
--- a/nengo_gui/viz.py
+++ b/nengo_gui/viz.py
@@ -231,9 +231,11 @@ class Viz(object):
                             line = nengo_gui.monkey.determine_line_number()
                             print('nengo_gui.Viz() started on line %d. '
                                   'Ignoring all subsequent lines.' % line)
-                    except:
+                    except Exception as e:
                         if not force:
                             raise
+                        else:
+                            print("Error loading file: %s" % e)
                     finally:
                         if filedir is not None:
                             sys.path.remove(filedir)
--- a/nengo_gui/viz.py
+++ b/nengo_gui/viz.py
@@ -1,9 +1,10 @@
 import importlib
+import json
 import os
+import socket
+import sys
 import time
 import threading
-import json
-import socket
 
 import nengo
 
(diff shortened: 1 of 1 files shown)


tbekolay:
Did a skim of the code and it LGTM. Will test at some point...

bjkomer:
Just tried it on some of my models, works well for me!

tcstewar:
Cool, I like that.  :)

Bizarrely enough, this means that you can now do things like:

```python
import nengo
model = nengo.Network():
with model:
    import my_example1
    import my_example2
    import my_example3
```

and those will show up in separate Networks in this model.  Not sure if that's a good idea or not, but kinda wacky.  :)

tbekolay:
Haha... just if anyone comes across this issue and is tempted to do the above, it's generally frowned upon to do a lot on import, because it's not clear when looking at that script what's supposed to happen. Instead, your scripts should define some functions or classes that you can call in your model to do the heavy lifting. For example:

```python
import nengo
import my_example1
import my_example2
import my_example3
model = nengo.Network():
with model:
    net1 = my_example1.network()
    net2 = my_example2.network()
    net3 = my_example3.network()
```
Now it's clear that you're creating three networks within your overall `model` network.

`network` in this case would be a function within each of the `my_example*.py` scripts. We wrote a [basic](http://pythonhosted.org/nengo/examples/network_design.html) and [advanced](http://pythonhosted.org/nengo/examples/network_design_advanced.html) network design example that gives some ideas for writing functions that create networks.

tbekolay:
Also, @tcstewar I notice that you don't delete PR branches after merging. Are you keeping them around for a particular purpose? I'm inclined to delete this branch, and others that have been merged, since at the moment we have [95 branches](https://github.com/nengo/nengo_gui/branches). But I'll hold off if you want to keep them around for some purpose!

tcstewar:
> Also, @tcstewar I notice that you don't delete PR branches after merging. Are you keeping them around for a particular purpose?

Oh, good question.  I'd generally been only deleting branches that I made, just in case other poeple wanted them.  But you're right there's no real reason to do that (if someone else made the branch, then they have a local copy if they need it after all....)  I'll try to go through and delete any branch that's been merged....  :)

PR #396, opened 2015-06-11
Handle exceptions when processing websocket msgs

tcstewar:
Missing this exception has caused the "editor stops updating anything" bug that we currently work around by reloading the page.

### DIFF
--- a/nengo_gui/monkey.py
+++ b/nengo_gui/monkey.py
@@ -45,9 +45,11 @@ def determine_line_number(filename='<string>'):
     pay attention to the line numbers in the main evaluated script (which
     is evaluated using exec(), so it doesn't have a normal filename).
     '''
-    if hasattr(sys, 'exc_traceback'):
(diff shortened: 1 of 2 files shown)


tcstewar:
Note: I still would like to know exactly what is causing that exception, so I'm printing the exception.

tcstewar:
This has now been pretty stable for me for a while, so it may be handy to have in the next release.

jgosmann:
I think, this is good to merge, if the too general except is addressed as it fixes an annoying bug and we can refactor the code later to put more code in the try-except (if this is the right thing to do).

tbekolay:
I added 0b54924. Does this address your concerns @jgosmann?

tcstewar:
> I added 0b54924. Does this address your concerns @jgosmann?

That looks pretty good to me...  Had to test when I haven't figured out how to replicate the bug, but it looks right.  :)

jgosmann:
LGTM :cake:

PR #397, opened 2015-06-11
Added hide-editor console line option.

xchoo:


tcstewar:
Looks good to me.

jgosmann:
`interactive` does some more stuff (mostly related to console output) in addition to disabling the editor.

I would name the option `--non-interactive` or something like that.

tcstewar:
Would it make sense to change the ```interactive``` parameter to something like ```embedded``` or ```jupyter``` or something like that?  ```interactive``` isn't all that clear to me, as both the Jupyter and standard nengo_gui both seem "interactive" to me.

jgosmann:
My reasoning was that it was not started from an interactive console, but `embedded` would also make sense.

xchoo:
Should the option to hide the code editor do something else then (instead of just setting the ```embedded``` or ```interactive``` flag to ```False```) when the Viz object is created?

I added this option in as a way to start the gui when you are using your own editor to edit code. It is true that you can just hide the editor, but you'll have to do that for every model, and every time you reload the page.

jgosmann:
If you just do not want to have the editor be displayed on startup, probably. In that case it migt be better to set this via the config file which should be implemented at some point. It might also be convenient if the GUI would just store the last state of showing/hiding the editor.

tbekolay:
#649 persists the state of the code editor, so with that PR if you hide it it will stay hidden the next time your open up the GUI. Does that address this issue, or do we still want a method to have it hidden on startup through a command-line argument?

tcstewar:
> #649 persists the state of the code editor, so with that PR if you hide it it will stay hidden the next time your open up the GUI.  Does that address this issue, or do we still want a method to have it hidden on startup through a command-line argument?

I think having the persistent state means we don't need a command line option.  But I'd also check with @xchoo and @jgosmann

jgosmann:
We still need a way to set the visibility in the code at startup, but no command line option.
Maybe we could do a general command line option that allows to override any of the persisted configuration options?

tcstewar:
Any thought son what we want to do with this Issue for version 0.2?  I'm fine to leave it as it is for 0.2 and think about adding a generic override-persistent-state command line option stuff in the future (as per @jgosmann 's suggestion)

jgosmann:
I would merge the server refactoring before messing around with this.

PR #404, opened 2015-06-12
HTMLView component for custom visualizations

tcstewar:
This lets a Node output arbitrary HTML to nengo_gui.

This lets you do this sort of thing:

```python
import nengo
model = nengo.Network()
with model:
    def my_node(t, x):
        my_node.html = '<h1>%1.3f %1.3f</h1>' % (x[0], x[1])
    node = nengo.Node(my_node, size_in=2)    
    values = nengo.Node([0,0])
    nengo.Connection(values, node)
```

The HTML can be anything you want, including SVG, making it possible to create visualizations of the 3-link arm, or of the Spaun brain, or of the task, anything else we want.

The basic idea is that if the output of a Node has a ```.html``` attribute, then you'll be able to make an HTMLView plot.  After the Node is run, the value of that ```html``` attribute is sent to the visualizer.


jgosmann:
Does it make sense to make the naming convention similar to the [IPython rich display system](https://ipython.org/ipython-doc/dev/config/integrating.html)? Something like `._repr_html_` (maybe not exactly that as it might interfere with IPython)?

Seanny123:
I also like the idea of marking the attribute with underscores. Would `._repr_gui_` be too verbose?

jgosmann:
It might be helpful to indicate the actual type of data. Right now we only have HTML, but maybe at some point we support (pure) SVG, PNG, ...

tcstewar:
Hmm...  I definitely like marking it with underscores.  I'm kinda tempted to use exactly ```._repr_html_``` since it's pretty much the same as what we're doing (I'd change the code so that if it finds a callable, it'll call it, so that it could work in both IPython and nengo_gui).  But that does seem a bit fragile.

What about ```._nengo_html_```?

jgosmann:
I like `._nengo_html_`, it's similar to `._ipython_display_`.

Seanny123:
+1 for `._nengo_html_`

tbekolay:
Also +1 for `._nengo_html_`

tcstewar:
Changed!

Seanny123:
Given the change, I think this is ready to be merged.

tcstewar:
> Speaking of which... Terry, what's the policy on merging pull requests to this repository?

The informal policy is that I've been doing it, but for PRs that I wrote I've been waiting for someone else to mark it as "needs merge".  That's mostly so that I'm the one responsible when the merges go horribly wrong.  :)

tcstewar:
> That's mostly so that I'm the one responsible when the merges go horribly wrong. :)

Speaking of which, this PR really needs to get rebased and tweaked to take into account the refactoring changes, even though github claims it can be merged.  I think I'll wait until #491 is merged before doing that, since it's pretty close.

Seanny123:
In regards to the scope of this pull request, is this also supposed to make it easier for people to make their own netgraph-plot-thingy for one-off experiments? Should I open an issue in regards to that?

If it is in scope, how is the user supposed to access the JavaScript/HTMLview side of things?

tcstewar:
> In regards to the scope of this pull request, is this also supposed to make it easier for people to make their own netgraph-plot-thingy for one-off experiments? Should I open an issue in regards to that?
> 
> If it is in scope, how is the user supposed to access the JavaScript/HTMLview side of things?

Good question...  I'd say that's outside the scope of this PR.  This is just for making raw dumb HTML output.  Things get much more complex if people what to make their own full plots.  So, for example, I think this would be suitable for making the drawing of an arm used in the Spaun videos, or for making the display that shows the Ravens Matrices values (also in the Spaun videos).  If people want more complex plotting abilities (like showing data over time), that involves much more javascript and is more like making your own Component.

tcstewar:
Okay, this is now all updated with the recent refactoring.  It had been previously marked as needs_merge, so I think it's all good to go, but I'll wait a day just to be sure.  :)

Seanny123:
Yep. It's still good to go. Merge away!

tbekolay:
I'd be more comfortable with this being merged in if there were at least one example of it being used somewhere...

tcstewar:
> I'd be more comfortable with this being merged in if there were at least one example of it being used somewhere..

Ooo...  good idea.  Okay, I've added an example as ```examples/basics/html.py```.  It shows three different usages, including a nice minimal 2-joint arm visualization that turned out to be easier to implement than I thought it would be.  :)

Let me know what you think...

PR #409, opened 2015-06-14
Switched sys.exc_traceback to sys.exc_info

tcstewar:
As @jgosmann pointed out to me:

>  exc_traceback is deprecated since version 1.5(!).

So this switches to the more modern (and thread-safe) approach.


jgosmann:
LGTM :cake:

PR #410, opened 2015-06-14
Include ace searchbox file

### LINKED ISSUE #408: Missing some Ace components
My emacs-fingers pressed Ctrl+f in the code editor, which of course doesn't do the emacs thing. But it's apparently supposed to bring up a search box; instead I get the following in my console:

```python
  File "/home/tbekolay/Code/nengo_gui/nengo_gui/swi.py", line 276, in handle_request
    text = getattr(self, command)(*args[1:], **db)
  File "/home/tbekolay/Code/nengo_gui/nengo_gui/server.py", line 55, in swi_static
    data = pkgutil.get_data('nengo_gui', fn)
  File "/usr/lib/python2.7/pkgutil.py", line 591, in get_data
    return loader.get_data(resource_name)
  File "/usr/lib/python2.7/pkgutil.py", line 255, in get_data
    return open(pathname, "rb").read()
IOError: [Errno 2] No such file or directory: '/home/tbekolay/Code/nengo_gui/nengo_gui/static/lib/js/ace-src-min/ext-searchbox.js'
```

We should either 1) include all the extensions that we might want or 2) disable these hotkeys.

tcstewar:
Fixes #408

I'm not sure why this extension is activated by default. The other extensions https://github.com/ajaxorg/ace-builds/tree/master/src look like you have to manually activate them, so hopefully there aren't any other examples like this.

Still, this now means we have a full search-and-replace feature available on Ctrl-F....  :)


Seanny123:
LGTM.

PR #415, opened 2015-06-15
Reset slider positions on simulation reset

### LINKED ISSUE #365: Sliders do not reset when the model resets
If I run a model, move some sliders, and then change the code and hit play again, the sliders stay at their old value.

tcstewar:
This handles slider positions when the simulation resets (so that the sliders don't end up staying wherever they got left in the previous simulation).  Handles both constant sliders and callable sliders.

Fixes #365


tbekolay:
Tried this out; works for me! If anyone's curious, this makes it such that the slider moves back to what it was originally set to; it doesn't override the initial value with what the slider currently shows. Which is far preferable, IMO, because otherwise you'd have to shut down the whole server if you had a callable output and wanted it back.

Merging to get this in the 0.1.4 release.

Issue #406, opened 2015-06-12
Value Plot: Changing a population from Ensemble to EnsembleArray crashes the server


xchoo:
If you have a value plot stored in the config file, and you change a population from and Ensemble to an EnsembleArray, it will crash the server. It happens because the value plot is looking for the ```size_out``` attribute of the ensemble to determine how many dimensions it should plot. 

However, an EnsembleArray is a network which does't have this property, and so the server crashes.

This problem will also occur if you have a value plot up for an ensemble, and convert that ensemble to a network (value plots are not valid for networks).

Adding logic in to check the population type when loading up a config file, and disabling plots that are not longer valid should fix this issue.

bjkomer:
I think this is related to #398

tbekolay:
Fixed in #416.

PR #416, opened 2015-06-15
Handle class changes when code changes

### LINKED ISSUE #398: Components changing type is not reloaded properly in the editor
For example, if you make a node by first typing `a = nengo.Node()` and then later change it to `a = nengo.Node([1])` it will still visually display it as a black dot (passthrough node) when it should be a grey box. 

Although less common the same bug happens if you switch it to something like `a = nengo.Network()`. This can lead to other weird bugs where the gui gives you the option to 'expand/collapse' a regular node, and then throws some errors when you try to do that.

tcstewar:
Fixes #398 and #406

It was supposed to be detecting an object changing its nengo class, but that update was sent to the javascript *after* the new objects were created, confusing things.  This change fixes that and also detects Nodes changing to passthrough Nodes (and vice-versa).

### DIFF
--- a/nengo_gui/components/netgraph.py
+++ b/nengo_gui/components/netgraph.py
@@ -211,7 +211,10 @@ class NetGraph(Component):
                               (nengo_gui.components.SimControlTemplate,
                                nengo_gui.components.NetGraphTemplate,
                                nengo_gui.components.AceEditorTemplate)):
-                self.viz.add_template(template)
+                try:
+                    self.viz.add_template(template)
+                except:
+                    print('failed to recreate plot for %s' % template)
 
         self.viz.changed = True
 
--- a/nengo_gui/components/netgraph.py
+++ b/nengo_gui/components/netgraph.py
@@ -108,12 +108,22 @@ class NetGraph(Component):
             if new_uid != uid:
                 new_item = None
 
-            if new_item is None or not isinstance(new_item, old_item.__class__):
-                self.to_be_sent.append(dict(
-                    type='remove', uid=uid))
-                del self.uids[uid]
-                removed_uids[old_item] = uid
+            # find reasons to delete the object.  Any deleted object will
+            # be recreated, so try to keep this to a minimum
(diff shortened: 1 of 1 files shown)


tbekolay:
Tried this out! This fixes #398 for me, but not #406. If I have an ensemble and change it to an ensemble array, it does indeed update the netgraph properly! But, then when I press the play button I still get this in the console:
```bash
  File "/home/tbekolay/Code/nengo_gui/nengo_gui/server.py", line 158, in ws_viz_component
    component.update_client(client)
  File "/home/tbekolay/Code/nengo_gui/nengo_gui/components/netgraph.py", line 243, in update_client
    self.check_for_reload()
  File "/home/tbekolay/Code/nengo_gui/nengo_gui/components/netgraph.py", line 50, in check_for_reload
    self.reload(code=new_code)
  File "/home/tbekolay/Code/nengo_gui/nengo_gui/components/netgraph.py", line 54, in reload
    self._reload(code=code)
  File "/home/tbekolay/Code/nengo_gui/nengo_gui/components/netgraph.py", line 214, in _reload
    self.viz.add_template(template)
  File "/home/tbekolay/Code/nengo_gui/nengo_gui/viz.py", line 74, in add_template
    c = template.create(self)
  File "/home/tbekolay/Code/nengo_gui/nengo_gui/components/component.py", line 37, in create
    *self.args, **self.kwargs)
  File "/home/tbekolay/Code/nengo_gui/nengo_gui/components/value.py", line 16, in __init__
    self.n_lines = int(obj.size_out)
AttributeError: 'EnsembleArray' object has no attribute 'size_out'
```
I did have a value graph open, but it was a value graph of the ensemble inside the ensemblearray (as confirmed by the label). So it seems like the value graph I had open previously is still around on the Python side even though I can't see it on the JS side?

tcstewar:
I think I've got that problem handled now...  :)

tbekolay:
Cool, works for me! Merging!

Issue #394, opened 2015-06-11
Convert spike charts to a single SVG path


tcstewar:
As pointed out by Roland, "you are drawing each line with its own SVG Line element.  It turns out that each SVG element causes (at least) 1 call from the CPU to the GPU, and this quickly becomes the perf bottleneck for many apps.  If you change the spike chart code to use a single SVG Path element to draw all of the lines in the chart (moveTo, LineTo, …), you will find a dramatic speedup when it is drawn.  This will give you a lot of the speedup you would experience by converting the chart to use WebGL.  In the future, if you want to be able to control the color, size, line type, etc. of each line, it would be a good time to check out WebGL."

studywolf:
awesome tip!

tbekolay:
Super awesome! It would be great to have speedy rasters :rainbow:

tcstewar:
Done in #417

PR #417, opened 2015-06-15
use a single SVG path for rasters

tcstewar:
This is *much* faster.  Thank you to Roland Fernandez for the suggestion!

Addresses #394


tbekolay:
Faster, and shorter code! Double win!

tbekolay:
Opened 5 rasters and it was smooth as butter. Merging!

celiasmith:
Now we just need to control the number of neurons :smiley:

PR #424, opened 2015-06-15
Added console below text editor

tcstewar:
This console shows both the stack trace (if there's an error) and anything printed to stdout.  This means you can do things like:

```python
print 'hello world'
help(model)
```

The console is resizable by dragging

### DIFF
--- a/nengo_gui/monkey.py
+++ b/nengo_gui/monkey.py
@@ -3,6 +3,10 @@ import importlib
 import threading
 import traceback
 import sys
+try:
+    from StringIO import StringIO
+except ImportError:
+    from io import StringIO
 
 
 # list of Simulators to check for
@@ -67,12 +71,20 @@ def determine_line_number(filename='<string>'):
         return line
     return None
 
+stdout = StringIO()
 
 @contextlib.contextmanager
 def patch():
+    global stdout
+
     flag.executing = True
     del found_modules[:]
     simulators = {}
+
+    # add hooks to record stdout
+
+    sys.stdout = StringIO()
+
     for name in known_modules:
         try:
             mod = importlib.import_module(name)
@@ -85,3 +97,6 @@ def patch():
     for mod, cls in simulators.items():
         mod.Simulator = cls
     flag.executing = False
+
+    stdout = sys.stdout
+    sys.stdout = sys.__stdout__
--- a/nengo_gui/monkey.py
+++ b/nengo_gui/monkey.py
@@ -3,10 +3,7 @@ import importlib
 import threading
 import traceback
 import sys
-try:
-    from StringIO import StringIO
-except ImportError:
-    from io import StringIO
+from nengo.utils.compat import StringIO
 
 
 # list of Simulators to check for
@@ -71,32 +68,33 @@ def determine_line_number(filename='<string>'):
         return line
     return None
 
-stdout = StringIO()
 
-@contextlib.contextmanager
-def patch():
-    global stdout
+class Patch(object):
+    def __enter__(self):
+        self.stdout = StringIO()
 
-    flag.executing = True
-    del found_modules[:]
-    simulators = {}
+        flag.executing = True
+        del found_modules[:]
+        self.simulators = {}
 
-    # add hooks to record stdout
+        # add hooks to record stdout
 
-    sys.stdout = StringIO()
+        sys.stdout = self.stdout
 
-    for name in known_modules:
-        try:
-            mod = importlib.import_module(name)
-        except:
-            continue
-        found_modules.append(name)
-        simulators[mod] = mod.Simulator
-        mod.Simulator = make_dummy(mod.Simulator)
-    yield
-    for mod, cls in simulators.items():
-        mod.Simulator = cls
-    flag.executing = False
+        for name in known_modules:
+            try:
+                mod = importlib.import_module(name)
+            except:
+                continue
+            found_modules.append(name)
+            self.simulators[mod] = mod.Simulator
+            mod.Simulator = make_dummy(mod.Simulator)
 
-    stdout = sys.stdout
-    sys.stdout = sys.__stdout__
+    def __exit__(self, exc_type, exc_value, traceback):
+        for mod, cls in self.simulators.items():
+            mod.Simulator = cls
+        flag.executing = False
+
+        sys.stdout = sys.__stdout__
+
+patch = Patch()
--- a/nengo_gui/components/ace_editor.py
+++ b/nengo_gui/components/ace_editor.py
@@ -1,6 +1,9 @@
-import nengo
 import json
+
+import nengo
+
 from nengo_gui.components.component import Component, Template
+import nengo_gui.monkey
 
 class AceEditor(Component):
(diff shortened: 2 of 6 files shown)


nengo_gui/monkey.py
```diff
@@ -3,6 +3,10 @@
 import threading
 import traceback
 import sys
+try:
+    from StringIO import StringIO
+except ImportError:
+    from io import StringIO
```
tbekolay: Alternatively, can do `from nengo.utils.compat import StringIO`.

nengo_gui/monkey.py
```diff
@@ -67,12 +71,20 @@ def determine_line_number(filename='<string>'):
         return line
     return None
 
+stdout = StringIO()
 
 @contextlib.contextmanager
 def patch():
+    global stdout
```
tbekolay: Instead of using `global`, perhaps consider implementing this as a context manager class? `contextlib.contextmanager` is cool, but needing to store something is probably the right time to switch over just implementing `__enter__` and `__exit__` methods.

tbekolay:
This is how it looks, for those intrigued but not intrigued enough to check out this branch:

screenshot from 2015-06-15 16 11 53

I made a few code comments. I also have some aesthetic comments!

The grey background / dark text looks weird IMO. I think ideally, we would have white background, black text for print statements and white background, red text for printing exceptions. Implementation wise, we can implement this by printing tracebacks to `stderr`, and showing `stderr` in red.

It looks like there's a rounded radius on this box? If you look at the screenshot above, the top left corner of the console is rounded, which doesn't square up and looks weird. It's also rounded at the bottom left, which looks weird, and there's a bunch of whitespace at the very bottom, between the console and the sim control. Should just a bunch of CSS changes for this.

I also found a z-index bug when I had the window resized...

screenshot from 2015-06-15 16 12 03

Definitely a great thing to have though! Will help a ton for debugging models!

tcstewar:
> The grey background / dark text looks weird IMO. I think ideally, we would have white background, black text for print statements and white background, red text for printing exceptions.

Sounds good!  made that change....  I also made the error stuff in a bold font as I think that's helpful.  But someone who can see red should check my colours

> It looks like there's a rounded radius on this box?

Removed

> I also found a z-index bug when I had the window resized...

Fixed!

nengo_gui/monkey.py
tcstewar: Good idea...  that's now switched over.

tbekolay:
Looking pretty sweet! Here it is now:

screenshot from 2015-06-15 18 18 10

Also for reference, here it is without bold:

screenshot from 2015-06-15 18 20 39

I think I prefer the non-bold version, but I'm not too fussed either way.

Also I added a super important commit to make sure that the console font-family matches the ace editor font family. When I tried this this time, the Bootstrap default style for `pre` had kicked in and it was Courier New, which I am not cool with! But now it matches the ace editor, which on my machine makes it Ubuntu Mono! Tight.

tbekolay:
Oh! I did encounter a weird bug though. Steps:

1. Load up `nengo_gui`
1. Change code so that an error occurs
1. Mouseover the top-border of the console

I expect that border to change the pointer to a resize handle, but it stays as default. If I go to the far right portion of the border, though, I can get a resize handle. And then, if I resize it to be high enough, the correct mouseover behavior kicks in, somehow... even shrinking down the console again keeps the right resize handlers, so yeah, I dunno what's up with that but it's weird.

tcstewar:
> I think I prefer the non-bold version, but I'm not too fussed either way.

I find it much nicer to be bold on whatever font it has picked for me...  and I definitely agree that it's way better than Courier New.  ;)

tcstewar:
> Oh! I did encounter a weird bug though.

Weird.  I get the same thing...  I think this is similar to the resize issues we've generally had with interact.js....  It does seem to get confused by borders.  :(

bjkomer:
Looks very nice!
The only issue I've seen is that if you shrink it all the way to the bottom so that it disappears, it is impossible to bring it back up (it seems to be able to go slightly under the bottom bar). On a related note, I think there should be a way to show/hide the console.
I slightly prefer the non-bold version of the font.

tcstewar:
> The only issue I've seen is that if you shrink it all the way to the bottom so that it disappears, it is impossible to bring it back up

Fixed...  I just gave it a minimum height.  In the future, it'd be nice to add a "close" icon or something like that, but I think that's a job for a future PR.

jgosmann:
-1 bold font

Can we get some padding in there? It looks weird that the text is touching the border.

When changing the height of the console, but without much text in it, the scroll bar will be in the middle of the console which is weird.

tcstewar:
> -1 bold font

If no one likes the bold font, does any one have any suggestions for what to do about the 10% of users who are colour-blind?  I cannot see the difference between the red and the black font.  That's why I had it be bold.

tcstewar:
> Can we get some padding in there? It looks weird that the text is touching the border.

Left padding added  :)

> When changing the height of the console, but without much text in it, the scroll bar will be in the middle of the console which is weird.

I'm not sure what you mean here.  I don't get a scroll bat in the middle of the console.  Could you post a screen shot?

jgosmann:
scroll

Also, the text is wrapped in Chrome, but not Firefox as shown here. That should probably made to be consistent.

jgosmann:
Also there is sometimes like a 1px white line just above the gray border going through the vertical border in Firefox. This might just be a rendering bug of Firefox ... especially as it seems to me to be a true retina display pixel (a CSS 1px line should be thicker, I believe).

tcstewar:
> Also, the text is wrapped in Chrome, but not Firefox as shown here.

Ah, that's the actual problem here.  Fixed now, and this makes the horizontal scroll thing go away

tcstewar:
> I cannot see the difference between the red and the black font.

Here's a slightly different attempt at the colours.  I've made the normal text blue and the error text red, and I'm using the same blue and red as in the graphs (for consistency).  I've also softened the bold a bit.  How do people feel about this?

tbekolay:
I like the softened bold! Though, I'm on the Mac now, seeing it on Linux will be a different ballgame.

This version of red looks very orange to me. Is that just me? Speaking to those with normal color vision here...

screen shot 2015-06-16 at 11 15 49 am

celiasmith:
Yeah, it does look kinda orange, but it is also in the colour palette that is colour-blind optimized that we use on the graphs.

jgosmann:
It's not real read, but I like the look. :)

tbekolay:
Can confirm, the soft bold also looks great on Linux!

Fair point about using the colour-blink optimized palette... I tried a few alternatives (using our "magenta", inverting it so that on error we get light text on dark background) but none of them look as good as our "red", so this works for me.

tbekolay:
Made #430 for the color stuff. I also added a commit to remove the background from `#console pre` as when I was playing around with the `#console` background, it was easier not to have to change it in both places.

Are there any other lingering issues on this, or are we good to merge?

bjkomer:
LGTM!

tbekolay:
Me too; @tcstewar confirms this is good to go. Merging!

PR #426, opened 2015-06-15
Reset simulation and clear graphs

celiasmith:
This adds a button that resets the simulation.  If the simulation is running it immediately resets. If it's paused, it clears all the graphs and resets the sliders.  It uses the 'sim_reset' event that is used for sliders (although i'm not sure that's working correctly; but that's someone else's code :sweat_smile: ). I've defined reset functions for all the components, for future component design...

### DIFF
--- a/nengo_gui/static/sim_control.js
+++ b/nengo_gui/static/sim_control.js
@@ -174,10 +174,9 @@ Nengo.SimControl.prototype.on_pause_click = function(event) {
     }
 };
 
-Nengo.SimControl.prototype.reset = function(event) {
-    //this.time = 0;
-    //this.rate = 0;
-    //this.set_status('pause')
+Nengo.SimControl.prototype.reset = function() {
+    this.time = 0.0;
+    this.rate = 0.0;
     this.ws.send('reset');
     self.sim.div.dispatchEvent(new Event('sim_reset'));
 };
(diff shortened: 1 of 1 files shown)


drasmuss:
Would be nice to have a hotkey as well (ctrl-r?)

tcstewar:
Hmm, it seems to work, but if you reset the simulation and then press play again, it doesn't show the spinning gear while it rebuilds.

tbekolay:
The simcontrol doesn't get updated on reset (it stays at whatever time you were at when you hit reset).

tbekolay:
Me and Chris were checking this out and ran into some issues with slider resetting, which I attempted to fix with efa1735b8f3470.

There's still two very bizarre issues though, both of which deal with the very beginning of the simulation.

Issue 1: Auto rebuilding keeps around some weird state?

How to reproduce:

1. Open `default.py` (open when you open `nengo_gui`).
1. Make a slider for `stim` and a value plot for `a`.
1. Set the slider to 1.0.
1. Play it for a little bit.
1. Add a spike plot for `a`. This forces a rebuild when you press play.
1. Press play. Pause soon after so you can go back to time 0!
1. The value initially jumps to 1.0, then back down to the initial value of 0.0.

Screenshot:

screenshot from 2015-06-15 17 56 17

Note that I didn't touch the slider once I hit play again! The cog shows up, so it did rebuild. But there's some information left over about what the slider used to have.

Issue 2: Reset button has weird startup transient

Issue 1 doesn't occur when using the snazzy new reset button. But instead there's this:

1. Open `default.py`
1. Make a slider for `stim` and a value plot for `a`
1. Set the slider to -1.0
1. Play it for a bit.
1. Press the reset button
1. Press play. Pause soon after so you can go back to time 0!
1. There's a weird startup transient, but the value doesn't jump like in issue 1.

Screenshot:

screenshot from 2015-06-15 18 07 25

tcstewar:
For people working on this PR, you might want to ```git cherry-pick 130ca3c``` to get the fix I just tossed in here https://github.com/nengo/nengo_gui/pull/428

celiasmith:
I think this is ready to merge... :pray:

PR #428, opened 2015-06-16
Makes sliders reset on simulation restart

tcstewar:
The slider resetting system wasn't working in the
case where the simulation was being reset due to new graphs
being created.  I believe this is at least part of the problems
noted in https://github.com/nengo/nengo_gui/pull/426#issuecomment-112223850

To replicate, run a model with a graph and a slider.  Move the slider around a bit and leave it at a non-zero value.  Now make a new graph.  Hit play again.  Without this fix the Node will continue to output the old value, even though the slider moves back to its original value.  With this fix the Node's output value will be reset.

### DIFF
--- a/nengo_gui/components/slider.py
+++ b/nengo_gui/components/slider.py
@@ -10,6 +10,7 @@ class Slider(Component):
         self.node = node
         self.base_output = node.output
         self.override = [None] * node.size_out
+        self.last_time = None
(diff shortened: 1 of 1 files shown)


bjkomer:
Works for me!

tbekolay:
Works for me too! Merging!

PR #429, opened 2015-06-16
Fix resize.

### LINKED ISSUE #422: Ensembles become larger on first resize movment
When resizing an ensemble by dragging an edge, after the first mouse movement the ensemble pops to being a larger size.

jgosmann:
Fixes #422.

### DIFF
--- a/nengo_gui/static/netgraph_item.js
+++ b/nengo_gui/static/netgraph_item.js
@@ -186,8 +186,9 @@ Nengo.NetGraphItem = function(ng, info, minimap, mini_item) {
                         var vertical_resize = event.edges.bottom || event.edges.top;
                         var horizontal_resize = event.edges.left || event.edges.right;
 
-                        var w = pos[0] - event.clientX;
-                        var h = pos[1] - event.clientY;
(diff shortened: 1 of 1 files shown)


tcstewar:
Looks good to me...  :)

tbekolay:
This fixes the issue that I was having for the top and bottom edge! But the resize jump still happens for me on the diagonals... It might be better than before though! Here's [a video](https://www.youtube.com/watch?v=fQL7QwSvk1Y) of what happens on my machine. Note that I hacked the CSS to make the transparent resize rectangle to have a stroke to make it more obvious what's happening. Basically, it jumps in size on the diagonals for me; the edge handlers work pretty well (if they're jumping, it's less noticeable). But on the diagonals, I'm clicking and trying to drag it smaller but instead it grows.

jgosmann:
I would propose to merge this anyways before the next release as it improves things.
The remaining problem might be the the reported size of the object is slightly off, but I have to check in detail and might take a while.

tbekolay:
Sounds good to me!

PR #436, opened 2015-06-17
Fix window bug with repeating aborted connections

tcstewar:
Occasionally when shutting things down on Windows, I get:

```
  File "c:\users\terry\documents\github\nengo_gui\nengo_gui\server.py", line 125, in ws_viz_component
    msg = client.read()
  File "c:\users\terry\documents\github\nengo_gui\nengo_gui\swi.py", line 594, in read
    data = data + bytearray(self.socket.recv(512))
error: [Errno 10053] An established connection was aborted by the software in your host machine
  File "c:\users\terry\documents\github\nengo_gui\nengo_gui\server.py", line 125, in ws_viz_component
    msg = client.read()
  File "c:\users\terry\documents\github\nengo_gui\nengo_gui\swi.py", line 594, in read
    data = data + bytearray(self.socket.recv(512))
error: [Errno 10053] An established connection was aborted by the software in your host machine
  File "c:\users\terry\documents\github\nengo_gui\nengo_gui\server.py", line 125, in ws_viz_component
    msg = client.read()
  File "c:\users\terry\documents\github\nengo_gui\nengo_gui\swi.py", line 594, in read
    data = data + bytearray(self.socket.recv(512))
error: [Errno 10053] An established connection was aborted by the software in your host machine
```
repeated forever and ever.  This PR should let the system notice this failure.

### DIFF
--- a/nengo_gui/swi.py
+++ b/nengo_gui/swi.py
@@ -601,6 +601,8 @@ class ClientSocket(object):
                 pass
             elif e.errno == 9:  # "Bad file descriptor" means socket closed
                 raise SocketClosedError("Cannot read from closed socket.")
+            elif e.errno == 10053:  # aborted connection
+                raise SocketClosedError("Cannot read from closed socket.")
             else:
                 raise
         except socket.timeout:


jgosmann:
Can you add a comment with the appropriate constant from the `errno` module? We can't actually use it there with some more refactoring because it will only be defined on windows, but might make it a bit easier to do later (especially for someone without a windows system). And while you're at it, can you fill in the other windows error constants to?

`errno.errorcode` provides the mapping from error code to constant.

jgosmann:
Otherwise this looks good to me.

tcstewar:
Replaced by #648 which also does the errorcode stuff @jgosmann requested above.

PR #445, opened 2015-06-18
Fix resize

### LINKED ISSUE #422: Ensembles become larger on first resize movment

jgosmann:
Fixes #422.

Makes some changes to the layout (using flex layout for editor and console) to have the browser handle the size calculations. Hope this doesn't break anything else.

### DIFF
--- a/nengo_gui/static/ace.css
+++ b/nengo_gui/static/ace.css
@@ -1,22 +1,27 @@
-#editor {
-    position: fixed;
-    float:right;
-    right: 0;
-    z-index: 99998;
+#rightpane {
+    -webkit-flex: 0 0 auto;
+    flex: 0 0 auto;
     border-left:5px solid #606060;
+
+    display: -webkit-flex;
+    display: flex;
+    -webkit-flex-direction: column;
+    flex-direction: column;
+}
+
+#editor {
+    height: 100%;
+    flex: 1 1 auto;
 }
 
 #console {
-    position: fixed;
-    float:right;
-    right: 0;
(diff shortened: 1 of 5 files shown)


tcstewar:
I just finally got around to testing this, and it looks good to me, except for one interface weirdness:  This gets rid of the ability to do a corner-resize on the error console (so I can resize both the height and the width of the error console at the same time).  I've found myself using the corner-resize a fair bit....  I'm not quite sure how to get that functionality back in with these changes, though...  thoughts?

jgosmann:
Corner resize works on OS X, but not Linux. :confounded:

jgosmann:
Nevermind, I am too stupid to checkout the correct branch. :unamused:

jgosmann:
Should be fixed now.

tcstewar:
Cool, that fixes the corner resize for me....  :)

However, I just noticed another bug in the flex layout.  It seems to work fin in Firefox (36.0.4) but in Chromium (37.0.2062.94) something seems wrong with the vertical height of the page.  I'm getting a scroll bar on the right-hand side and the placement of the minimap is wrong

image

jgosmann:
Also fixed.

tcstewar:
That's looking good to me now!  :)  thanks!

tcstewar:
Argh, found another bug, after I merged into master.  All the graphs are now on top of the editor.  I let @jgosmann know about it and he's taking a look now.

Issue #450, opened 2015-06-22
Y axis labels get cut off if they're too long


celiasmith:
If you sent the scale on a value plot to be 0 to 1000 (or -1,-.1), the margin cuts off some of the text.  Here's a purported solution to this kind of problem that @tcstewar found:

http://stackoverflow.com/questions/21603800/how-to-fit-variable-length-tick-labels-on-a-d3-line-chart

cchan987:
closed by #502

PR #451, opened 2015-06-22
Fixes #441

### LINKED ISSUE #441: Not able to set value graph range to have a negative maximum value
If I try to set the maximum of a graph to a negative value e.g. range -1 to -0.5 it is not possible as the gui thinks it is not in the correct format

e.g. typing "-1.0,-0.5" into the range setting box

celiasmith:
This fixes the problem with modals in #441, but it also highlighted a problem with margins on graphs, which I've posted as another problem in #450, but that shouldn't block these changes.


Seanny123:
LGTM

PR #454, opened 2015-06-30
Show slider range

cchan987:
Showing the max and min slider values adjacent to the slider

### DIFF
--- a/nengo_gui/static/slider.js
+++ b/nengo_gui/static/slider.js
@@ -28,7 +28,7 @@ Nengo.Slider = function(parent, sim, args) {
     this.group = document.createElement('div');
     this.group.style.height = this.slider_height;
     this.group.style.marginTop = this.ax_top;
-    this.group.style.marginLeft = '2em';
+    //this.group.style.marginLeft = '2em';
     this.group.style.whiteSpace = 'nowrap';
     this.group.position = 'relative';
     this.div.appendChild(this.group);
--- a/nengo_gui/static/slider.js
+++ b/nengo_gui/static/slider.js
@@ -28,6 +28,7 @@ Nengo.Slider = function(parent, sim, args) {
     this.group = document.createElement('div');
     this.group.style.height = this.slider_height;
     this.group.style.marginTop = this.ax_top;
+    this.group.style.marginLeft = '2em';
     this.group.style.whiteSpace = 'nowrap';
     this.group.position = 'relative';
     this.div.appendChild(this.group);
@@ -62,6 +63,44 @@ Nengo.Slider = function(parent, sim, args) {
         this.sliders.push(slider);
     }
 
+    this.bound_labels = document.createElement('div');
+    this.bound_labels.style.height = this.slider_height;
+    this.bound_labels.style.width = '2em';
+    this.bound_labels.style.marginTop = this.ax_top;
+    this.bound_labels.style.position = 'absolute';
+    this.bound_labels.style.top = '0px';
+    var range = this.sliders[0].scale.domain();
+
+    this.max_label = document.createElement('p');
+    this.min_label = document.createElement('p');
+
+    this.max_label.innerHTML = range[0];
+    this.min_label.innerHTML = range[1];
+
+    this.min_label.style.left = '1em'
+    this.max_label.style.left = '1em';
+
+    this.max_label.style.position = 'absolute';
+    this.min_label.style.position = 'absolute';
+
+    this.max_label.style.zIndex = -1;
+    this.min_label.style.zIndex = -1;
+
+    this.min_label.style.bottom = '0px';
+
+    this.bound_labels.appendChild(this.max_label);
+    this.bound_labels.appendChild(this.min_label);
+
+    this.div.appendChild(this.bound_labels);
+    this.hide_bound_labels();
+
+    $(this.div).mouseenter(function(){
+        self.show_bound_labels();
+    });
+    $(this.div).mouseleave(function(){
+        self.hide_bound_labels();
+    });
+
     /** call schedule_update whenever the time is adjusted in the SimControl */
     this.sim.div.addEventListener('adjust_time',
             function(e) {self.schedule_update();}, false);
@@ -73,6 +112,17 @@ Nengo.Slider = function(parent, sim, args) {
 };
 Nengo.Slider.prototype = Object.create(Nengo.Component.prototype);
 Nengo.Slider.prototype.constructor = Nengo.Slider;
+
+Nengo.Slider.prototype.show_bound_labels = function () {
+    $(this.min_label).css('display', 'block');
+    $(this.max_label).css('display', 'block');
+}
+
+Nengo.Slider.prototype.hide_bound_labels = function () {
+    $(this.min_label).fadeOut('fast');
+    $(this.max_label).fadeOut('fast');
+}
+
 
 Nengo.Slider.prototype.set_axes_geometry = function(width, height) {
     this.width = width;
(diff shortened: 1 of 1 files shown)


nengo_gui/static/slider.js
```diff
@@ -297,7 +349,10 @@ Nengo.Slider.prototype.set_range = function() {
             for (var i in self.sliders) {
                 self.sliders[i].set_range(min, max);
             }
+            console.log('here')
```
jgosmann: Remove debug output?

nengo_gui/static/slider.js
```diff
@@ -28,6 +28,7 @@ Nengo.Slider = function(parent, sim, args) {
     this.group = document.createElement('div');
     this.group.style.height = this.slider_height;
     this.group.style.marginTop = this.ax_top;
+    //this.group.style.marginLeft = '2em';
```
jgosmann: Remove commented out code?

jgosmann:
* The labels can randomly pop up and fade out again during resizing.
* It might look better to have the labels closer to the vertical bar.
* I like the fade out animation, but it is strange the the hover box instantly disappears.

nengo_gui/static/slider.js
```diff
@@ -145,6 +192,8 @@ Nengo.Slider.prototype.on_resize = function(width, height) {
     this.group.style.height = height - this.ax_top - 2 * this.border_size;
     this.group.style.marginTop = this.ax_top;
 
+    this.bound_labels.style.height = height - (this.ax_top - 10) - this.border_size ;
```
jgosmann: It is unclear what the relevance of magic number `10` is in this line. This should be a named constant/variable. I wonder whether this height calculation is needed at all ...

jgosmann:
Note that I refactored the sliders. `slider.js` is only a container which contains multiple `SliderControl` instances. I think, the appropriate place to implement those range labels would be `slidercontrol.js` and not `slider.js`.

nengo_gui/static/slider.js
cchan987: > It is unclear what the relevance of magic number 10 is in this line. This should be a named constant/variable. I wonder whether this height calculation is needed at all ...

The height calculation here is used to have the numbers move up and down to correspond with vertical resizing. Agreed, on assigning this 10 somewhere.

cchan987:
> Note that I refactored the sliders. slider.js is only a container which contains multiple SliderControl instances. I think, the appropriate place to implement those range labels would be slidercontrol.js and not slider.js.

Hmm, I was thinking that even if we had 5 sliders lined up, we should only have the min max labelled once so as to prevent it from getting to cluttered.

cchan987:
> I like the fade out animation, but it is strange the the hover box instantly disappears.

Hover box disappears? Can you explain this one a bit more?

jgosmann:
> Hmm, I was thinking that even if we had 5 sliders lined up, we should only have the min max labelled once so as to prevent it from getting to cluttered.

True, as long as all sliders are required to have the same range. But it could still be a more flexible implementation to implement it in `slidercontrol.js` and only display the range for the first `SliderControl` (either by having a flag in `SliderControl` whether to display the range on hover or having the mouse event in `Slider` and toggle the display of the range for the first slider in that event handler).

> Hover box disappears? Can you explain this one a bit more?

If you move the mouse over the slider, you get this gray box around it. If you move the cursor away, that box disappears instantly whereas the range fades out.

tcstewar:
> Hmm, I was thinking that even if we had 5 sliders lined up, we should only have the min max labelled once so as to prevent it from getting to cluttered.

I agree that there should only be one min/max no matter how many sliders there are in the one Component.

cchan987:
For positioning, Would it be preferred to have the range labels follow the first slider guideline left and right while the slider is being resized horizontally, or have the labels to always be on the left hand side of the slider component (current implementation)?

I am in favour the latter option so the labels act as a sort of anchor while resizing horizontally, also moving more items makes resizing a more visually 'busy' operation. Keeping things simple this way.

cchan987:
> The labels can randomly pop up and fade out again during resizing.

this is also the case with the border of all components currently. This should probably be addressed in it's own PR if we want to rectify this.

tcstewar:
> this is also the case with the border of all components currently. This should probably be addressed in it's own PR if we want to rectify this.

If you want to do that as a separate PR, sure, but it's important enough to this PR that I'm pretty sure it'll have to be addressed before this PR is accepted.  (Basically, this PR is making that hover effect do a lot more than it used to, so it makes that resize problem much more obvious)

jgosmann:
See issue #464

tcstewar:
How do people feel about the look of these range indicators?  I think the functionality is about right, but there's something wrong with the look to me.  Perhaps the numbers could be shown in the same font as the numbers in the ranges shown on the graphs?  

Also, how do people feel about the numbers being underneath the sliders themselves?  Here's what it looks like now:

image

jgosmann:
> Perhaps the numbers could be shown in the same font as the numbers in the ranges shown on the graphs?

Would make sense to me.

I don't like that the numbers get covered. That looks like a bug.

cchan987:
By widening the component it can be made so that the range labels are no longer overlapped.
The alternative could be to have the labels in their own column to the left of the sliders, however I feel that makes the labels disconnected.

jgosmann:
How about rotating the labels by 90 degress to use the whitespace inbetween sliders more effectively?

tcstewar:
> How about rotating the labels by 90 degress to use the whitespace inbetween sliders more effectively?

Definitely an option to try out.

I'd also try out actually using the same D3 Axis thing so that there's both the axis labels and a line with three tick marks on the far left of all the sliders.

tbekolay:
> three tick marks on the far left of all the sliders.

All the sliders or just the left one? I could see the tick marks being on all, but only the range labels on the far left one? Since they all share a range, having it repeated would be redundant / cluttered.

tcstewar:
> All the sliders or just the left one? I could see the tick marks being on all, but only the range labels on the far left one? Since they all share a range, having it repeated would be redundant / cluttered.

Sorry, that's me not being clear.  I meant a single separate axis on the far left.  So something like this, only with the numbers correct and everything lined up right:

image

jgosmann:
I think that separate axis would look good. Can we get rid of the tick to the right at 0? Ant the extension to the bottom of the axis?

tcstewar:
> I think that separate axis would look good. Can we get rid of the tick to the right at 0? Ant the extension to the bottom of the axis?

Definitely. Those just didn't get removed because I was making a mockup in a paint program I wasn't familiar with.  :)

cchan987:
> How about rotating the labels by 90 degress to use the whitespace inbetween sliders more effectively?

Updated to demo this 5be35ce , doesn't seem right to me. I will try out the axis idea next.

cchan987:
In 6795a12 I reduced the left margin so the slider guidelines don't cover it unless the component is being squeezed extremely narrow. This seems to be a clean simple fix.

cchan987:
I wonder if it's more convenient to quote commit IDs or should I set up different branches for testing ideas.

tcstewar:
> I wonder if it's more convenient to quote commit IDs or should I set up different branches for testing ideas.

Different branches is useaully preferred, I think -- that way it doesn't clutter up the history.

tcstewar:
> Updated to demo this [rotation] in 5be35ce , doesn't seem right to me. I will try out the axis idea next.

Hmm... yeah, I agree that the rotation doesn't seem to look right

tcstewar:
> In 6795a12 I reduced the left margin so the slider guidelines don't cover it unless the component is being squeezed extremely narrow. This seems to be a clean simple fix.

That's definitely an improvement.  I still find it odd that it's in a completely different font and size from all our other range labels, though.  Are you planning on trying the single axis suggestion?

cchan987:
> Are you planning on trying the single axis suggestion?

see #562

Seanny123:
LGTM. Is there any reason this still needs changes? Other than the fact that the commits need to be squashed together?

tcstewar:
> Is there any reason this still needs changes? Other than the fact that the commits need to be squashed together?

I much prefer the look of what's being attempted in #526 .  As it stands, I'm not sure this PR is an improvement over what we currently have, and I think #526 with some tweaks would work quite well.

PR #459, opened 2015-07-03
reduced sends from code_editor to every 2s

### LINKED ISSUE #444: When GUI slows down, checking for errors for each character is bad
One of the students of the summer school had to deal with very annoying delays because the server would check for errors for each character they typed. Alternatively, could it wait for the user to stop modifying the text file before sending updates? At least for several seconds?

cchan987:
Fixes #444

### DIFF
--- a/nengo_gui/static/ace.js
+++ b/nengo_gui/static/ace.js
@@ -42,6 +42,8 @@ Nengo.Ace = function (uid, args) {
     this.console.appendChild(this.console_stdout);
     this.console.appendChild(this.console_error);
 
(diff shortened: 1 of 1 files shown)


tcstewar:
Hmm, this isn't quite what was suggested for a fix in #444 .  The suggested fix there was to "wait for the user to stop modifying the text file before sending updates? At least for several seconds?"

This patch just reduces how often it sends, rather than waiting until after editting has stopped for 2 seconds.

A more robust approach would be to actually have the server report back when it's done executing the script.

cchan987:
Needs reworking

Issue #358, opened 2015-06-08
Starting up nengo_gui with local locals() causes an error.


xchoo:
If you try to start nengo_gui and pass in local variables to the initialization call (like so: 
```python
...
nengo_gui.viz.Viz(__file__, model=model, locals=locals()).start()
```

the following error occurs:
```Starting nengo_gui server at http://localhost:8080
Traceback (most recent call last):
  File "d:\fchoo\documents\github\nengo_gui\nengo_gui\swi.py", line 276, in hand
le_request
    text = getattr(self, command)(*args[1:], **db)
  File "d:\fchoo\documents\github\nengo_gui\nengo_gui\server.py", line 85, in sw
i
    viz_sim = self.server.viz.create_sim()
  File "d:\fchoo\documents\github\nengo_gui\nengo_gui\viz.py", line 394, in crea
te_sim
    viz_sim = VizSim(self)
  File "d:\fchoo\documents\github\nengo_gui\nengo_gui\viz.py", line 47, in __ini
t__
    self.add_template(template)
  File "d:\fchoo\documents\github\nengo_gui\nengo_gui\viz.py", line 73, in add_t
emplate
    c = template.create(self)
  File "d:\fchoo\documents\github\nengo_gui\nengo_gui\components\component.py",
line 37, in create
    *self.args, **self.kwargs)
  File "d:\fchoo\documents\github\nengo_gui\nengo_gui\components\ace_editor.py",
 line 11, in __init__
    self.current_code = self.viz.viz.code
AttributeError: 'Viz' object has no attribute 'code'```

Looking at the code in ```load``` function in ```nengo_gui/viz.py```, it looks like when you provide the locals variable, the block of code that initializes the ```code``` attribute in the ```viz``` object is skipped over entirely (lines 205 to 212).  

This problem can be skirted by setting ```interactive=False``` in the ```Viz``` object constructor.

**Comment by tcstewar** [comment]
This still seems to be the case, but with a wacky new error:

```
Starting nengo server at http://localhost:8080
ERROR:nengo_gui.server:Error response
Traceback (most recent call last):
  File "c:\users\terry\documents\github\nengo_gui\nengo_gui\server.py", line 233, in do_GET
    self.handle_upgrade()
  File "c:\users\terry\documents\github\nengo_gui\nengo_gui\server.py", line 260, in handle_upgrade
    self.upgrade_to_ws()
[...]

Issue #456, opened 2015-06-30
Refactor server


jgosmann:
The server code is somewhat convoluted and it's hard to figure out which class takes responsibility to manage certain threads. This should be refactored to have a clear hierarchy. A potential problem in this might be that `BaseHTTPServer` was probably never intended to be used with persistent WebSocket connections.

cchan987:
was this addressed as part of #460

tcstewar:
Nope, this is about refactoring server.py and swi.py

jgosmann:
Do we want this in the 0.2 release or should it wait for 0.3? The server code seems to be stable enough at the moment and the refactoring would have no user-facing changes.

If we want it for 0.2, I might start with it soon-ish.

tcstewar:
That's a good point...  There really isn't a pressing need for it to go into 0.2, and I think focusing on the other changes would be a better use of time.

Seanny123:
I know this isn't a huge priority right now, but since I'm building the documentation anyways, is there a specific library that you would consider useful?

Do you want to build something from scratch using [SocketServer](https://docs.python.org/2/library/socketserver.html) as seen in [this Gist](https://gist.github.com/jkp/3136208)? Alternatively, would you prefer using some sort of generic library like either:

- [websockets](https://pypi.python.org/pypi/websockets)
- [simple-websocket-server](https://github.com/dpallot/simple-websocket-server) which certainly seems to be the simplest, since it only has one file
- [python-websocket-server](https://github.com/Pithikos/python-websocket-server)

jgosmann:
I can't answer the question at this point. I still have to evaluate what a good class structure with clear responsibilities would be and if this can be realized with the Python standard libraries.

Issue #449, opened 2015-06-22
Cannot import file from a different folder as script


bjkomer:
This is related to #379 but the fix in #384 doesn't cover this case.

If I run nengo_gui to make a new script, and then try to import something from a different folder, it won't find it, even though the other folder has an `__init__.py` in it.

The importing I'm trying to do is something like this:
```Python
import utils.gain_sets
```
where utils is a folder in the same directory where I run nengo_gui from and a gain_sets.py file in it.

bjkomer:
Actually, it looks like the problem in #379 is still happening in this case too, and importing from a different folder is fine if you open your file with `nengo_gui` or open a file you made from within `nengo_gui`.

The case where no imports work is when you run `nengo_gui` from the command line with no arguments or a new file name.

Issue #446, opened 2015-06-19
Simulator might still run in the GUI


jgosmann:
Creating a model and simulator within a function and running it in this function will cause the GUI to run this simulator.

Seanny123:
Fixed by pull #460.

PR #460, opened 2015-07-04
Refactor viz

tcstewar:
This PR reworks the Viz / VizSim mess.  

The most important change is now there is no state shared across simulations.  Previously, the one Viz object held a lot of information that should be different across different tabs open to nengo_gui.  This caused all sorts of threading problems because one tab could change something that was required by another tab.  A similar thing could happen when reloading a page (since that's the same as opening a new tab).

In the new approach, there is one SimServer object (the new version of Viz), and the only things it holds are immutable default values.  Every page has its own Sim object (the new version of VizSim).  This is the main workhorse.  It maintains all of its own state and is not dependent on SimServer after initialization.

This change also cleaned up the code for executing the script, and now the same code is used for executing the script the first time (on startup) and when code is changed in the text editor.  

I've also added documentation to the new Sim and SimServer classes.  :)

I believe that this reorganization should address the dreaded "pop" bug.  But it's been difficult to replicate that bug, so it's hard to test.  But in any case this is a *much* cleaner code organization and so as long as I haven't broken any features, I think this is the right way to go.  :)


nengo_gui/components/ace_editor.py
```diff
@@ -6,18 +6,18 @@
 import nengo_gui.monkey
 
 class AceEditor(Component):
-    def __init__(self, viz, config, uid):
-        super(AceEditor, self).__init__(viz, config, uid)
-        self.viz = viz
+    def __init__(self, sim, config, uid):
+        super(AceEditor, self).__init__(sim, config, uid, z_order=-8)
```
Seanny123: Any reason in particular for `z_order=-8`? Was there a problem with it being placed on top of stuff?

tcstewar: Good point; I should document that better.  Basically, in the old version there was code scattered all over the place that ensured that SimControl was first in the components list, followed by AceEditor and NetGraph, followed by everything else.  This consolidates all of that into one sort.  

Old code:

https://github.com/nengo/nengo_gui/blob/master/nengo_gui/viz.py#L76
https://github.com/nengo/nengo_gui/blob/master/nengo_gui/viz.py#L140
https://github.com/nengo/nengo_gui/blob/master/nengo_gui/components/netgraph.py#L211

tcstewar: Hmm, and I shouldn't call it z_order.  It's more of a ```component_sort_key``` or ```component_order```.

tcstewar:
Oh, and this fixes #358, #380, #443, #446, and #449

nengo_gui/components/ace_editor.py
tcstewar: I've cleaned up the code and documented this now...  and discovered that you're completely right that AceEditor doesn't need to be in a particular sort order.  So I got rid of that setting.  :)

nengo_gui/components/component.py
```diff
@@ -31,9 +35,9 @@ class Template(object):
     def __init__(self, *args, **kwargs):
         self.args = args
         self.kwargs = kwargs
-    def create(self, vizsim):
-        uid = vizsim.viz.get_uid(self)
-        c = self.cls(vizsim, vizsim.viz.config[self], uid,
+    def create(self, sim):
+        uid = '_uid_%d' % id(self)
+        c = self.cls(sim, sim.config[self], uid,
```
Seanny123: How does this work? I can't find the `cls` attribute defined anywhere and it doesn't seem to be a default attribute.

tcstewar: It's defined in the individual Templates

https://github.com/nengo/nengo_gui/blob/master/nengo_gui/components/value.py#L44

That said, one of the things I'm very much looking forward to is that I'm pretty sure that now that this change has been made, I can get rid of the whole Template/Component thing, getting rid of a lot of magic and annoyance in this code.....

Seanny123:
Other than the comments I left, LGTM.

nengo_gui/components/component.py
jgosmann: I think, `Template` should at least declare `cls` with some default value (`None`, `object`, or whatever) just to make clear that this attribute is supposed to exist (and to be set by subclasses). Or is this not possible because it is a class attribute?

nengo_gui/components/netgraph.py
```diff
@@ -16,74 +16,65 @@
...
         self.to_be_sent = collections.deque()
         self.uids = {}
         self.parents = {}
-        self.networks_to_search = [self.viz.model]
+        self.networks_to_search = [self.sim.model]
         self.initialized_pan_and_zoom = False
-        try:
-            self.last_modify_time = os.path.getmtime(self.viz.viz.filename)
-        except OSError:
+        if self.sim.filename is None:
```
jgosmann: Is this additional `if` statement necessary? Isn't this case covered by the `try`-`catch` already?

nengo_gui/components/netgraph.py
```diff
@@ -16,74 +16,65 @@
...
+            except OSError:
+                self.last_modify_time = None
         self.last_reload_check = time.time()
 
     def check_for_reload(self):
-        try:
-            t = os.path.getmtime(self.viz.viz.filename)
-        except OSError:
-            t = None
+        if self.sim.filename is not None:
```
jgosmann: As above, why is this `if` introduced?

nengo_gui/components/netgraph.py
```diff
@@ -179,23 +170,20 @@ def _reload(self, code=None):
 
                 self.uids[uid] = new_item
 
-        self.to_be_expanded.append(model)
+        self.to_be_expanded.append(self.sim.model)
+
+        t_uids = [self.sim.get_uid(c.template) for c in self.sim.components]
```
jgosmann: Is `t_uids` used anywhere below? (Maybe I just missed it because the function is so long)
What is the `t` supposed to stand for? Is there a more expressive name that could be used?

nengo_gui/components/netgraph.py
tcstewar: It's not an OSError if self.sim.filename is None.  Also, I figured it was clearer to explicitly show that I was checking for this, rather than catching TypeError and and adding a comment saying that I'd expect a TypeError in the cases where filename is None (which is what happens all the time in the IPython integration case).

nengo_gui/components/netgraph.py
jgosmann: oh, it is used below ... can this be moved closer to the actual usage place?

nengo_gui/main.py
```diff
@@ -23,9 +24,12 @@ def main():
         import logging
         logging.basicConfig(level=logging.DEBUG)
 
-    viz = nengo_gui.Viz(filename=args.filename)
-    viz.default_backend = args.backend
-    viz.start(port=args.port, password=args.password)
+    #viz = nengo_gui.Viz(filename=args.filename)
+    #viz.default_backend = args.backend
+    #viz.start(port=args.port, password=args.password)
```
jgosmann: Delete commented out code.

nengo_gui/components/netgraph.py
tcstewar: Yeah, a better name would be good...  probably ```template_uids``` or even ```old_template_uids```, since it's the uids of the templates for the old components.  That's also why this code is up here -- I have to compute these before wiping out the names and loading everything new, so this is as close as I could get it to where they are used.

That said, the plan is to wipe out the whole template thing...  still, a better comment/name here would be good.  :)

nengo_gui/sim.py
```diff
@@ -0,0 +1,436 @@
...
+        self.stdout = ''
+
+        patch = nengo_gui.monkey.Patch(self.filename)
+        try:
+            with patch:
+                exec(code, locals)
+        except nengo_gui.monkey.StartedSimulatorException:
+            # no running a simulator inside a script
+            pass
+        except nengo_gui.monkey.StartedVizException:
```
jgosmann: Should these cases produce a warning because there might be code that did not get executed?

nengo_gui/sim.py
```diff
@@ -0,0 +1,436 @@
...
+            # modify the model for the various Components
+            for c in self.components:
+                c.add_nengo_objects(self)
+
+            # determine the backend to use
+            backend = importlib.import_module(self.backend)
+            # if only one Simulator is allowed at a time, finish the old one
+            old_sim = Sim.singleton_sims.get(self.backend, None)
+            if old_sim is not None:
+                if old_sim is not self:
```
jgosmann: Combining the two if statements with `and` would save one line and an indentation level which would be a minor readability improvement IMO.

nengo_gui/main.py
tcstewar: Oops.  Thanks -- fixed now.

nengo_gui/sim_server.py
```diff
@@ -0,0 +1,115 @@
...
+        return sim
+
+    def remove_sim(self, sim):
+        self.sims.remove(sim)
+
+    def count_sims(self):
+        return len(self.sims)
+
+
+class Viz(SimServer):
```
jgosmann: What is the purpose of this class? If it is only for internal debugging, it should be removed once it is ensured that all occurrences of `Viz` have been replaced (which shouldn't be too hard). If it's for compatibility with existing user scripts calling `Viz`, this class should be at the same import path it used to be.

nengo_gui/sim.py
tcstewar: Probably a good idea.

nengo_gui/sim.py
tcstewar: weird...  why did I do that?  No idea.....  I'll combine them..  :)

nengo_gui/components/netgraph.py
jgosmann: Hm, not sure what the best solution is. My worry is that the code currently is slightly convoluted because it handles an error condition, the success condition, and another error condition in this order. It is probably easier to follow if all error conditions are handled either before or after the success condition.

nengo_gui/sim_server.py
tcstewar: Good point.  It was for the compatibility with other scripts, and I'd been assuming they'd use ```nengo_gui.Viz```, but you're right that they might be using ```nengo_gui.viz.Viz```.

jgosmann:
This branch breaks the IPython integration.

Further comments:
* The way `sim.new_code` is used seems a bit convoluted because (if I understand in correctly) it has side effects at some point later in time and different classes other than `Sim` manipulate it. This doesn't seem like good encapsulation.
* The `NetGraph.reload` function is quite long which makes it hard to understand and keep the overview. It would be good to split it up into smaller functions, I think.
* As @Seanny123 mentioned, `Patch` is not a very descriptive name. Maybe something like `GuiExecutionEnv` is better?
* Should `Viz` renamed to `GUI`? (#266)

nengo_gui/sim.py
```diff
@@ -0,0 +1,435 @@
...
+        self.config_save_needed = False
+        self.config_save_time = None   # time of last config file save
+
+        self.uid_prefix_counter = {}
+
+        self.create_components()
+
+    def create_components(self):
+        """Generate the actual Components from the Templates"""
+        self.components = []
```
jgosmann: There are a number of attributes (this is just the first one) in this class which are not defined within the `__init__` function which would be preferable (especially if there are obvious default values like the empty list). That would also help with auto-completion in some editors.

nengo_gui/sim.py
```diff
@@ -0,0 +1,435 @@
...
+        self.locals['nengo_gui'] = nengo_gui
+        self.locals['_viz_config'] = config
+        fname = self.filename_cfg
+        if os.path.exists(fname):
+            with open(fname) as f:
+                config_code = f.readlines()
+            for line in config_code:
+                try:
+                    exec(line, self.locals)
+                except Exception as e:
```
jgosmann: `e` is not used

nengo_gui/sim.py
tcstewar: Added warning messages!  We don't have a nice way of doing warnings (as opposed to errors), so I just put it the same place print statements go to.  I think in a future PR we could do something nicer with this, with a way to show debug and other logging statements....

nengo_gui/sim.py
```diff
@@ -0,0 +1,436 @@
...
+                    if self.sim_server.interactive:
+                        logging.debug('error parsing config: %s' % line)
```
jgosmann: `logging.debug('error parsing config: %s', line)` works too and is apparently preferred.

nengo_gui/sim.py
```diff
@@ -0,0 +1,436 @@
...
+        else:
+            webpage_title_js = ''
+
+        assert isinstance(self.components[0], nengo_gui.components.SimControl)
+
+        component_js = '\n'.join([c.javascript() for c in self.components])
+        component_js += webpage_title_js
+        if not self.sim_server.allow_file_change:
+            component_js += "$('#Open_file_button').addClass('deactivated');"
+            pass
```
jgosmann: `pass` is unnecessary here

nengo_gui/sim_server.py
```diff
@@ -0,0 +1,115 @@
+import os
+import warnings
+
+import nengo
```
jgosmann: unused import

nengo_gui/sim_server.py
```diff
@@ -0,0 +1,115 @@
...
+        print("Starting nengo_gui server at http://localhost:%d" % port)
+        if password is not None:
+            nengo_gui.server.Server.add_user('', password)
+            addr = ''
+        else:
+            addr = 'localhost'
+        nengo_gui.server.Server.start(self, port=port, browser=browser,
+                                      addr=addr)
+
+    def prepare_server(self, viz, port=8080, browser=True):
```
jgosmann: `viz` is an unused argument and could be removed (this was already the case before the refactoring and I don't know why)

jgosmann: Maybe to match the `swi` method with the same name? (But the function signatures are not exactly identical.) If that's the case is should be renamed to `sim_server`.

tcstewar:
> This branch breaks the IPython integration.

What the....  ??  It wasn't broken this morning.  Hmm, and if I revert to d64ad00 the integration still works...  okay, which one of my tiny little bug fixes killed things....  hmm....

jgosmann:
btw: `git bisect` is a useful tool to locate the first bad commit

On last comment:
It would be nice if all files using the `print` statement/function would include `from __future__ import print_function` to ensure the same behavior in Python 2 and 3.

tcstewar:
> The way sim.new_code is used seems a bit convoluted because (if I understand in correctly) it has side effects at some point later in time and different classes other than Sim manipulate it. This doesn't seem like good encapsulation.

Hmm, good point...   Hmm, how about if I get rid of Sim.new_code and just have AceEditor do something like ```sim.net_graph.set_new_code(new_code)``` directly?  I don't think there's a reason to have the intermediary any more....

tcstewar:
> The NetGraph.reload function is quite long which makes it hard to understand and keep the overview. It would be good to split it up into smaller functions, I think.

Hmm...   possibly.  It's a bit ugly to do so as there's a lot of shared state across those aspects of the function, but I'll stare at it a bit.  :)

jgosmann:
Sounds good as far as I can tell (not sure if the intermediary is needed or not). I would probably call it `update_code(new_code)` instead of `set_new_code` as I wouldn't be surprised if the former would have some more side effects than just setting a variable.

tcstewar:
> As @Seanny123 mentioned, Patch is not a very descriptive name. Maybe something like GuiExecutionEnv is better?

<sigh>  I liked ```monkey.Patch``` as a name...  ;)  But, it is causing more confusion than the cuteness of the name is worth.  I think I'll go with ```ExecutionEnvironment``` or something like that...  :)

tcstewar:
> Should Viz renamed to GUI?

Which Viz?  The only Viz that's left is the one that's specifically there to have the same name to allow old scripts to run.

However, I could see ```SimServer``` being renamed to something.  Not sure what the best name would be....  I could maybe see GUI, which would mean new scripts could trigger the gui with ```nengo_gui.GUI(__file__).start()```.  That still seems a little ugly though...  Not sure if there are any other good suggestions for it....

nengo_gui/components/action.py
```diff
@@ -64,7 +64,7 @@ def __init__(self, viz_sim, component, new_cfg, old_cfg):
 
     def load(self, cfg):
         for k, v in iteritems(cfg):
-            setattr(self.viz_sim.viz.config[self.component.template], k, v)
+            setattr(self.viz_sim.config[self.component.template], k, v)
```
jgosmann: Is this supposed to be `viz_sim`?

jgosmann:
> Which Viz? The only Viz that's left is the one that's specifically there to have the same name to allow old scripts to run.

I think, you're right. No idea what I was thinking.

tcstewar:
> This branch breaks the IPython integration.

Should be fixed now.  Turns out that the AceEditor does have to be near the beginning of the list.  :)

(note: in a future PR, I'd like to separate things so that there are fixed components (SimControl, AceEditor, and NetGraph) and non-fixed components (everything else))

tcstewar:
> Sounds good as far as I can tell (not sure if the intermediary is needed or not). I would probably call it update_code(new_code) instead of set_new_code as I wouldn't be surprised if the former would have some more side effects than just setting a variable.

Fixed.  :)  Much cleaner that way.  Thank you for the suggestion!  (and I went with ```update_code```)

jgosmann:
> Should be fixed now.

It is. :smiley:

nengo_gui/components/action.py
tcstewar: Ah, that should be ```sim```....    Hmm, and scrolling down a bit it looks like I completely didn't update the RemoveGraph undo function...  oops....

tcstewar: Okay, those are fixed now.  :)

nengo_gui/sim_server.py
tcstewar: Good catch.  Removed!  :)

nengo_gui/components/netgraph.py
tcstewar: I've reworked the logic in both places a bit.  I think this is slightly clearer.

Also, I realized that we weren't updating the code editor in the browser if the file changes -- oops.  That's fixed now too.

nengo_gui/sim.py
tcstewar: Good point -- that class got pretty bad for that.  I've added some explicit initialization and comments in ```__init__```

Seanny123:
Just looked at your new commits. They seem to have fixed the issues that Jan and I had. LGTM.

tcstewar:
I think the last thing I want to stare at is refactoring the reload() code a bit....  other than that I feel pretyt good on this change.  Thanks to both of you for all the feedback and comments!

tbekolay:
I still want to look at this in more detail, but so far it looks really great! :D 

One thing that is a bit confusing reading through the code top to bottom are the names `Sim` and `SimServer`. Looking through a lot of Nengo code makes has trained me to think that `self.sim = sim` means that you're assigning a `nengo.Simulator` (or another backend equivalent) instance, but I think in most cases that's not true here. While it is of course true that the Nengo GUI stuff is visualizing simulations and simulators, I think we should shy away from using that term here unless we're actually referring to simulator instances ...

My understanding of the architecture is that there is one `SimServer` which manages multiple `Sim`s. But to start `nengo_gui` from a script, you make an instance of `SimServer`, so that's what the user will mostly be seeing.

How about we rename `SimServer` to just `Server` (user scripts will create a `nengo_gui.Server` object) and renamed `Sim` to `Page` to avoid the confusion between a `nengo.Simulator` instance and a `nengo_gui.Sim` instance? I think it is true that every HTML page has one and only one associated `Page` (now `Sim`). If that's not true, then maybe rename it to `View` or something like that? Just something short that can't be confused with `Simulator`, but still makes sense for what it is...

nengo_gui/viz.py
```diff
@@ -1,419 +1,11 @@
...
-        viz_sim = VizSim(self)
-        self.viz_sims.append(viz_sim)
-        return viz_sim
-
-    def remove_sim(self, viz_sim):
-        self.viz_sims.remove(viz_sim)
-
-    def count_sims(self):
-        return len(self.viz_sims)
+# dummy class so that old scripts still work, but get a deprecation message
```
tbekolay: I can't imagine this will actually break anything for anyone, but it's a bit weird to me that calling this, I would get a different type than if I'd just made the `SimServer`. Could instead do this:
```python
def Viz(*args, **kwargs):
    warnings.warn("nengo_gui.Viz is deprecated. Use nengo_gui.SimServer instead.")
    return nengo_gui.SimServer(*args, **kwargs)
```

nengo_gui/viz.py
```diff
@@ -1,419 +1,11 @@
...
+class Viz(nengo_gui.sim_server.SimServer):
+    def __init__(self, *args, **kwargs):
+        warnings.warn("nengo_gui.Viz is deprecated.  "
```
tbekolay: This might be controversial? But I think it would be good to get out of the habit of using two spaces between sentences. It's a relic of the typewriter days, and even though this text will be in fixed-width font, it looks really weird to me to see two spaces after a period. I know it's muscle memory, and hard to change, but I think it's worth the week or so of training to conform to modern typography conventions. For my own amusement, I'll post a different [pedantic blog post](http://www.slate.com/articles/technology/technology/2011/01/space_invaders.html) each time I see this in future PRs (unless people strongly think I should let this go) :rainbow:

jgosmann: +1

drasmuss: Doesn't pep8 say two spaces after a period? https://www.python.org/dev/peps/pep-0008/#comments

tbekolay: My god. To be fair, they're talking about in comments, but that still grinds my gears!

tbekolay: Interesting [discussions](http://stackoverflow.com/questions/21571994/why-does-pep8-suggest-using-two-spaces-in-comments) and [comments](https://www.reddit.com/r/Python/comments/2zjqkl/do_you_follow_pep_8/cpjpcjj) on this.

tcstewar:
> How about we rename SimServer to just Server (user scripts will create a nengo_gui.Server object) and rename Sim to Page to avoid the confusion between a nengo.Simulator instance and a nengo_gui.Sim instance? I think it is true that every HTML page has one and only one associated Page (now Sim). If that's not true, then maybe rename it to View or something like that? Just something short that can't be confused with Simulator, but still makes sense for what it is...

Hmm...  I definitely agree Sim should be renamed...  I like Page, as it is the case that it's now one Page to a web page.  :)

As for renaming SimServer, I'm more tempted to call it something like GUI, since if people are manually creating it I think ```nengo_gui.GUI(__file__).start()``` is a bit better than ```nengo_gui.Server(__file__).start()```.  Plus, we already have a ```Server```.  :)  Although I do think it may be possible to merge or clarify what's going on between ```SimServer``` and ```Server``` as part of #456

nengo_gui/viz.py
tcstewar: Ooo, I like that approach.

nengo_gui/viz.py
tcstewar: I've always been a two-space person....  But I think this is the only time I've seen any support for that point of view (and we all know that my instincts on code style really shouldn't be listened to...).  Still, in a fixed-width font situation it makes a lot of sense to me....  But I'm definitely willing to be outvoted on this one.  :)

Seanny123:
@tcstewar are you still looking for a way to replicate the pop bug? I have a script that does it every time I open it.

tcstewar:
> @tcstewar are you still looking for a way to replicate the pop bug? I have a script that does it every time I open it.

yes please!

Seanny123:
@tcstewar sent it to you by email

tcstewar:
> @tcstewar are you still looking for a way to replicate the pop bug? I have a script that does it every time I open it.

Sweet!  That's a really interesting version of the pop bug...  I'm going to make a separate issue for it.  This PR is still susceptible to that bug, but it recovers from it well, so that makes me happy.  :)  Thank you!

tcstewar:
Okay, I think I'm happy with this now.  Thank you everyone for all the comments and suggestions!

```reload``` isn't quite as refactored as it could be, but I'm going to hold off on that until I get rid of Templates, since that'll also simplify that a bit.

tbekolay:
Cool, looks good to me! Will be glad to get rid of templates also ;)

PR #461, opened 2015-07-04
Filtered value plots in nengo, not javascript

tcstewar:
This reduces the amount of computation in javascript, and it greatly improves the look of nengo_spinnaker plots (since they're not guaranteed to sample at a regular rate, so doing the filtering after sampling is a bad idea).

### DIFF
--- a/nengo_gui/components/value.py
+++ b/nengo_gui/components/value.py
@@ -20,7 +20,7 @@ class Value(Component):
         with viz.model:
             self.node = nengo.Node(self.gather_data,
                                    size_in=self.obj.size_out)
-            self.conn = nengo.Connection(self.obj, self.node, synapse=None)
+            self.conn = nengo.Connection(self.obj, self.node, synapse=0.01)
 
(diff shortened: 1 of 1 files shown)


Seanny123:
LGTM

tbekolay:
This particular change looks good to me. Is there any ideas as to how we might make the chosen filter configurable? 10 ms lowpass filters don't make sense on all connections. It seems like this might be hard to do per connection if it needs a call back to Python each time, but perhaps it's equivalent to what would be needed in Javascript anyhow.

tcstewar:
> Is there any ideas as to how we might make the chosen filter configurable?

The only drawback to this change is that If we do make the filter configurable in the future, then it'll force a rebuild (since it'd modify the Connection).  I'm a little annoyed at that, but the improved smoothness on non-fixed-dt backends more than makes up for that.  

I think interface-wise, we should add an option in the right-click menu on the graphs to set the filter...  We could also go the global-setting approach that we did in 1.4, but I'm not sure that's right either.

PR #466, opened 2015-07-06
Prefiltered xy plots as well

tcstewar:
This applies the #461 fix (filtering in nengo, rather than in javascript) to XY plots, for consistency


Seanny123:
LGTM

PR #473, opened 2015-07-08
Tutorials

tcstewar:
Here's a collection of tutorials that have come out of the various hands-on nengo demos I've done.  I tried to add a bit of text to each one that vaguely maps on to the sorts of things I've said when presenting them.  I still have to add some text to some of the spa examples, but it's pretty close.


Seanny123:
Although I didn't read through every bit of text, this looks like a good thing to merge.

nengo_gui/examples/tutorial/24-spa-unbinding.py
```diff
@@ -0,0 +1,23 @@
...
+
+model = spa.SPA()
+with model:
+    model.color = spa.Buffer(D)
+    model.shape = spa.Buffer(D)
+    model.memory = spa.Memory(D, synapse=0.1)
+    model.query = spa.Buffer(D)
+    model.answer = spa.Buffer(D)
+
+    actions = spa.Actions(
```
Seanny123: These constants should be explained.

nengo_gui/examples/tutorial/25-spa-parse.py
```diff
@@ -0,0 +1,26 @@
+# Tutorial 25: Parsing Simple Commands
```
Seanny123: What's the expected functionality of this network?

nengo_gui/examples/tutorial/20-spa-actions.py
```diff
@@ -0,0 +1,61 @@
...
+# that computes the utility of the action (how good the action is right
+# now) and what the effect of the action should be (what to do when
+# the action is chosen).  The Basal Ganglia model will monitor the
+# utilities of each action and select the one with the highest value.
+#
+# The following code defines four actions.  The utilities look like
+# this: "dot(vision, DOG)".  This means that the utility of the action
+# is the dot product of whatever vector is in the vision Buffer and
+# the ideal vector for DOG.  This will be a value near 1 if you see a
+# dog, and will be near zero otherwise.  The close what you see is to
```
s72sue: tiny typo: should it be "closer" ? :)

nengo_gui/examples/tutorial/20-spa-actions.py
```diff
@@ -0,0 +1,61 @@
...
+# an action selection network matches well to human data:
+# http://compneuro.uwaterloo.ca/files/publications/stewart.2012a.pdf
+#
+# This model has been built in to the Nengo SPA system for use in your
+# models.  To use it, for every action you need to define some function
+# that computes the utility of the action (how good the action is right
+# now) and what the effect of the action should be (what to do when
+# the action is chosen).  The Basal Ganglia model will monitor the
+# utilities of each action and select the one with the highest value.
+#
```
s72sue: Would it be useful to add a line about how thalamus is used to implement the effects for the bg?

nengo_gui/examples/tutorial/25-spa-parse.py
s72sue: In theory I think this is the functionality:  If you are shown a command to SAY or to WRITE -> then store that as a verb in memory e.g., if you are shown SAY, store SAY*VERB*0.4 in memory. Similarly if you are shown either A,B,C or D -> then store it as a noun in memory. If you are not shown any of the above, and the verb in memory is either WRITE or SAY, then depending on whether it is WRITE or SAY, either write or speak the noun stored in memory. 

But it doesn't quite work for me when I run the model. This might be dependant on the 0.4 factor that is being used (specially because the same memory is being used to store both the verb and the noun).

s72sue:
Overall I am pretty happy with the tutorials, would be definitely useful to beginners. Thanks Terry! =)

nengo_gui/examples/tutorial/25-spa-parse.py
tcstewar: Should behave better now...  I've gotten rid of the constants and added a description.  :)

nengo_gui/examples/tutorial/20-spa-actions.py
tcstewar: Good idea.  Added.  :)

nengo_gui/examples/tutorial/20-spa-actions.py
tcstewar: Thanks!

nengo_gui/examples/tutorial/24-spa-unbinding.py
tcstewar: Good idea...  in fact, better yet, they can be removed.  They were just there to slow down how quickly things were shoved in to memory, but an even better way of doing that is increasing the synapse time a bit (since that improves the memory as well).

nengo_gui/examples/tutorial/25-spa-parse.py
s72sue: works great! :)

nengo_gui/examples/tutorial/24-spa-unbinding.py
s72sue: I like increasing the synapse instead of using constants. Constants are a
bit confusing.

s72sue:
LGTM, should be merged ...

PR #475, opened 2015-07-09
Allows all valid semantic pointer expressions in modal

### LINKED ISSUE #383: Setting value for semantic pointer plot validation
This should be a valid semantic pointer, but I can't enter it into the "set value..." dialog for the semantic pointer plot:  ```0.3*A + 0.7*B```.

The work-around right now is to do ```A * 0.3 + B*0.7```.  However, that trick doesn't work for ```(A+B)*0.5```

celiasmith:
fixes #383

The modal on semantic pointer graphs now checks with the server if the entered pointer expression is a good one, so it should allow all the same expressions as nengo.spa


tcstewar:
Cool!   The functionality seems pretty robust to me.  :)

One small bug: Hitting "cancel" seems to clear the stored semantic pointer, rather than cancelling.  So if you set it to A, then pull up the Set value... thing again, and then hit cancel, the A will go away.

Also, could we add some suggestions to the error message?  Instead of "Invalid semantic pointer expression." how about "Invalid semantic pointer expression.  Semantic pointers must start with a capital letter.  You can also use mathematical operators such as +, * (circular convolution), and ~ (pseudo-inverse), so (A+~(B*C)*2)*0.5 would be a valid semantic pointer expression."

tcstewar:
Hey @celiasmith I finally got a chance to poke at this, and I think I've got a decent fix.  Basically, I just changed the communication protocol so that there's a special message ```':empty:'``` that means the same as what used to be sent as ```''```.  That gets rid of the problem of the infinitely repeating empty messages....

I also added a quick fix that means that any terms added during a "check only" parse don't end up staying in the vocab.

nengo_gui/components/pointer.py
```diff
@@ -70,8 +70,18 @@ def javascript(self):
...
             self.override_target = None
+        elif msg[0:12] == ':check only:':
+            if len(msg) == 12:
+                self.data.append("good_pointer")
+            else:                
+                vocab = copy.deepcopy(self.vocab_out)
+                try:
+                    vocab.parse(msg[12:])
+                    self.data.append("good_pointer")
+                except:
```
Seanny123: This bare exception doesn't feel right to me. Then again, should this wait to be resolved until we have custom Nengo errors?

tcstewar: Hmm, I don't think there'll be a custom Nengo error for this.  The problem is that there are a large number of different things that could be thrown there, depending on what bizarre thing someone gives to ```parse```.  And no matter what ```Exception``` someone manages to generate in there, we want to handle all of them with the response ```bad_pointer```.  So I think this is a legit use of a bare ```except```.

Seanny123: Right, I forgot that aspect of the vocab parsing. I'm okay with this now.

celiasmith:
Ah, nice, thanks @tcstewar ... this looks good to me now.  I presume you are still the right one to merge :)

Issue #349, opened 2015-06-07
Indication of server connection status


jgosmann:
It would be nice to have an indication in the GUI of the connection status to the server to be easily able to tell whether the server is still alive or shutdown.

tcstewar:
Ooo...  very good idea.  +1

cchan987:
closed as per #480
Related issue opened post-merge #492

PR #480, opened 2015-07-13
disconnection warning with modal

cchan987:
When the simcontrol websocket closes, warning modal is displayed.

User is prompted to restart server and refresh page 

In regards to #349

### DIFF
--- a/nengo_gui/static/modal.js
+++ b/nengo_gui/static/modal.js
@@ -52,6 +52,10 @@ Nengo.Modal.prototype.footer = function(type, ok_function, cancel_function){
         $('#confirm_reset_button').on('click', function() {
             toolbar.reset_model_layout();
         });
+    } else if (type === 'refresh') {
+        this.$footer.append('<button type="button" ' +
+            'id="refresh_button" class="btn btn-primary">Refresh</button>');
+        $('#refresh_button').on('click', function() {location.reload()})
     } else {
         console.warn('Modal footer type ' + type + ' not recognized.')
     }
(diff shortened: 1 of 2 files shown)


tcstewar:
seems pretty solid to me, but I think the error message itself needs some tweaking.  Right now it says 

```
Server Communication Error
Please ensure that the GUI server is running then refresh page
```

How about something like this:

```
Nengo has stopped running
To continue working with your model, re-run nengo_gui and click Refresh.
```

tcstewar:
looks good...  merging...

jgosmann:
I think, the red background looks too aggressive. Maybe we could do something more unobstrusive like "connection lost" text in the toolbar (which can have a lighter red background)?

tcstewar:
I could also see just getting rid of the colour change...  the modal does a pretty good job on its own.

jgosmann:
Yeah, that would also be fine. Though, some form of indication can help to figure out wether the server corresponding to a realy old tab (where I might have closed the modal) is still running.

tcstewar:
> Though, some form of indication can help to figure out wether the server corresponding to a realy old tab (where I might have closed the modal) is still running.

Maybe have it show a message in the console?

jgosmann:
I like the Jupyter way with the red connection message and that broken link symbol to the right (which gives you a tooltip explaining the meaning:
snapshot2

tcstewar:
Since this PR has closed, I've started a new issue to track this discussion.  :)  #492

Issue #267, opened 2015-06-02
Setting range on plot does not redraw it


jgosmann:
Thus, the displayed plot is wrong until it's resized.

cchan987:
fixed by #481

PR #481, opened 2015-07-13
Redraw plot

cchan987:
Redraws the plot so lines are in their correct spot as soon as set range is finished

Regards #267


tcstewar:
Looks good to me.

PR #483, opened 2015-07-13
Handle build and run errors

tcstewar:
Exceptions thrown during the build process and when running the model should now be caught by nengo_gui and displayed to the user.

### DIFF
--- a/nengo_gui/page.py
+++ b/nengo_gui/page.py
@@ -187,22 +187,22 @@ class Page(object):
         self.error = None
         self.stdout = ''
 
-        patch = nengo_gui.exec_env.ExecutionEnvironment(self.filename)
+        exec_env = nengo_gui.exec_env.ExecutionEnvironment(self.filename)
         try:
-            with patch:
+            with exec_env:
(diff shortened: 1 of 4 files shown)


nengo_gui/page.py
```diff
@@ -416,11 +416,21 @@ def build(self):
                 old_sim.sim = None
                 old_sim.finished = True
 
+            patch = nengo_gui.exec_env.ExecutionEnvironment(self.filename,
```
Seanny123: Why is this variable called `patch`?

tcstewar: Oops, good catch -- that's a holdover from the old name.  Changed to ```exec_env``` now.  :)

Seanny123:
LGTM, from a syntax and style basis. I don't know the code well enough to know if there was a more elegant way of accomplishing this, so it's up to you if you want to wait for someone more familiar with the code base.

tcstewar:
I just caught a debug message I left in there...  so I'll give it another day or so before merging.  :)  But I'm pretty happy with this code organization.

PR #484, opened 2015-07-13
Fix graphs appearing on top of editor.

jgosmann:
Fixes https://github.com/nengo/nengo_gui/pull/445#issuecomment-121063929


tcstewar:
Fixes it for me!  Thank you again!

PR #485, opened 2015-07-14
deactivate hotkeys when modal is active

### LINKED ISSUE #385: Disable hotkeys when modal is shown
Right now, we have the same problem as before with the code editor... if you press the spacebar when typing in one of the modal input boxes, the simulation starts up, which we don't want. We should disable all the hotkeys when the modal is shown.

cchan987:
Built in a flag to see if the modal is open or not. If it's open, hotkeys are deactivated.
Fixes #385

### DIFF
--- a/nengo_gui/static/hotkeys.js
+++ b/nengo_gui/static/hotkeys.js
@@ -4,62 +4,62 @@ Nengo.Hotkeys = function () {
     this.active = true;
     
     document.addEventListener('keydown', function(ev) {
-    if (self.active) {
+        if (self.active) {
 
-        var on_editor = (ev.target.className === 'ace_text-input');
+            var on_editor = (ev.target.className === 'ace_text-input');
 
-        if (typeof ev.key != 'undefined') {
-            var key = ev.key;
-        } else {
-            switch (ev.keyCode) {
-                case 191:
-                    var key = '?';
-                    break;
-                default:
-                    var key = String.fromCharCode(ev.keyCode)
+            if (typeof ev.key != 'undefined') {
+                var key = ev.key;
+            } else {
+                switch (ev.keyCode) {
+                    case 191:
+                        var key = '?';
+                        break;
+                    default:
+                        var key = String.fromCharCode(ev.keyCode)
+                }
+            }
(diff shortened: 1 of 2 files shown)


tcstewar:
Handy.  :)

One minor thing that might be nice to change about this is to use a ```set_active(true)``` and ```set_active(false)``` instead of a ```toggle_hotkeys()``` function.  Basically, in this case there isn't much difference, but the moment we ever add a third thing that needs to control the state of the hotkeys, then we might get into a situation where toggle_hotkeys() gets called by someone else, and now all of a sudden the hotkeys are only active when the modal *is* displayed, rather than when it isn't.

cchan987:
Updated

nengo_gui/static/hotkeys.js
```diff
@@ -65,5 +69,17 @@ Nengo.Hotkeys.prototype.callMenu = function () {
     Nengo.modal.help_body();
     Nengo.modal.show();
 }
+ 
+//Set active is provided with a boolean argument, which will either
+//turn the hotkeys on or off
+Nengo.Hotkeys.prototype.set_active = function(bool) {
+    console.assert(typeof(bool) == 'boolean')
```
tcstewar: I think this would be a bit more concise as

```this.active = bool;```

File: nengo_gui/static/hotkeys.js [hunk 2]
```diff
@@ -1,7 +1,10 @@
 Nengo.Hotkeys = function () {
     var self = this;
 
+    this.active = true;
+    
     document.addEventListener('keydown', function(ev) {
+    if (self.active) {
```
tcstewar: I think I'm going to be picky here an enforce indenting.  At first glance, when looking over the code I misread things twice because of the indenting -- first I wondered how ```if (self.active)``` could ever be false when three lines up from that it was set to true, and then I stumbled over the lined-up closing braces at the end of the if block....  Indenting the contents of the AddEventListener would help.  :)

cchan987:
^updated

tcstewar:
Looks good to me!  Thanks!

PR #487, opened 2015-07-15
Fix Reset Button bugs

### LINKED ISSUE #467: Resetting the simulation while it is running leaves the gear spinning at the bottom
If you press the reset button without stopping the simulation first, the simulation will be stopped and reset correctly, but instead of a play button on the bottom right, there will be a spinning gear. You now need to click that gear to turn it into the arrow and click it again to play. It should just be an arrow right away.

cchan987:
Displayed time now goes to 0 upon reset press
Gear no longer spins on reset press
Only 1 click required to rebuild/re-play 
Fixes #467

### DIFF
--- a/nengo_gui/components/sim_control.py
+++ b/nengo_gui/components/sim_control.py
@@ -30,7 +30,7 @@ class SimControl(Component):
         self.last_status = None
         self.next_ping_time = None
         self.send_config_options = False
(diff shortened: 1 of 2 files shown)


tcstewar:
Nice!  That's a huge improvement.  :)  Looks good to me!

PR #488, opened 2015-07-15
Stop moving sliders easily

### LINKED ISSUE #214: Quick way to stop moving sliders
Currently the only way to stop a moving slider is to right click and set value.

It is very hard to grab and take control otherwise.

Should we put a quick 'take control' option/button in?

cchan987:
Stop a moving slider by clicking on the associated guideline
Fixes #214

### DIFF
--- a/examples/basic.py
+++ b/examples/basic.py
@@ -11,7 +11,6 @@ with model:
     nengo.Connection(stimulus_B, ens[1])
     nengo.Connection(ens, result, function=lambda x: x[0] * x[1],
                      synapse=0.01)
-    nengo.Node(np.sin)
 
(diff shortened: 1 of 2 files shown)


tcstewar:
Interesting...  Seems to work well, although it's a little weird that the slider stops moving *after* I release the mouse, rather than when I start the click.  What about changing it from a ```click``` handler to a ```mousedown```?

Ideally, I think it would jump the slider to where your mouse currently was so you could start the sliding behaviour right away, but this is still an improvement.  :)

cchan987:
updated

tcstewar:
Looks good to me!

Issue #431, opened 2015-06-16
Slider disappears when making value plot


cchan987:
First make a slider for a node. 
Next make a value plot for the same node.
The slider disappears.

tcstewar:
Fixed by #491

Issue #307, opened 2015-06-04
Hidden state of labels isn't saved in cfg


celiasmith:
If you hide a label on a graph/slider, then close and reopen, it's shown.

tcstewar:
Fixed by #491

PR #491, opened 2015-07-16
Remove component templates

tcstewar:
This gets rid of the Component/Template distinction that complicates the server-side code.

Now that Page and GUI have been reorganized, the entire reason for Templates is gone.  They were originally there because the config file was shared across multiple Viz objects, and so the config file was processed before the actual Components themselves existed.  So the config file actually configured these Component Templates, and the Components (when they were created when a page was actually generated) got their information from the Templates.

But now we don't look at the config file until after the Page is being created, so all that indirection is unnecessary, and so this PR gets rid of the Templates.

Some notes:
 - This slightly changes the .cfg file format, in that it is no longer necessary to refer to ValueTemplate in that file, and it's just Value instead.  For backwards compatibility, ValueTemplate still works (otherwise all old .cfg files would fail).
 - There used to be two separate ```uid```s for the Component and the Template.  The one for the component was just ```id(self)``` and was used by the websocket infrastructure to figure out which component the browser is trying to talk to.  The one for the Template was the python string that could be used to refer to the object.  This is the one used in the .cfg file and is consistent with all the other ```uid``` values in the rest of the system.  In the new code, all ```uid```s in the Python code are always python strings for referring to things, and in the places where we need ```id(self)``` we just call ```id(self)```.

### DIFF
--- a/nengo_gui/components/__init__.py
+++ b/nengo_gui/components/__init__.py
@@ -1,9 +1,21 @@
-from .slider import Slider, SliderTemplate
-from .value import Value, ValueTemplate
-from .xyvalue import XYValue, XYValueTemplate
-from .sim_control import SimControl, SimControlTemplate
-from .raster import Raster, RasterTemplate
-from .voltage import Voltage, VoltageTemplate
-from .pointer import Pointer, PointerTemplate
-from .netgraph import NetGraph, NetGraphTemplate
-from .ace_editor import AceEditor, AceEditorTemplate
+import inspect
+import sys
+
+from .component import Component
+from .slider import Slider
+from .value import Value
+from .xyvalue import XYValue
+from .sim_control import SimControl
+from .raster import Raster
+from .voltage import Voltage
+from .pointer import Pointer
+from .netgraph import NetGraph
+from .ace_editor import AceEditor
+
+# Old versions of the .cfg files used Templates which had slightly different
+# names than the Components currently usef.  This code is needed to
+# successfully parse those old .cfg files
+this_module = sys.modules[__name__]
+for name, obj in inspect.getmembers(this_module, inspect.isclass):
+    if issubclass(obj, Component):
+        setattr(this_module, name + 'Template', obj)
--- a/nengo_gui/components/__init__.py
+++ b/nengo_gui/components/__init__.py
@@ -1,6 +1,3 @@
-import inspect
-import sys
-
 from .component import Component
 from .slider import Slider
 from .value import Value
@@ -13,9 +10,14 @@ from .netgraph import NetGraph
 from .ace_editor import AceEditor
 
 # Old versions of the .cfg files used Templates which had slightly different
-# names than the Components currently usef.  This code is needed to
+# names than the Components currently use.  This code allows us to
 # successfully parse those old .cfg files
-this_module = sys.modules[__name__]
-for name, obj in inspect.getmembers(this_module, inspect.isclass):
-    if issubclass(obj, Component):
-        setattr(this_module, name + 'Template', obj)
+SliderTemplate = Slider
+ValueTemplate = Value
+XYValueTemplate = XYValue
+SimControlTemplate = SimControl
+RasterTemplate = Raster
+VoltageTemplate = Voltage
+PointerTemplate = Pointer
+NetGraphTemplate = NetGraph
+AceEditorTemplate = AceEditor
--- a/nengo_gui/components/component.py
+++ b/nengo_gui/components/component.py
@@ -18,7 +18,7 @@ class Component(object):
 
     # The parameters that will be stored in the .cfg file for this Component
     # Subclasses should override this as needed.
-    config_params = dict(x=0, y=0, width=100, height=100, label_visible=True)
+    config_defaults = dict(x=0, y=0, width=100, height=100, label_visible=True)
 
     def __init__(self, component_order=0):
         # when generating Javascript for all the Components in a Page, they
--- a/nengo_gui/components/component.py
+++ b/nengo_gui/components/component.py
@@ -30,7 +30,7 @@ class Component(object):
         # to swap out the old Component with the new one.
         self.replace_with = None
 
-    def initialize(self, page, config, uid):
+    def attach(self, page, config, uid):
         """Connect the Component to a Page."""
         self.config = config  # the nengo.Config[component] for this component
         self.page = page      # the Page this component is in
--- a/nengo_gui/components/component.py
+++ b/nengo_gui/components/component.py
@@ -1,41 +1,111 @@
 import json
 
 class Component(object):
-    default_params = dict(x=0, y=0, width=100, height=100, label_visible=True)
+    """Abstract handler for a particular Component of the user interface.
+
+    Each part of the user interface has part of the code on the server-side
+    (in Python) and a part on the client-side (in Javascript).  These two sides
+    communicate via WebSockets, and the server-side is always a subclass of
+    Component.
+
+    Each Component can be configured via the nengo.Config system.  Components
+    can add required nengo objects into the model to allow them to gather
+    the required data from the running model.  Communication from server to 
+    client is done via Component.update_client(), which is called regularly
+    by the Server.ws_viz_component handler.  Communication from client to
+    server is via Component.message().
+    """
+
+    # The parameters that will be stored in the .cfg file for this Component
+    # Subclasses should override this as needed.
+    config_params = dict(x=0, y=0, width=100, height=100, label_visible=True)
 
     def __init__(self, component_order=0):
-        # the order this component will be defined in the javascript
+        # when generating Javascript for all the Components in a Page, they
+        # will be sorted by component_order.  This way some Components can
+        # be defined before others.
         self.component_order = component_order
+
+        # If we have reloaded the model (while typing in the editor), we need
+        # to swap out the old Component with the new one.
         self.replace_with = None
 
     def initialize(self, page, config, uid):
-        self.config = config
-        self.page = page
-        self.uid = uid
+        """Connect the Component to a Page."""
+        self.config = config  # the nengo.Config[component] for this component
+        self.page = page      # the Page this component is in
+        self.uid = uid        # The Python string referencing this component
 
     def update_client(self, client):
+        """Send any required information to the client.
+
+        This method is called regularly by Server.ws_viz_component().  You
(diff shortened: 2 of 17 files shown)


tcstewar:
Another note:  While fixing how this system interacts with the undo system, I accidentally discovered that boolean parameters on Components were being saved as 1 or 0 in the .cfg file.  Fixing this solves #307

tcstewar:
And this also fixes #431

nengo_gui/components/netgraph.py
```diff
@@ -453,7 +469,4 @@ def create_connection(self, client, conn, parent):
         info = dict(uid=uid, pre=pres, post=posts, type='conn', parent=parent)
         client.write(json.dumps(info))
 
-
-class NetGraphTemplate(Template):
-    cls = NetGraph
-    config_params = dict()
+NetGraphTemplate = NetGraph
```
Seanny123: Although you removed the Template class, you still create a reference to the object named as the Template. Why is that?

jgosmann: Backwards compatibility?

tcstewar: Yes.  I'll add a comment to make that clear (and perhaps find a way to do it in one place, rather than in each Component file).

tcstewar:
> If this is going to be one of the final design-level iterations, it would be nice to have some comments/docstrings to specify the design-scope and purpose of the Component.

I've added comments to component.py and value.py.  Let me know if those help!

nengo_gui/components/netgraph.py
tcstewar: > Although you removed the Template class, you still create a reference to the object named as the Template. Why is that?

I've moved all that code to ```components/__init__.py``` and added a comment describing why it's there.  :)

nengo_gui/static/netgraph.js
```diff
@@ -335,7 +335,7 @@ Nengo.NetGraph.prototype.on_message = function(event) {
         var uid = data.uid;
         for (var i = 0; i < Nengo.Component.components.length; i++) {
             if (Nengo.Component.components[i].uid === uid) {
-                Nengo.Component.components[i].remove(true);
+                Nengo.Component.components[i].remove(true, data.report_back);
```
Seanny123: `report_back` seems like a bit of an ambiguous variable name. Is the only use of this variable to check whether the item was being undone or deleted? Is it associated to a particular event and is it likely that it will always be associated to that event?

Seanny123:
Other than my last two comments, this pull request LGTM.

nengo_gui/static/netgraph.js
tcstewar: Hmm..  the intent is that it's a flag to indicate whether the NetGraph should report back to the server that the removal happened.  We don't want it to do so in this case, as the server is the one who asked for it to be removed in the first place.  Not sure what a better name would be.

Seanny123: How about `server_removed`?

tcstewar: Yuck...   I usually prefer function parameters that indicate what the function is supposed to do, rather than function parameters that indicate the context in which someone else called the function.  If ```report_back``` isn't clear enough, I'd be more tempted to do something like ```indicate_client_should_report_removal_back_to_server```, but maybe not quite as silly as that.  Can you say more about why you find ```report_back``` to be ambiguous?  I'm having trouble seeing what's unclear about it.

Seanny123: Lol, Terry you have the most visceral reactions to bad naming suggestions.

Now that you've explained it to me, as well as your methodology for naming parameters, it makes sense to me. If it's not too much trouble adding a quick comment might be a good idea, especially considering how deep it is in the if-statements.

tcstewar: :)  Gah, rereading my comment it does come off as a bit... visceral.  Sorry about that.  

And now that I'm rereading it, I wonder if part of it is that both "report" and "back" have many meanings, so it's unclear without a comment even how to parse this (is "report" a verb or a noun? Is "back" a noun or a verb particle?)  So it's way more potentially confusing than I was initially believing.

So how about ```notify_server```?   That's a little less ambiguous to parse....  :)

Seanny123: That sounds delightful. (:

tcstewar: Changed!  :)  Thank you for pushing on this....

nengo_gui/components/__init__.py
```diff
@@ -1,9 +1,21 @@
...
+from .raster import Raster
+from .voltage import Voltage
+from .pointer import Pointer
+from .netgraph import NetGraph
+from .ace_editor import AceEditor
+
+# Old versions of the .cfg files used Templates which had slightly different
+# names than the Components currently usef.  This code is needed to
+# successfully parse those old .cfg files
+this_module = sys.modules[__name__]
```
tbekolay: I know this saves us a couple of lines, but this seems like a weird place to put in the mental effort to read/write this Python magic and realize what's going on. If I'm reading this all correctly, you could replace it with
```python
SliderTemplate = Slider
ValueTemplate = Value
XYValueTemplate = XYValue
# and so on
```
which is 9 lines versus these 4, but it definitely takes less time to read and understand the 9 line version. There's also no reason to bother doing this for new components, so it doesn't matter if it gets out of date. Ideally, we'll remove the 9 lines eventually anyway.

nengo_gui/components/component.py
```diff
@@ -1,49 +1,111 @@
...
+    can add required nengo objects into the model to allow them to gather
+    the required data from the running model.  Communication from server to 
+    client is done via Component.update_client(), which is called regularly
+    by the Server.ws_viz_component handler.  Communication from client to
+    server is via Component.message().
+    """
+
+    # The parameters that will be stored in the .cfg file for this Component
+    # Subclasses should override this as needed.
+    config_params = dict(x=0, y=0, width=100, height=100, label_visible=True)
```
tbekolay: I'm curious why this gets stored in the class rather than in instances. I can see why you might store the keys, as those are more associated with the class not the instance, but why the values? Or are these values defaults? If so, it might make more sense to rename this to something that includes 'default'... `param_defaults` or some such?

nengo_gui/components/__init__.py
tcstewar: That is a really good point.  :)  I've updated it to use your much clearer approach.

nengo_gui/components/component.py
```diff
@@ -1,49 +1,111 @@
...
+        # when generating Javascript for all the Components in a Page, they
+        # will be sorted by component_order.  This way some Components can
+        # be defined before others.
         self.component_order = component_order
+
+        # If we have reloaded the model (while typing in the editor), we need
+        # to swap out the old Component with the new one.
         self.replace_with = None
 
+    def initialize(self, page, config, uid):
```
tbekolay: `initialize` as a method name always make me cringe, since it seems like it should be done in `__init__`. There are times when you need to defer parts of what could be called initialization, but regardless I would not call the method `initialize`. If the docstring here is correct, then how about `connect`?

nengo_gui/components/component.py
tcstewar: Yes, those are the defaults.  So this is meant to indicate both the keys and the default values for those keys.

I could see ```param_defaults``` working....  what about ```config_defaults```?  I'd like to somehow have the name indicate it's about the config system....

nengo_gui/config.py
```diff
@@ -43,10 +43,14 @@ def dumps(self, uids):
...
-            elif isinstance(obj, nengo_gui.components.component.Template):
+            elif isinstance(obj, nengo_gui.components.component.Component):
                 lines.append('%s = %s' % (uid, obj.code_python(uids)))
                 for k in obj.config_params.keys():
                     v = getattr(self[obj], k)
-                    lines.append('_viz_config[%s].%s = %g' % (uid, k, v))
+                    if isinstance(v, bool):
+                        val = '%s' % v
+                    else:
+                        val = '%g' % v
```
tbekolay: Definitely a nitpick, but you could use an inline `if/else` here:
```python
val = '%s' % v if isinstance(v, bool) else '%g' % v
```

nengo_gui/components/component.py
tbekolay: What do you mean by "it's about the config system"? I thought these were the arguments you're passing into a particular Component subclass constructor, is that not correct?

nengo_gui/components/component.py
tcstewar: > initialize as a method name always make me cringe, since it seems like it should be done in __init__.

:)  fair enough.

I could see ```connect``` working, although that might be a bit ambiguous with the actual socket connections and nengo connections going on.  Perhaps ```attach```?  or ```set_page```?

tbekolay: I like `attach`, but I don't know about `set_page` since you're setting more than the page right? If you were just setting the page, I'd recommend making `page` a property, but it's more than that.

tbekolay:
Aside from my inline nitpicks, this looks great to me! Should be much clearer for those implementing new components, which hopefully will be several people ;)

nengo_gui/components/component.py
tcstewar: No, those are the defaults for the nengo.Config system that's storing all the stuff in the .cfg file.  They never get passed to any constructors anywhere.

nengo_gui/config.py
tcstewar: Hmm...  or I could do

```python
val = ('%s' if isinstance(v, bool) else '%g') % v
```

but that might not be all that readable either....   I've generally tried to only use the inline if/else in situtations where there's only one thing going on....  this is a borderline case where I'm not only doing the if/else but also the string reformatting, so I think it's slightly clearer separated like this.

nengo_gui/components/component.py
tbekolay: Ah, I just read `config.py`. `config_defaults` sounds good :)

tcstewar: Fixed!  :)

nengo_gui/components/component.py
tcstewar: I just switched it over to ```attach```.  :)

tcstewar:
I think I've responded to all the suggestions here..  anyone mind if I merge this?  :)

Seanny123:
Go for it!

PR #494, opened 2015-07-17
Multiple slider jumping bug, described in #493

### LINKED ISSUE #493: Slider reset bug - they revert to their old values
I've tried this with a slider with 2 values but the same problem should appear with multiple knobs as well: select one of the sliders and change it's values, pause the simulation, right click and reset the values of all the sliders, then change the other slider (**not** the origin one)  and the original one will revert to its old value (the one right before the reset).

I've just tried it with 3 values and if I change 2 of the values, pause + reset and change the 3rd value the other 2 revert to their old values.

Quick example so you can test it out yourself rapidly.
```python
import nengo

model = nengo.Network(label="NetworkName")

with model:
    some_input = nengo.Node(output=lambda t: [0, 0, 0])
    A = nengo.Ensemble(100, 1)
    nengo.Connection(some_input, A, transform=[[1, 1, 1]])
```

cchan987:
Fixes #493

### DIFF
--- a/nengo_gui/static/slider.js
+++ b/nengo_gui/static/slider.js
@@ -272,8 +272,9 @@ Nengo.Slider.prototype.user_value = function () {
 Nengo.Slider.prototype.user_reset_value = function() {
     for (var i = 0; i < this.sliders.length; i++){
         this.notify('' + i + ',reset');
+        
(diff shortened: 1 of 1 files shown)


Seanny123:
LGTM.

Issue #470, opened 2015-07-07
Zooming with mouse pointer over plot or slider has wrong behaviour


jgosmann:
The zooming isn't centered on the mouse pointer in this case.

cchan987:
odd i'm not getting the behaviour from yesterday. 
Was anyone able to replicate this on a different computer from Jan's?

jgosmann:
Yes, I have the same behaviour on my desktop.

tcstewar:
> Was anyone able to replicate this on a different computer from Jan's?

It's also replicated for me.  That said, it's a much more subtle effect than we were seeing on Jan's computer -- maybe off by a few pixels, but it's definitely there and only there when the mouse is over a graph/slider.

cchan987:
The root of this issue is the use of layerX and layerY on line 153 of netgraph.js
This property has different meanings if it is triggered inside or outside of an element.
https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/layerX

Switched to Clientx and ClientY as they have appropriate behaviour inside and outside components

tcstewar:
This should be fixed now by #496

PR #496, opened 2015-07-20
switched to clientx from layerx

cchan987:
addresses #470

### DIFF
--- a/nengo_gui/static/netgraph.js
+++ b/nengo_gui/static/netgraph.js
@@ -149,9 +149,10 @@ Nengo.NetGraph = function(parent, args) {
             event.preventDefault();
 
             self.menu.hide_any();
+            var tool_h = $(toolbar.toolbar).height();
+            var x = (event.clientX) / $(self.svg).width()
+            var y = (event.clientY - tool_h) / $(self.svg).height();
 
-            var x = (event.layerX / $(self.svg).width())
-            var y = (event.layerY / $(self.svg).height());
 
             switch (event.deltaMode) {
                 case 1:  // DOM_DELTA_LINE


nengo_gui/static/netgraph.js
```diff
@@ -150,8 +150,10 @@ Nengo.NetGraph = function(parent, args) {
 
             self.menu.hide_any();
 
-            var x = (event.layerX / $(self.svg).width())
-            var y = (event.layerY / $(self.svg).height());
+            var x = (event.clientX / $(self.svg).width())
+            var y = (event.clientY / $(self.svg).height());
+
+            console.log(event)
```
jgosmann: debug code to remove

nengo_gui/static/netgraph.js
```diff
@@ -179,7 +181,7 @@ Nengo.NetGraph = function(parent, args) {
                 scale = 1. / scale;
             }
 
-            Nengo.Component.save_components();
+            //Nengo.Component.save_components();
```
jgosmann: Why is this commented out?

jgosmann:
This works now much better. I noticed that the zooming anywhere (not just over plots) is not centered on the mouse pointer along the vertical axis. Probably the height of the toolbar isn't respected in the calculations. Though, you might prefer to open a separate issue for that.

Another note: Apparently we have to places where the zooming is implemented (netgraph items and plots)? It might be worthwhile to do some refactoring to remove that code duplication. But that is probably another PR, too.

cchan987:
updated

tcstewar:
Hmm, this gives very wrong zooming for me (Chrome on Windows).  

For example, if I just open up default.py and put my mouse cursor inside on of the circles in the outer ring of the Ensemble and try scrolling, my mouse cursor fairly quickly is outside the circle.  So that means it's not zooming based on where the mouse cursor is located.  A similar effect seems to be happening for Value graphs and other Components.

cchan987:
> this gives very wrong zooming for me (Chrome on Windows).
This looks to be an issue on master as well actually.

tcstewar:
> This looks to be an issue on master as well actually. Perhaps it should be a seperate PR?

Chrome on Windows in master is working fine for me....

jgosmann:
> This looks to be an issue on master as well actually. Perhaps it should be a seperate PR?
> 
> Chrome on Windows in master is working fine for me....

Firefox on OS X in master is working fine for me too.

cchan987:
Ah, It seems to be a discrepancy in the mouse Y location caused by the toolbar. hunting for fix...

cchan987:
Updated, this seems to be working for me now.

nengo_gui/static/top_toolbar.js
```diff
@@ -53,6 +53,8 @@ Nengo.Toolbar = function(filename) {
 
     this.toolbar = $('#toolbar_object')[0];
 
+    $(this.toolbar).css('height', '2.5em')
```
jgosmann: This should not be necessary, I think.

tcstewar:
This is now looking good to me.  I'm going to squash the commits together before merging though (mostly as practice because I've never done that before).

PR #497, opened 2015-07-21
less harsh red

### LINKED ISSUE #492: More subtle indication that nengo_gui isn't running
As discussed in #480 (after it had been merged), the bright red indication that nengo_gui has stopped running is perhaps a bit much.   I've suggested just not changing the background, but @jgosmann pointed out that if you do happen to close the modal then there's no indication that it has stopped.

Three options come to mind:

 - a more subtle color change
 - disable closing the modal
 - do something like what Jupyter does: snapshot2

Thoughts/suggestions?

cchan987:
Fix #492


tcstewar:
Looks good to me

PR #502, opened 2015-07-23
Y axis labels for value plots

### LINKED ISSUE #261: Axis values not showing up on value plots properly
Depending on what you set your range to, the numbers along the plot's vertical axis won't show up. For example, if you set the range on a value plot to "-2,2" these numbers will show up, but if you set it to "-3,3" then no numbers on the ends will show up.

cchan987:
Instead of letting d3 automatically generate ticks as it want, Enforce that it labels the min and max values of the value plot range.

Fix #261

### DIFF
--- a/nengo_gui/static/2d_axes.js
+++ b/nengo_gui/static/2d_axes.js
@@ -33,8 +33,6 @@ Nengo.Axes2D = function(parent, args) {
         .attr("class", "axis axis_x unselectable")
         .call(this.axis_x);
 
-
-
     /** define the y-axis */
     this.axis_y = d3.svg.axis()
         .scale(this.scale_y)
--- a/nengo_gui/static/2d_axes.js
+++ b/nengo_gui/static/2d_axes.js
@@ -23,7 +23,7 @@ Nengo.Axes2D = function(parent, args) {
 
     /** spacing between the graph and the outside edges (in pixels) */
     this.set_axes_geometry(args.width, args.height);
-    console.log()
+    
     /** define the x-axis */
     this.axis_x = d3.svg.axis()
         .scale(this.scale_x)
@@ -37,10 +37,13 @@ Nengo.Axes2D = function(parent, args) {
     this.axis_y = d3.svg.axis()
         .scale(this.scale_y)
         .orient("left")
-        .ticks(0)
+        .ticks(2)
+        .tickValues([-1,0,1])
+
     this.axis_y_g = this.svg.append("g")
         .attr("class", "axis axis_y unselectable")
         .call(this.axis_y);
+
 };
 
 Nengo.Axes2D.prototype.set_axes_geometry = function(width, height) {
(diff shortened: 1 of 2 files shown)


nengo_gui/static/2d_axes.js
```diff
@@ -38,9 +38,12 @@ Nengo.Axes2D = function(parent, args) {
         .scale(this.scale_y)
         .orient("left")
         .ticks(2)
+        .tickValues([-1,0,1])
```
tcstewar: Hmm, shouldn't this be ```[args.min_value, args.max_value]```?  And isn't the ```ticks(2)``` redundant?

nengo_gui/static/value.js
```diff
@@ -148,6 +150,7 @@ Nengo.Value.prototype.set_range = function() {
             var max = parseFloat(new_range[1]);
             self.update_range(min, max);
             self.save_layout();
+            self.axes2d.axis_y.tickValues([min,max])
```
tcstewar: spacing after the comma....  :)

cchan987:
updated

tcstewar:
Hmm...  when the graphs render for me (Windows, Chrome and Firefox), I get the "minus" sign disappearing from the bottom one (due to it rendering as "-1.0" rather than "-1".  This means that the graphs look like they go from "1.0" to "1.0", which is a bit of a problem.  So this change is making the bug in #450 much worse (since it now happens for pretty much all graphs, rather than just ones where people have changed the range).

Not sure what the right solution is, other than to modify this one somehow so that it would end up displaying the axis labels as "1" and "-1" rather than "1.0" and "-1.0"....

cchan987:
I'll probably merge this and #450 into one PR. Feels like they're related enough that we can get 2 birds with 1 stone here.

cchan987:
This PR also fixes #450 now

tcstewar:
Nice!  That looks like a pretty good fix to me....

nengo_gui/static/2d_axes.js
```diff
@@ -85,3 +87,21 @@ Nengo.Axes2D.prototype.on_resize = function(width, height) {
...
 };
 
+Nengo.Axes2D.prototype.fit_ticks = function(parent){
+    var self = this;
+    setTimeout(function(){
+        var ticks = $(parent.div).find('.tick');
+        var max_w = 0;
+        for (var i = 0; i < ticks.length; i++) {
+            var w = ticks[i].getBoundingClientRect().width
+            console.log(w)
```
tcstewar: This log should probably be removed

nengo_gui/static/2d_axes.js
```diff
@@ -85,3 +87,21 @@ Nengo.Axes2D.prototype.on_resize = function(width, height) {
     this.axis_y_g.call(this.axis_y);
 };
 
+Nengo.Axes2D.prototype.fit_ticks = function(parent){
+    var self = this;
+    setTimeout(function(){
+        var ticks = $(parent.div).find('.tick');
+        var max_w = 0;
+        for (var i = 0; i < ticks.length; i++) {
+            var w = ticks[i].getBoundingClientRect().width
```
tcstewar: semicolon missing....

tcstewar:
Seems okay to me.  :)  merging...

Issue #499, opened 2015-07-23
Nengo GUI visual bug


pabogdan:
Screenshot 1

Starting nengo gui this morning I get this weird visual bug. The code is inconsequential, I've tried with empty files and simpler examples as well. The browser used is Chrome and the setup has remained the same for weeks now. This first image shows the problem: no simulation bar at the bottom.
![Screenshot 1](http://i.imgur.com/08gdgFb.png)

Screenshot 2

In my attempt to understand what was happening I tried zooming in and out, which didn't have the desired effect. Instead, scrolling down revealed the missing bar, but with missing timeline and simulation start button.
![Screenshot 2](http://i.imgur.com/mfMp19U.png)

Screenshot 3

Workaround 1: Putting the browser in fullscreen mode
![Screenshot 3](http://i.imgur.com/tom0EaK.png)

Screenshot 4

Workaround 2: Closing the text editor
![Screenshot 4](http://i.imgur.com/bDVnzpJ.png)

I want to just point out that simulations still run, you just can't select the amount of simulation time you want to see and you have to start / pause it by pressing space.

pabogdan:
Have just tried it in Firefox and everything works as expected. Must be an issue with the latest Chrome update.

tcstewar:
> Have just tried it in Firefox and everything works as expected. Must be an issue with the latest Chrome update.

Argh...  yup, I just confirmed this on a Windows 8.1 machine and on a Linux Mint machine...  The problem occurs with Chrome 44.  Although on Mint going to fullscreen doesn't work as a workaround.

Thank you for finding this!  We'll sort out what the problem is -- at least it's very replicable!

tcstewar:
Looks like this is a bug in Chrome that is breaking a lot of websites for people:

http://code.google.com/p/chromium/issues/detail?id=506893

tcstewar:
I think I've got a fix for it...  if you could try out the ```fix-editor-scrolling``` branch and see if that works for you, that'd be great!

pabogdan:
Yes, it does seem to fix it. Good job!

PR #503, opened 2015-07-24
Fix editor scrolling problem (#499)

tcstewar:
I believe this will fix the scrolling problem encountered in Chrome 44 (See issue #499)

If people could test it out on their browsers, that would be great!

### DIFF
--- a/nengo_gui/static/ace.css
+++ b/nengo_gui/static/ace.css
@@ -10,7 +10,6 @@
 
 #editor {
     border-left:5px solid #606060;
-    height: 100%;
     flex: 1 1 auto;
 }
 
--- a/nengo_gui/static/ace.js
+++ b/nengo_gui/static/ace.js
@@ -168,7 +168,7 @@ Nengo.Ace.prototype.on_message = function (event) {
 
 Nengo.Ace.prototype.show_editor = function () {
     var editor = document.getElementById('rightpane');
-    editor.style.display = 'block';
+    editor.style.display = 'flex';
     this.hidden = false;
     if (!this.console_hidden) {
         this.console.style.display = 'block';


jgosmann:
Assigned this to myself because I want to give this a test before it's merged (but might not be able to test it right away as I am on a limited WIFI connection right now). I have a feeling this change might break other things related to some layout changes I introduced.

jgosmann:
Ok, tested it out and doesn't seem to break anything and fixes the problem in Chrome 44 on OS X for me.

tcstewar:
> Ok, tested it out and doesn't seem to break anything and fixes the problem in Chrome 44 on OS X for me.

Yay!  I was hoping you'd take a look as you've got more experience with the flex layout stuff than anyone else.  :)

tcstewar:
Argh, I just tested this under Chrome 43 and it works great....  except that if you toggle the editor window off and then on the editor window does not reappear.  :(

tcstewar:
And it's also the case that on Firefox this PR means that if you toggle the editor off and then on again, it doesn't reappear.  :( :(  Any suggestions for fixing that?  I'll start trying stuff at random again...  :)

jgosmann:
Interesting, I didn't expect that behavior ...

jgosmann:
I think, the issue is simply that the `display` property of `#rightpane` gets set to `block` instead of flex if the editor is toggled.

tcstewar:
> I think, the issue is simply that the display property of #rightpane gets set to block instead of flex if the editor is toggled.

Good catch!  Yup, that seems to fix it.  :)  Working fine on Firefox and Chrome 43 right now (the two browsers I have access to at the moment).  Does that fix it for you?

jgosmann:
Now the console doesn't appear anymore in case of a syntax error.

tcstewar:
> Now the console doesn't appear anymore in case of a syntax error.

???  Weird...  it does for me in Firefox and Chrome 43....  grr....

jgosmann:
Just wanted to look into this, but now it does work for me? :confused: 
No idea what was going on ...

tcstewar:
> Just wanted to look into this, but now it does work for me?  
> No idea what was going on ..

Curiouser and curiouser.....  I'll try it on a bunch of browsers at the office tomorrow and if it still seems to work we'll merge and deal with the consequences when they come up...  :)

tcstewar:
I've had enough people indicate that it works for them that I'm going to go ahead and merge this.  :)

PR #508, opened 2015-07-28
Handling of new-style LearningRule Connections

tcstewar:
The gui currently crashes if you try to use the new syntax for learning rules.  This quick change just updates things so that it handles them exactly like it did the old syntax.  In some future PR we probably would want a visual indication of the learning rule, but this at least lets people work with learning rules and the new syntax.

### DIFF
--- a/nengo_gui/components/netgraph.py
+++ b/nengo_gui/components/netgraph.py
@@ -317,10 +317,14 @@ class NetGraph(Component):
             new_post = new_item.post_obj
             if isinstance(old_pre, nengo.ensemble.Neurons):
                 old_pre = old_pre.ensemble
+            if isinstance(old_post, nengo.connection.LearningRule):
+                old_post = old_post.connection.post_obj
             if isinstance(old_post, nengo.ensemble.Neurons):
                 old_post = old_post.ensemble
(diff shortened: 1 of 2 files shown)


Seanny123:
When I brought this up to date with master, it worked well with [Aaron's Voja code](https://gist.github.com/Seanny123/fa5fbcf91e2f540129a1), so I think it's ready to merge until we figure out a better way to display these types of things.

Speaking of displaying learning rules, I seem to remember talking about this during a meeting and coming to some sort of conclusion. Is there a copy of those conclusions somewhere?

tcstewar:
> When I brought this up to date with master, it worked well with Aaron's Voja code, so I think it's ready to merge until we figure out a better way to display these types of things.

Thanks!

> Speaking of displaying learning rules, I seem to remember talking about this during a meeting and coming to some sort of conclusion. Is there a copy of those conclusions somewhere?

I don't think there was a solid conclusion other than this sort of thing being what eventual visual customization hooks would need to support.

tcstewar:
Just added the same logic in the layout system.  @Seanny123 seems happy, so I'll merge now.

PR #510, opened 2015-07-30
fix menu off screen

### LINKED ISSUE #373: Menu is hidden when right-clicking at the bottom of the window
As reported by @jgosmann in #342. Looks like this:

screen shot 2015-06-09 at 2 58 43 pm

cchan987:
This solution moves the menu to keep it on screen in a way that I hope is non-offensive to people. :P

Fixes #373

### DIFF
--- a/nengo_gui/static/menu.js
+++ b/nengo_gui/static/menu.js
@@ -27,6 +27,8 @@ Nengo.Menu.prototype.show = function (x, y, items) {
         return;
     }
 
+    this.menu_margin = 15;
+
     // TODO: move this to the constructor
     this.menu_div = document.createElement('div');
     this.menu_div.style.position = 'fixed';
@@ -68,6 +70,7 @@ Nengo.Menu.prototype.show = function (x, y, items) {
         this.menu.appendChild(b);
     }
     this.visible = true;
+    this.check_overflow(x, y);
     Nengo.Menu.visible_menus[this.div] = this;
 };
 
@@ -100,5 +103,22 @@ Nengo.Menu.prototype.visible_any = function () {
 Nengo.Menu.prototype.hide_any = function () {
     for(var k in Nengo.Menu.visible_menus) {
         Nengo.Menu.hide_menu_in(k);
+    }
+}
+
+Nengo.Menu.prototype.check_overflow = function (x, y) {
+    var corrected_y = y - $(toolbar.toolbar).height();
+    var h = $(this.menu).height();
+    var w = $(this.menu).width();
+    
+    var main_h = $('#main').height()
+    var main_w = $('#main').width()
+
+    if (corrected_y + h  + this.menu_margin > main_h) {
+        this.menu_div.style.top = y - h - this.menu_margin;
+    }
+
+    if(x + w + this.menu_margin > main_w) {
+        this.menu_div.style.left = x - w;
     }
 }
--- a/nengo_gui/static/menu.js
+++ b/nengo_gui/static/menu.js
@@ -27,7 +27,7 @@ Nengo.Menu.prototype.show = function (x, y, items) {
         return;
     }
 
-    this.menu_margin = 15;
+    this.menu_margin = 12;
 
     // TODO: move this to the constructor
     this.menu_div = document.createElement('div');
@@ -118,7 +118,7 @@ Nengo.Menu.prototype.check_overflow = function (x, y) {
         this.menu_div.style.top = y - h - this.menu_margin;
     }
 
-    if(x + w + this.menu_margin > main_w) {
-        this.menu_div.style.left = x - w;
+    if(x + w  > main_w) {
+        this.menu_div.style.left = main_w - w;
     }
 }
--- a/nengo_gui/static/menu.js
+++ b/nengo_gui/static/menu.js
@@ -106,7 +106,6 @@ Nengo.Menu.prototype.hide_any = function () {
 
 Nengo.Menu.prototype.check_overflow = function (x, y) {
     var menu_margin = 0;
-    console.log(this.menu)
     var corrected_y = y - $(toolbar.toolbar).height();
     var h = $(this.menu).height();
     var w = $(this.menu).width();
--- a/nengo_gui/static/menu.js
+++ b/nengo_gui/static/menu.js
@@ -27,8 +27,6 @@ Nengo.Menu.prototype.show = function (x, y, items) {
         return;
     }
 
-    this.menu_margin = 12;
-
     // TODO: move this to the constructor
     this.menu_div = document.createElement('div');
     this.menu_div.style.position = 'fixed';
@@ -107,6 +105,8 @@ Nengo.Menu.prototype.hide_any = function () {
 }
 
 Nengo.Menu.prototype.check_overflow = function (x, y) {
+    var menu_margin = 0;
+    console.log(this.menu)
     var corrected_y = y - $(toolbar.toolbar).height();
     var h = $(this.menu).height();
     var w = $(this.menu).width();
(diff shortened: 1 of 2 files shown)


nengo_gui/static/menu.js
```diff
@@ -27,6 +27,8 @@ Nengo.Menu.prototype.show = function (x, y, items) {
         return;
     }
 
+    this.menu_margin = 15;
+
```
tcstewar: What's this for?

tcstewar: (Sorry, to be more clear in my comment, why is there a ```menu_margin``` at all?  Do other browsers do that?  I don't see that effect in either Chrome or Firefox on my computer....)

cchan987: > (Sorry, to be more clear in my comment, why is there a menu_margin at all? Do other browsers do that? I don't see that effect in either Chrome or Firefox on my computer....)

I have this effect on my browser. It is more of a buffer so that the menu doesn't press right up against the edge of the window even if it would have fitted in.

tcstewar:
This is definitely an improvement over it going off the screen, but I still think we should be using the same algorithm that the majority of browsers are using.

Right now, the algorithm is:

```python
if x+width > screen_width: 
    x -= width
if y+height > screen_height: 
    y -= height
```

From looking at the other browsers, I'd thought the majority were doing either

```python
if x+width > screen_width: 
    x = screen_width - width
if y+height > screen_height: 
    y -= height
```

or

```python
if x+width > screen_width: 
    x -= width
if y+height > screen_height: 
    y = screen_height - height
```

I believe it was only Firefox on OS X that did the last option, and most of the rest did the second option.  Am I misremembering that?  I really should have started making a list....

nengo_gui/static/menu.js
```diff
@@ -102,3 +105,20 @@ Nengo.Menu.prototype.hide_any = function () {
...
+Nengo.Menu.prototype.check_overflow = function (x, y) {
+    var corrected_y = y - $(toolbar.toolbar).height();
+    var h = $(this.menu).height();
+    var w = $(this.menu).width();
+    
+    var main_h = $('#main').height()
+    var main_w = $('#main').width()
+
+    if (corrected_y + h  + this.menu_margin > main_h) {
+        this.menu_div.style.top = y - h - this.menu_margin;
```
cchan987: Saying the above, menu margin is misused here

nengo_gui/static/menu.js
tcstewar: What browser is that?  On Chrome and Firefox for me it goes right to the edge.  Is it the same for you in both Chrome and Firefox?

tcstewar: (basically, I'd argue that you want it to go right to the edge -- there's no reason to have a buffer there, so I'm curious as to what platforms put a buffer there and if we can figure out why)

cchan987: > What browser is that? On Chrome and Firefox for me it goes right to the edge. Is it the same for you in both Chrome and Firefox?

In chrome I have the buffer.
In firefox the menu will overflow the bottom out of the window, and go up against the edge for the right hand side

tcstewar: > In firefox the menu will overflow the bottom out of the window, and go up against the edge for the right hand side

What happens in firefox if the window is fullscreen so it can't overflow out the bottom?  (indeed, all of these checks should be in situations where the window is fullscreen so there's no option for the right-click menu to overflow)

And in any case, from what you're saying, out of the 4 browsers looked at, 3 go right up against the edge and 1 doesn't.  We should look at a few more examples to get a consensus, but the majority seems to be to go right to the edge.

cchan987: > What happens in firefox if the window is fullscreen so it can't overflow out the bottom? (indeed, all of these checks should be in situations where the window is fullscreen so there's no option for the right-click menu to overflow)

This was in the case of it being full screen, it would simply spill over onto the taskbar.

cchan987: > This was in the case of it being full screen, it would simply spill over onto the taskbar.

unfortunately when I try to take a screencap , the menu closes making it difficult to show.

tcstewar: > This was in the case of it being full screen, it would simply spill over onto the taskbar.

But is there an equivalent to this ```menu_margin``` thing?  Does it keep a buffer of a fixed amount of minimum space between the menu and the farthest edge of the screen?

cchan987: > But is there an equivalent to this menu_margin thing? Does it keep a buffer of a fixed amount of minimum space between the menu and the farthest edge of the screen?

ah good call, this does press right up against the edge of the screen. It does seem like the norm.

nengo_gui/static/menu.js
```diff
@@ -27,6 +27,8 @@ Nengo.Menu.prototype.show = function (x, y, items) {
         return;
     }
 
+    this.menu_margin = 12;
```
tcstewar: I think there are two problems here.  First, where is this magic number 12 coming from? What is it?  It's being used as a 12-pixel vertical offset inside the check_overflow code, and that is not clear from the variable naming here.  And instead of it being just a magic number, it should be computed from something so that it changes appropriately when the things that affect it change.

Second, why is this code out here rather than inside check_overflow, where it is used?  Why is it a member variable of the class, rather than just a temporary computed thing inside check_overflow?

cchan987:
Is the menu also overflowing slightly at the bottom on other computers?

cchan987:
I am wondering if the menu is overflowing at the bottom of the screen slightly for others when the margin is set to 0?

s72sue:
Yes, it does overflow (see the screen shot attached). However I didn't
change the margin. Is the default margin 0 or do I need to change it?

cchan987:
Ok thanks, The latest commit has the margin set to 0.

tcstewar:
> I am wondering if the menu is overflowing at the bottom of the screen slightly for others when the margin is set to 0?

Yes, it overflows at the bottom.  It also doesn't move up by the correct amount, so the mouse cursor isn't at the bottom of the menu.  (Windows, Chrome and Firefox)

jgosmann:
I added a commit which should fix it.

tcstewar:
> I added a commit which should fix it.

Thank you!  That works well for me.

Now I think the only thing left is the question of whether we should allow the right-click menu to overlap the editor.  It seems a bit weird to me that we bound the menu to the left side of the screen, when we could just as easily allow it to overlap the editor (assuming the editor is there, of course).

cchan987:
> I added a commit which should fix it.

Ah ha this is looking good now. Good fix!

cchan987:
> question of whether we should allow the right-click menu to overlap the editor.

If we were to allow the menu to overlap the editor I would expect that the GUI menu would also be activated when I right-click in the editor. For which we would likely want different options than the menu toggled on the left hand side of the screen.

tcstewar:
> If we were to allow the menu to overlap the editor I would expect that the GUI menu would also be activated when I right-click in the editor.

Why would you expect that?  That doesn't happen in a similar situation in any other user interface I can think of...

cchan987:
> That doesn't happen in a similar situation in any other user interface I can think of...
hmm you're right there. I withdraw my previous expectation. :)

cchan987:
> whether we should allow the right-click menu to overlap the editor.

I don't really have strong opinions on this either way. As long as the menu shows up in a somewhat logical position I'm happy.

Seanny123:
Does this need further review or is it ready to merged?

tcstewar:
> Does this need further review or is it ready to merged?

The only question for me is whether we should be using the edge of the editor as our right-most bound, but that can be left for a future PR, I suppose...

PR #511, opened 2015-07-30
Fixed axes on raster and xyvalue plots

tcstewar:
When I merged #502 I didn't notice that it broke the labelling and layout for the raster plots and the xy plots.  So this is a quick fix to make those look right again.

This is a small change and fixes a pretty big bug in master, so I'll merge it quickly and if there are discussions about a better way of doing this fix, we can do that in a separate PR.


cchan987:
This looks like a good fix to me

tcstewar:
Thanks!

Issue #512, opened 2015-08-04
Using nengo_gui in IPython crashes


tbekolay:
I just tried to run some summer school tutorials that use `IPythonViz`. The network builds fine, and GUI pops up in the cell as expected, but when I try to run the model, I get the following error a whole bunch of times (probably in a `while True` loop) before I have to stop the kernel:

```python
Traceback (most recent call last):
  File "/home/tbekolay/Code/nengo_gui/nengo_gui/server.py", line 143, in ws_viz_component
    component.update_client(client)
  File "/home/tbekolay/Code/nengo_gui/nengo_gui/components/sim_control.py", line 90, in update_client
    status = self.get_status()
  File "/home/tbekolay/Code/nengo_gui/nengo_gui/components/sim_control.py", line 104, in get_status
    if self.page.error is None:
AttributeError: 'Page' object has no attribute 'error'
```

tcstewar:
Argh!  Yup, that's broken.

I found a fix and will put in a PR asap.  Thank you!

tcstewar:
Check out #513 and see if that works for you!  :)

I really need to force myself to always do the minimal test of running a model with one of each type of graph, both on normal nengo_gui and in IPython/Jupyter......

PR #513, opened 2015-08-04
Fix broken ipython

tcstewar:
As pointed out in #512 we broke the IPython integration again.  This fixes it.


tbekolay:
This works for me! Thanks for the quick turnaround :)

tcstewar:
Cool.  This is small enough and important enough that I'm going to consider @tbekolay 's comment a review and merge this....

Issue #425, opened 2015-06-15
Update sliders when the dimensionality of a Node changes


bjkomer:
It would be nice if the sliders automatically updated to reflect the dimensionality of the Node they are controlling. Currently, if you create a 1D node, make a slider for it, and then change it to a 2D node, there is still only one thing to slide on the gui. This still runs fine, but you can only control one of the dimensions. In the reverse case with 2 sliders for a 1D Node, it also still runs fine but the second slider does nothing. Deleting and adding the slider fixes it, but it would be more convenient if this happened automatically.

tcstewar:
This should be happening now, thanks to #514

PR #514, opened 2015-08-05
Improved updating of display as components change

tcstewar:
This addresses #425 and a variety of other fixes.  Basically, if anything that's set in the constructor for a Component is changed, then we need to recreate that Component on the client side.

This fixes the following:
- if you change the dimensionality of a Node, the number of sliders will change
- if you change the dimensionality of an Ensemble or Node, the number of lines in the value graph will change
- if you change from 1 dimension 2 dimensions, the option to show an XY Value plot should appear (or disappear)
- if you add a ```_nengo_html_``` attribute to your Node, then the ```HTML``` plot option should appear.
- changing the label on an item should cause that label to change in the graphs


Seanny123:
I reviewed all the commits and tested it on my Windows 8 machine and it looks good to me. Great job Terry!

PR #516, opened 2015-08-05
Fixed value plots being behind on their data

### LINKED ISSUE #357: When a model is running fast, speed > 5x... Graphs can lag behind
Graphs do not plot their lines until the user pauses. It appears there is a lag of what is being plotted.
By increasing the complexity of the model, which causes the model to run slower, the graphs will update in real time.
Venturing a guess I would say that this is due to slow hardware.
A speed throttle would solve this issue

tcstewar:
This might finally fix #357 !   We should no longer have those value plots that lose their data when the simulation is running quickly.

The fix basically involves sending all the pending data at once, rather than sending a packet for every time step.

### DIFF
--- a/nengo_gui/components/value.py
+++ b/nengo_gui/components/value.py
@@ -59,6 +59,9 @@ class Value(Component):
         # while there is data that should be sent to the client
         while len(self.data) > 0:
             item = self.data.popleft()
+            while len(self.data) > 0:
+                item = item + self.data.popleft()
+
             # send the data to the client
             client.write(item, binary=True)
 
--- a/nengo_gui/components/value.py
+++ b/nengo_gui/components/value.py
@@ -1,5 +1,4 @@
 import struct
-import collections
 
 import nengo
 import numpy as np
@@ -21,7 +20,7 @@ class Value(Component):
         self.obj = obj
 
         # the pending data to be sent to the client
-        self.data = collections.deque()
+        self.data = []
 
         # the number of data values to send
         self.n_lines = int(obj.size_out)
@@ -56,13 +55,14 @@ class Value(Component):
         self.data.append(self.struct.pack(t, *x))
 
     def update_client(self, client):
-        # while there is data that should be sent to the client
-        while len(self.data) > 0:
-            item = self.data.popleft()
-            while len(self.data) > 0:
-                item = item + self.data.popleft()
-
-            # send the data to the client
+        length = len(self.data)
+        if length > 0:
+            # we do this slicing because self.gather_data is concurrently
+            # appending things to self.data.  This means that self.data may
+            # increase in length during this call, so we do the slicing
+            # and deletion to maintain thread safety
+            item = bytes().join(self.data[:length])
+            del self.data[:length]
             client.write(item, binary=True)
 
     def javascript(self):
--- a/nengo_gui/static/value.js
+++ b/nengo_gui/static/value.js
@@ -56,7 +56,15 @@ Nengo.Value.prototype.constructor = Nengo.Value;
  */
 Nengo.Value.prototype.on_message = function(event) {
     var data = new Float32Array(event.data);
-    this.data_store.push(data);
+    data = Array.prototype.slice.call(data);
+    var size = this.n_lines + 1;
+    while (data.length >= size) {
+        this.data_store.push(data.slice(0, size));
+        data = data.slice(size);
+    }
+    if (data.length > 0) {
+        console.log('extra data: ' + data.length);
+    }
     this.schedule_update();
 };
 


Seanny123:
Didn't see any lag while playing with the value plot from the default example on Windows/Chrome. That's what this was supposed to fix right?

nengo_gui/components/value.py
```diff
@@ -59,6 +59,9 @@ def update_client(self, client):
         # while there is data that should be sent to the client
         while len(self.data) > 0:
             item = self.data.popleft()
+            while len(self.data) > 0:
```
Seanny123: why the double-nested while-loops?

tcstewar:
> Didn't see any lag while playing with the value plot from the default example on Windows/Chrome. That's what this was supposed to fix right?

It's not supposed to fix lag -- it's supposed to fix that problem that's been around for ages that if you have a very very simple model (like one Node and one Ensemble) and hit play, the line for the value graph disappears after a few seconds.  The current workaround was adding a Node that does a time.sleep(0.001)....

nengo_gui/components/value.py
tcstewar: Hmm, I suppose that does look a little weird.  What I could do instead is get rid of the ```deque``` and just keep the concatenated list of binary data to send....

Seanny123:
Right, that bug doesn't show for me on Windows/Chrome as well.

tcstewar:
> Right, that bug doesn't show for me on Windows/Chrome as well.

It's hard to generate on Windows/Chrome, but I do find that it happens if I do Windows/Chrome/Python3.....

nengo_gui/components/value.py
jgosmann: You could also change the first `while` to an `if`. I think, that would make it already better. I'm not sure what the datatype of `item` is (just looking at the changed code excerpt), but maybe something like
```python
item = ''
while len(self.data) > 0:
    item += self.data.popleft()
```
would work?

tcstewar: Hmm, that's a lot less aggressive than the current code, in that it doesn't immediately send the data that has queued up while it's been busy putting together this list of data.  But I could do some testing to see if that less complete approach also works.

jgosmann: I can't follow you there. To me all three variants look like they (almost, see below) exactly the same thing. If this is not the case, it should be made clear why this isn't the case, so that no one comes along and modifies the code in the way I proposed.

Also, I just noticed that the double while loop might introduce a race condition. I assume data is added to `self.data` concurrently? In that case it might be possible, that data is added after the inner while loop finished, but before the condition of the outer while loop is checked. In that case the code would enter the outer while loop again, overwriting `item` and thus it loses some data.

tcstewar: > In that case it might be possible, that data is added after the inner while loop finished, but before the condition of the outer while loop is checked. In that case the code would enter the outer while loop again, overwriting item and thus it loses some data.

That's the whole point of the nested while loops.  It's supposed to go enter the outer while loop again if new data is added concurrently after the inner while loop finishes.  And if that does happen, I want ```item``` to be overwritten because it has already been sent to the client.

I think part of the problem here is that a) it looks weird to have two nested while loops like that, and b) the blank line after the inner while loop is leading people to think that's also the end of the outer while loop (i.e. it's easy to not notice that the ```client.write``` line is inside the outer while loop).  

So I'm still pretty sure that the code is algorithmically correct (and doesn't have any race conditions), but the fact that two different people found it confusing and weird definitely means I should rewrite it.  And I think something like replacing the inner loop with ```''.join(self.data[:length])``` should do it and be more efficient anyway (and be thread-safe).

tcstewar: > Hmm, that's a lot less aggressive than the current code, in that it doesn't immediately send the data that has queued up while it's been busy putting together this list of data. But I could do some testing to see if that less complete approach also works.

So I did a bit of testing on a few machines, and it looks like replacing the outer ```while``` with an ```if``` seems okay.  It definitely lags a little more, as I was suggesting, but it turns out to only be around 10 timesteps more in the worst case I could find, which is completely unnoticeable to the user.  So I'm happy getting rid of the outer while loop just to make the code look clearer.

tcstewar:
Okay, I've rewritten the ```update_client``` code in a way that's clearer and is a bit more efficient.  Here's the result:

```python
        length = len(self.data)
        if length > 0:
            # we do this slicing because self.gather_data is concurrently
            # appending things to self.data.  This means that self.data may
            # increase in length during this call, so we do the slicing
            # and deletion to maintain thread safety
            item = bytes().join(self.data[:length])
            del self.data[:length]
            client.write(item, binary=True)
```

functionally, it's almost the same as it was before, but I changed the outer loop to be an ```if``` rather than a ```while```.  This introduces a little bit of lag (because any new data appended to self.data by another thread during the execution of this call won't be sent until the next time ```update_client``` is called), but that lag turns out to be very minimal (on the order of 3 milliseconds in the worst case I could generate, which really isn't worth worrying about).  

The other big change is that I got rid of the ```deque``` and used a ```join``` to get rid of the inner while loop.  This is way more efficient than popping everything off the top of the deque.

Seanny123:
LGTM now. You know how much I love comments. (:

cchan987:
> on the order of 3 milliseconds in the worst case I could generate

I'm just curious, how you would detect a 3 millisecond lag? Using a timer?

cchan987:
This is working nicely for me.

tcstewar:
> I'm just curious, how you would detect a 3 millisecond lag? Using a timer?

No, by checking ```len(self.data)``` at the end of ```update_client```.  :)  I'm lazy.  :)

PR #518, opened 2015-08-06
More robust way to compute bounding box for graph ticks

tcstewar:
The old way returned zero a lot in Firefox (tested on 39.0, Windows).  This meant that the ticks would take up exactly zero space, making them unreadable.


Seanny123:
So the code looks fine, however when you make a pull request this simple, what do you expect of the reviewer? Do you just want us to test it on our machines or something more?

tcstewar:
> So the code looks fine, however when you make a pull request this simple, what do you expect of the reviewer? Do you just want us to test it on our machines or something more?

Good question...  hmmm...  I think the biggest thing is just testing it on your machine on a few different browsers.

cchan987:
Tested on firefox, chrome works good for me.

Issue #382, opened 2015-06-10
Plot for viewing images


hunse:
When running a model that takes images as input, we need a good way to view the input images.

cchan987:
#404 Allows for custom html. Does this suffice for a way to have images displayed?

jgosmann:
I think it still requires to write Javascript to display images as the transmitted image data has to be parsed and drawn onto a canvas. It might be possible to encode that data directly into the HTML, but I'm not sure.

The `<img>` tag won't at the moment work as the server needs to be configured to deliver that image when a certain URL is requested.

tcstewar:
> Does this suffice for a way to have images displayed?

That's not what this issue is talking about.  This issue is talking about dealing with a Node whose output value is, say, a 10x10 image, and it'd be nice to actually display that image rather than doing a Value plot with 100 lines on it.

Seanny123:
User-interface wise, would this just be another option in the context menu or is there some other way that you'd like the user to access this option?

tcstewar:
> User-interface wise, would this just be another option in the context menu or is there some other way that you'd like the user to access this option?

Hmm...  good question.  Part of me think it's be really neat for this to be an option on the Value plot itself (toggling back and forth between this mode and the normal line mode), but that's rather difficult to implement given our current framework.  So I think for the first implementation, we just have it as another option in the context menu (as it was in Nengo 1.4), and then look to merging the two in a future PR.

Seanny123:
Why is it difficult to implement in our current framework? Is it because
the SVG can't handle click events, like I seem to remember you mentioning
earlier?

tcstewar:
> Why is it difficult to implement in our current framework?

Just that the entire Value class is written assuming that there's just one basic graph it shows, so it'd basically be a rewrite of that whole thing to make it toggleable

> Is it because
the SVG can't handle click events, like I seem to remember you mentioning
earlier?

Nope, and that thing I was mentioning earlier I was wrong about.  What I was remembering is that *markers* on SVG lines can't handle click events, which is why this time when I implemented the Connection lines the triangle arrow is done as a separate SVG path, rather than as a line marker.

Seanny123:
@tcstewar didn't a version of this get included in a demo of Eric's ImageNet thing? Where is that code stored and how did you implement it in that case?

hunse:
```python
import nengo

import base64
import PIL
import cStringIO

images  # (n_images, n_channels, height, width) ndarray
[...]

PR #522, opened 2015-08-08
Updated firing pattern plot

tcstewar:
Here's the latest version of the "sparkle plot" (a.k.a. "spike grid").  This is a rebased version of #152 with a few additions I made for stability.  Most of the discussion for this PR was in #152 and this PR continues that one.

Two fixes have been added since the discussion there.  First, the time data was being stored completely wrong, so if you tried scrolling backwards in time it just didn't work at all.  This is now fixed.  Second, it now updates the plot when you change the number of neurons.


Seanny123:
If I resize this plot (chrome/linux-mint) it remains black until I hit the pause button.

tcstewar:
> If I resize this plot (chrome/linux-mint) it remains black until I hit the pause button.

Hmm...  that seems like the intended behaviour to me, as if you haven't hit play yet, then there have been no spikes, and so the plot should be all black.  What do you think it should do?

Seanny123:
I meant that if I moved or resized the plot while the simulation is running, the plot turned black until I hit the pause button. When I hit the pause button, it froze to the last spike data (as it should) and will start again normally once I press play. The undesired behaviour occurs when moving or resizing the plot.

tcstewar:
> I meant that if I moved or resized the plot while the simulation is running, the plot turned black until I hit the pause button. When I hit the pause button, it froze to the last spike data (as it should) and will start again normally once I press play. The undesired behaviour occurs when moving or resizing the plot.

This looks like it's not actually anything to do with resizing, but is to do with timing -- if you're running a very simple model and your simulation gets ahead of the display, then the display goes black (much like the value plots used to disappear).

I've fixed this in the same way as we fixed the value plots -- all the data gets sent at once.  I'm not sure how this interacts with the frame dropping, though, so I'd like to do another round of testing on different machines.

@Seanny123, does this fix it on your machine?

Seanny123:
Yep, that fixed the problem on my machine

nengo_gui/components/spike_grid.py
```diff
@@ -0,0 +1,73 @@
...
+                                         self.node, synapse=0.01)
+
+    def remove_nengo_objects(self, page):
+        page.model.connections.remove(self.conn)
+        page.model.nodes.remove(self.node)
+
+    def gather_data(self, t, x):
+        self.max_value = max(self.max_value, np.max(x))
+
+        # TODO: pass only spiking neurons, using subclass of Nengo.Image?
```
Seanny123: What Nengo.Image are you talking about?

Seanny123:
The code LGTM. After it's merged I can start working on #382

nengo_gui/components/spike_grid.py
tcstewar: Ah, that comment's a bit outdated anyway, as we don't want to just send spike data when we're doing the synapse filter in Nengo.  I removed it.

tcstewar:
> Yep, that fixed the problem on my machine

Cool, and it's worked on the few other machines I've tried, so I'll merge this and see how it goes.

PR #524, opened 2015-08-10
Choose if new browser window should be opened

pabogdan:
Allow user to specify if they want a new browser window to pop-up after starting the server. This is specified by using --no-browser (or --browser). True by default

### DIFF
--- a/nengo_gui/main.py
+++ b/nengo_gui/main.py
@@ -19,6 +19,9 @@ def main():
     parser.add_argument(
         '-b', '--backend', metavar='BACKEND',
         default='nengo', type=str, help='default backend to use')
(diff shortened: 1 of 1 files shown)


pabogdan:
Example usage:
```
nengo_gui test.py -b nengo -p "qwerty" --no-browser --debug
```

This way you can run a nengo_gui server remotely and not start a browser session on the remote server.

Seanny123:
LGTM.

tcstewar:
Yup, that seems handy to me.  Thanks!

I do note that it we probably still would also want to have an option to turn off the auto-shutdown when the last browser window disconnects, but that'd be a separate PR.

Issue #464, opened 2015-07-06
Borders of components flicker on resize


cchan987:

jgosmann:
This seems to happen primarily on slower computers (I never noticed it on my Macbook). It appears that the corner does not stay fixed under the mouse pointer. I believe (but might be wrong), that the resize amount is calculated from the relative movement of the mouse (which seems to be either inaccurate or some resize events are missed because the processing is too slow). It might be better to calculate the size based on the absolute mouse coordinates which should always be accurate (fixed aspect ratio elements do it that way). But I am not completely sure that this will elimate all flicker ...

tcstewar:
Yup, looks to me that it's about the mouse cursor being outside the component, so it's no longer triggering the hover.  But it happens on my desktop machine, so it's not about the speed of the computer for me.

I'd say the solution is just to have a "dragging" class that's set during the drag, and use that to set the look rather than the ```graph:hover``` thing.

krisfrenette:
I would like to have this re-verified to see that this is actually happening to see if is was a bug with browser or a bug with the code. I am on a Mac so if anyone on a windows computer can double check again I would appreciate it.

jgosmann:
Does happen on my Mac with Firefox. Not sure whether it is exactly the same issue because I don't think it happens due to performance (though this would correspond to @tcstewar's comment). For specific cursor positions I could make the border disappear for as long as I held the mouse still just now. Might be something that only occurs on high DPI displays due to fractional pixel values.

jgosmann:
For some resize operations it is not happening at all, but if it is flickering in one resize operation it is reliable and it is quite easy to find mouse positions where the border stays hidden. I was also able to produce minimal mouse (or actual track pad) movements where the mouse cursor would not move, but the component/graph would shrink by one pixel and lose the border (presumably because it is no longer in hover state than). Seems like some sort of rounding problem in the size calculation.

krisfrenette:
Perfect, thanks for adding

PR #526, opened 2015-08-10
Slider with axis to show range

### LINKED ISSUE #124: corner resize broken on bottom right and bottom left of networks

cchan987:
Here is a demo of slider with an axis to show the range, If it looks ok I will clean up the code (which is not well formatted right now) for merge. 

fix #124, #464


tcstewar:
Cool...  I like the basic look.  The main weirdness is that it seems like the sliders should be squished a bit to the right so that they don't overlap with the line...  but then that means there's a bunch of empty space
to the left when the mouse isn't over the item.  :(  Not sure what the right thing to do is.  Thoughts?

Also, the numbers on the axis seem to be 0 and 1 rather than -1 and 1, and the "Set range..." doesn't work, but I assumer that's being worked on.

jgosmann:
This PR is quite old and marked as "needs changes". Is this still something that makes sense to review? Does it make sense to do so without the required changes?

Issue #500, opened 2015-07-23
Code folder re-organisation


Seanny123:
The layout of the JavaScript `static` code folders is a bit awkward. Would people be open to dividing it into `netgraph_components`, `panel_components` and leaving the core stuff at the root folder?

cchan987:
The thing is that netgraph also handles things that are not really netgraph specific (like event listener for scroll zoom)... so this may get confusing too

Seanny123:
Right, so when I mean `netgraph_components` I mean stuff like `value.js`, `slider.js`, `netgraph_item.js` and `netgraph_conn.js`.

tcstewar:
> Right, so when I mean netgraph_components I mean stuff like value.js, slider.js, netgraph_item.js and netgraph_conn.js.

Hmm, I wouldn't consider value.js or slider.js as part of the netgraph.  The NetGraph is meant to be just that thing in the background that has boxes and arrows (i.e. a graph of the network, in the sense of nodes and connections being a graph).  The NetGraph is used to generate the sliders and value plots and other Components.

tcstewar:
I could definitely see a ```components``` subdirectory, with value.js, slider.js, slidercontrol.js, xyvalue.js, raster.js, and pointer.js in it....  Not sure about any other reorganizing, though.

Seanny123:
Thanks for clarifying Terry. I'm still getting the hang of all the linguo in this code.

For the `panel_components` folder, I was (not super seriously) thinking of the `sim_control`, `menu` and `ace`. What do you folks think of that?

Seanny123:
I would also like to split up `nengo.js` into `nengo_component.js`, `nengo_datastore` and `nengo_utils`.

Seanny123:
Finally, is there a reason that `action.py` isn't it's own folder and all of the different classes split up into different files? Am I being a bit too harsh with the "one class per file" thing that I saw in the Nengo repository?

tcstewar:
> I would also like to split up nengo.js into nengo_component.js, nengo_datastore and nengo_utils

I like the splitting up, but I'd do ```component.js```, ```datastore.js```, and ```nengo.js```

tcstewar:
> Finally, is there a reason that action.py isn't it's own folder and all of the different classes split up into different files? It seems like a logical thing to do, since action.py is totally unlike any of the other components. Am I being a bit too harsh with the "one class per file" thing that I saw in the Nengo repository?

[...]

PR #529, opened 2015-08-16
Reorganised code

Seanny123:
Addresses #500 
- created a js components folder
- moved all the components (matching the components folder in python) into the aforementioned folder
- moved action from the components folder to nengo_gui and renamed it user_action
- extracted component.js and datastore.js from nengo.js


nengo_gui/user_action.py
```diff
@@ -1,5 +1,6 @@
 from nengo.utils.compat import iteritems
 
+# ah, that's probably the problem
```
tcstewar: Not sure what this comment means...  was it left in from debugging?

Seanny123: Yep. Don't know how it got there...

PR #537, opened 2015-08-22
changed y0 to x1

studywolf:
was just working through tutorials and noticed that in the comments y0 was used instead of x1 (pretty sure it's supposed to be x1 and this was just missed in changing it over from y0)

### DIFF
--- a/nengo_gui/examples/tutorial/13-oscillators.py
+++ b/nengo_gui/examples/tutorial/13-oscillators.py
@@ -18,22 +18,22 @@
 # representing multidimensional data.
 
 # Try adjusting the r value to 0.5.  Try 1.5.  What about 0?
-# Try adjusting the speed s.  What happens when it is very slow (0.5)?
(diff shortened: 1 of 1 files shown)


studywolf:
hmm, @tcstewar, should there be parenthesis added in too below, like
```
        return [synapse * (-x[1] * s + x[0] * (r - x[0]**2 - x[1]**2)) + x[0],
                synapse *  (x[0] * s + x[1] * (r - x[0]**2 - x[1]**2)) + x[1]]
```
instead of 
```
        return [synapse * -x[1] * s + x[0] * (r - x[0]**2 - x[1]**2) + x[0],
                synapse *  x[0] * s + x[1] * (r - x[0]**2 - x[1]**2) + x[1]]
```

tcstewar:
> should there be parenthesis added in too below,

yikes, good catch!  Yup, those parentheses should be added.  And that seems to have the nice side effect that you can actually see it grow to the right size, rather than having it jump there immediately.

Issue #538, opened 2015-08-23
reset pulls up default file? 


studywolf:
Anyone else getting this? When I hit reset it takes me back to the default file instead of clearing the layout...

Seanny123:
Yes, however it only happens when I have started the program with `nengo_gui` and not with a file specified. For example, when I run `nengo_gui spa_simi.py`, reset creates the expected behaviour.

studywolf:
ah i see!

i would vote for having the behaviour consistent in both cases! and staying
in the same file just resetting the layout, i hadn't saved the file that i
was working on and lost my network, which could be a frustration

Seanny123:
@tcstewar you apparently wrote this reset code. How can `sim_control.js` pass a filename while it calls reset?

tcstewar:
> How can sim_control.js pass a filename while it calls reset?

It can't and shouldn't, because that reset isn't the one being talked about here.  Due to some poor choices of terminology on my part, there are two different things that get referred to with the term "reset".  The one in sim_control.js is resetting the simulation back to t=0.  The one being referred to here is resetting the layout.  That's done here:

https://github.com/nengo/nengo_gui/blob/master/nengo_gui/static/top_toolbar.js#L91

```
    window.location.assign('/?reset=True');
```

you can add the filename to that call and it should reset correctly using that filename.

Seanny123:
Fixed via #540

PR #540, opened 2015-08-25
Fix reset layout

Seanny123:
Fixes the problem of #538


tcstewar:
Seems fine to me...  thanks for the fix....  :)

PR #528, opened 2015-08-10
Spa similarity plot

Seanny123:
Created a spa similarity plot which tracks the similarity of a SPA pointer over time in a value-like plot, which resulted in a lot of little changes:
- Creation of VariableDataStore for graphs whose lines grow over time
- Creation of pure-JavaScript unit tests for DataStore and VariableDataStore
- First draft of a legends-like object, but so far only for the similarity graph

To test the file use the following code
```
import nengo
import nengo.spa as spa

D = 16

vo = spa.Vocabulary(D)
vo.parse("SHIRT+PANTS+SHOES+SOCKS")

model = spa.SPA()
with model:
    model.display = spa.Buffer(D, vocab=vo)
    model.input_buf = spa.Buffer(D, vocab=vo)

    nengo.Connection(model.input_buf.state.output, model.display.state.input)
```

Issue #387, opened 2015-06-10
Implement a value plot like semantic pointer plot


jgosmann:

xchoo:
This is already a requested plot type (see: https://github.com/nengo/nengo_gui/wiki/Road-Map)

tbekolay:
Better to track in issues than in a road map that is probably out of date already...

PR #542, opened 2015-08-26
Basic SPA similarity plot

Seanny123:
Since feature creep started to happen with #528, this new pull request will focus on the most basic implementation of the SPA similarity plot. This pull request would address #387


Seanny123:
Current things that I would like input on:
- Should I just loop through colours (when I have lines than the palette) or generate random colours?
- Is there a font that doesn't awkwardly squish the minus sign?

jgosmann:
> Is there a font that doesn't awkwardly squish the minus sign?

Are you using an actual minus sign (−) or a hyphen (-) (which is shorte)? Also not to be confused with the en-dash (–) which is not at the same height as the plus sign (en dash, minus, plus: –−+).

Seanny123:
Apparently, the minus sign is unicode and my browser doesn't like that.

jgosmann:
Yes, it's unicode, but we have 2015 so it should be supported ... are you using the HMTL entity `&minus;`?

jgosmann:
We are currently not specifying the encoding anywhere in HTML. Maybe we should ...?

Seanny123:
We should probably specify the endcoding in `page.html`.

Seanny123:
Even with sticking in the unicode minus and unicode space, the labels still jerk around when going from positive to negative. Any ideas on how to fix this?

tbekolay:
Put the + or - label in a fixed width span?

jgosmann:
The `ipdb` dependency has to be removed before a merge.

jgosmann:
In the version I checked out the jerking is because there is no + in front of positive numbers. If we don't want to have it in front of positive numbers it might be possible to add an invisible + (which still takes up space).

jgosmann:
There might still be some jitter as different digits have different widths, but aligning the numbers on the decimal point (which would probably be the right thing to do) seems to be way to complicated in HTML/CSS/JS.

tcstewar:
It might make sense to just go with a fixed-width font to get rid of the jitter....

jgosmann:
(Currently it is also not using an actual minus, but a hyphen.)

> It might make sense to just go with a fixed-width font to get rid of the jitter....

I would try to use actual minus and plus signs first (the latter one maybe invisible). It might be acceptable, if not we can still go the fixed-width font route.

Seanny123:
BTW, I fixed the whitespace thing and the ipdb thing in my last commit.

tbekolay:
> might make sense to just go with a fixed-width font to get rid of the jitter...

Or a fixed-width span? With centered text, it would mean that only the number jitters, and it shouldn't be too distracting.

jgosmann:
> Or a fixed-width span?

The typographically correct minus and plus should have the same width and we don't need to guess the width.

Seanny123:
Guys. I fixed it, guys. Try the new version.

jgosmann:
:cake: Looks fine. :) (Haven't reviewed the code though)

tcstewar:
I like the look!  :)

> Should I just loop through colours (when I have lines than the palette) or generate random colours?

I'd stick with looping through colours for now.  Randomly generating colours is non-trivial (even if all you want to do is avoid colours near white).  The other graphs cycle through the colours, so let's stay consistent here and deal with random generation in a future PR.  :)

tcstewar:
Btw, I'm very much liking the look of the legend being off to the side *outside* the hover/resize box.

Seanny123:
> Btw, I'm very much liking the look of the legend being off to the side outside the hover/resize box.

I like it too, but @celiasmith does not.

Seanny123:
@tbekolay this is the branch that I need help with in terms of configuring Travis.CI. The tests for `data_store.js` run in Mocha, but I have no idea how to set up Travis.CI to run these things automatically. Would you mind helping me or pointing me towards some documentation? I tried looking by myself, but got a bit overwhelmed.

jgosmann:
I might be able to help to if it's easier to discuss in person. The process is basically the following:

1. Add a `.travis.yml` file with the configuration. [Here's one of my projects] which is less complicated than the Nengo one. The language and Python versions might need adjustment to run JS stuff (that's something to look up in the documentation). `before_install` and `install` list command to execute to setup the environment. `script` specifies the command to run the tests and `after_success` commands to run after succesful completion of tests (here it's submitting coverage information to coveralls.io).
2. After adding the file you activate Travis-CI in the admin interface of Travis-CI for that repository.
3. There's no step 3. (I believe)

Seanny123:
Would everyone be okay, with delaying the Travis.CI integration until later? I'd rather push this branch out without the integration testing.

tcstewar:
> Would everyone be okay, with delaying the Travis.CI integration until later? I'd rather push this branch out without the integration testing.

Definitely.  the Travis.CI integration will be great when it happens, but it's damn hard and outside the scope of this PR...  :)

nengo_gui/static/datastore.js
```diff
@@ -105,6 +98,7 @@ Nengo.DataStore.prototype.get_shown_data = function() {
 
     /* find the corresponding index values */
     var index = 0;
+    // Wouldn't a binary search be better?
```
jgosmann: Probably, except if we make only a few loops until the value is found (I don't know this part of the code enough to judge how many loops are expected on average.)
Also depends on how often this code is called (if it's only called a few times per execution, it might not be worth to improve the efficiency, but this seems to be called on a regular basis?)

How regular are the values in `this.times`? If we expect a constant increase in most cases (with maybe a few dropped times), we could use a heuristic to estimate the index which might be even better than binary search.

nengo_gui/static/datastore.js
```diff
@@ -127,7 +121,6 @@ Nengo.DataStore.prototype.is_at_end = function() {
     return (ts.last_time < ts.first_shown_time + ts.shown_time + 0.000000001);
```
jgosmann: Not part of this PR, but as it allows me to comment on code, I am misusing it for this. I think it is more readable to write `1e-9` than `0.000000001` (if I counted correctly). But I'm also wondering why has this magic constant to be added in the first place (and I don't comletely get the logic of this function).

nengo_gui/static/datastore.js
tcstewar: > Probably, except if we make only a few loops until the value is found (I don't know this part of the code enough to judge how many loops are expected on average.)

Often a lot...  In the normal use case there are ~4000 items in the list and the item it's looking for is # ~3500.  

> Also depends on how often this code is called (if it's only called a few times per execution, it might not be worth to improve the efficiency, but this seems to be called on a regular basis?)

It's called once per frame per graph.  So also a lot.  :)

> How regular are the values in this.times? If we expect a constant increase in most cases (with maybe a few dropped times), we could use a heuristic to estimate the index which might be even better than binary search.

99% of the time they are completely regular, since we get data every dt.  The only times that won't be the case will be when running on the spinnaker backend and other real-time hardware systems.

So yeah, some serious optimizing of that should be done.  But that should be a separate PR from this.

Seanny123: Noted in issue #549

nengo_gui/static/datastore.js
tcstewar: > But I'm also wondering why has this magic constant to be added in the first place (and I don't comletely get the logic of this function).

I think it has to do with rounding errors....   ```(i +1)*dt != (i*dt) + dt```.

nengo_gui/components/pointer.py
```diff
@@ -16,6 +16,8 @@ def __init__(self, obj, **kwargs):
         super(Pointer, self).__init__()
         self.obj = obj
         self.data = collections.deque()
+        # toggles whether to use semantic pointer value
+        # as set by the user in the GUI
```
jgosmann: The comment make me expect a boolean. What does a value of `None` mean?

nengo_gui/components/pointer.py
```diff
@@ -43,16 +45,18 @@ def remove_nengo_objects(self, page):
 
     def gather_data(self, t, x):
         vocab = self.vocab_out
-        m = np.dot(vocab.vectors, x)
-        matches = [(mm, vocab.keys[i]) for i, mm in enumerate(m) if mm > 0.01]
+        key_similarity = np.dot(vocab.vectors, x)
```
jgosmann: I would name this `key_similarities` because it is all the similarities for all of the vectors (if I'm not mistaken).

nengo_gui/components/pointer.py
```diff
@@ -43,16 +45,18 @@ def remove_nengo_objects(self, page):
...
+        matches = [(simi, vocab.keys[i]) for i, simi in enumerate(key_similarity) if simi > 0.01]
```
jgosmann: `key_similarity` is a Numpy array. That should allow for something like
```python
over_threshold = key_similarities > 0.01
matches = zip(key_similarities[over_threshold], vocab.keys[over_threshold])
```
I would expect that to be slightly faster and for me it is easier to understand.

Same thing applies below to the pairs stuff.

nengo_gui/components/pointer.py
```diff
@@ -76,6 +80,7 @@ def javascript(self):
     def code_python_args(self, uids):
         return [uids[self.obj], 'target=%r' % self.target]
 
+    # Override the value if set
```
jgosmann: This comment doesn't make sense to me at this place. Which value is overridden? Why does `message` override a value instead of sending a message?

nengo_gui/components/spa_similarity.py
```diff
@@ -0,0 +1,102 @@
+from .pointer import Pointer
```
jgosmann: Not sure if we have a policy on relative imports. But as the next one is absolute, this one should probably be absolute too.

jgosmann: In addition to that, the order of the imports should be:
1. from the Python standard library
2. non-standard packages/modules
3. imports from the package the module is part of (that is `nengo_gui` imports in this case)

nengo_gui/components/spa_similarity.py
```diff
@@ -0,0 +1,102 @@
...
+                           min_value=-1,
+                           show_pairs=False,
+                           **Component.config_defaults)
+
+    def __init__(self, obj, **kwargs):
+        super(SpaSimilarity, self).__init__(obj, **kwargs)
+        try:
+            target_key = kwargs['target']
+        except KeyError:
+            target_key = kwargs['args']
```
jgosmann: `target_key = kwargs.get('target', kwargs['args'])` would save three lines (and is potentially faster)

nengo_gui/components/spa_similarity.py
```diff
@@ -0,0 +1,102 @@
...
+        self.old_pairs_length = 0
+        self.labels = self.vocab_out.keys
+        self.previous_pairs = False
+
+    def gather_data(self, t, x):
+        vocab = self.vocab_out
+
+        # if there's been a change in the show_pairs
+        # Don't know how I feel about the timing of this
+        # maybe trigger on_message?
```
jgosmann: Not sure about this comment. If it stays it should probably explain what the issue with the timing is.

Also the line above: Do we care about capitalization and punctuation?

nengo_gui/components/spa_similarity.py
```diff
@@ -0,0 +1,102 @@
...
+        if self.config.show_pairs != self.previous_pairs:
+            #send the new labels
```
jgosmann: space after #

nengo_gui/components/spa_similarity.py
```diff
@@ -0,0 +1,102 @@
...
+            if self.config.show_pairs:
+                vocab.include_pairs = True
+                self.data.append(
+                    '["show_pairs_toggle", "%s"]' %(
+                        '","'.join(vocab.keys + vocab.key_pairs)))
+                # if we're starting to show pairs, track pair length
+                self.old_pairs_length = len(vocab.key_pairs)
+            else:
+                # Hmmm... an OH HELL NO was triggered...
```
jgosmann: Wat? ;)
What is an OH HELL NO?

nengo_gui/components/spa_similarity.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def update_client(self, client):
+        # while there is data that should be sent to the client
+        while len(self.data) > 0:
+            item = self.data.popleft()
+            # send the data to the client
+            client.write(item, binary=False)
+
+    def javascript(self):
+        """Almost identical to value.py"""
```
jgosmann: This isn't helpful as a docstring. As a comment it would be ok, though some information on what the differences are could be helpful.

nengo_gui/components/spa_similarity.py
```diff
@@ -0,0 +1,102 @@
...
+            self.conn = nengo.Connection(output, self.node, synapse=0.01)
+
+    def remove_nengo_objects(self, page):
+        """Undo the changes made by add_nengo_objects."""
+        page.model.connections.remove(self.conn)
+        page.model.nodes.remove(self.node)
+
+    def message(self, msg):
+        """This should never be called. To be used later for settings pairs?"""
+        raise AttributeError("You can't set the value of a plot!")
```
jgosmann: Would a `NotImplementedError` be more appropriate? Or maybe an `AssertionError` (the assertion being that the function is not called)?

nengo_gui/static/components/netgraph_item.js
```diff
@@ -356,8 +356,11 @@ Nengo.NetGraphItem.prototype.generate_menu = function () {
         }
     }
     if (this.sp_targets.length > 0) {
-        items.push(['Semantic pointer',
+        // Naming things is hard
```
jgosmann: Yes, one of the two things that are hard in computer science. The other:
1. cache invalidation
2. off-by-one errors

Fun aside: This comment doesn't really help.

nengo_gui/static/components/netgraph_item.js
```diff
@@ -356,8 +356,11 @@ Nengo.NetGraphItem.prototype.generate_menu = function () {
...
+        items.push(['Semantic pointer similarity text',
```
jgosmann: How about something like 'Semantic pointer cloud' (the plot reminds me somewhat of those tag clouds you see sometimes on blogs).

nengo_gui/static/components/netgraph_item.js
```diff
@@ -356,8 +356,11 @@ Nengo.NetGraphItem.prototype.generate_menu = function () {
...
                     function() {self.create_graph('Pointer', self.sp_targets[0]);}])
+        items.push(['Semantic pointer similarity graph',
```
jgosmann: Maybe it is helpful to somehow add the word time (or a synonym) into the name to make clear that it shows the similarity over time (whereas the other plot only shows one specific time point).

nengo_gui/static/components/pointer.js
```diff
@@ -178,6 +178,7 @@ Nengo.Pointer.prototype.update = function() {
 
     var items = [];
 
+    /** display the text in proportion to similarity */
```
jgosmann: I believe the `/** comment */` style comment is a JavaScript convention similar to docstrings (but I might be wrong). A docstring doesn't apply to a for-loop. So this should be a normal comment (`/* comment */` or `// comment`).

nengo_gui/static/components/spa_similarity.css
```diff
@@ -0,0 +1,11 @@
+.graph .line {
+    fill: none;
+    stroke: black;
+    stroke-width: 1.5px;    
+}
```
jgosmann: I would insert a newline here.

nengo_gui/static/components/spa_similarity.css
```diff
@@ -0,0 +1,11 @@
...
+.legend {
+    padding: 5px;
+    font: 10px sans-serif;
+    position: absolute;
+    display: inline;
+}
```
jgosmann: There should probably be a newline at the end of the file.

nengo_gui/static/components/spa_similarity.js
```diff
@@ -0,0 +1,202 @@
+/**
+ * Line graph showing semantic pointer decoded values over time
+ * @constructor
+ *
+ * @param {dict} args - A set of constructor arguments (see Nengo.Component)
+ * @param {int} args.n_lines - number of decoded values
+ * @param {float} args.min_value - minimum value on y-axis
+ * @param {float} args.max_value - maximum value on y-axis
+ * @param {Nengo.SimControl} args.sim - the simulation controller
+ */
```
jgosmann: This docstring does not match up with the function signature.

nengo_gui/static/components/spa_similarity.js
```diff
@@ -0,0 +1,202 @@
...
+Nengo.SpaSimilarity = function(parent, sim, args) {
+    Nengo.Value.call(this, parent, sim, args);
+
+    this.data_store = new Nengo.VariableDataStore(this.n_lines, this.sim, args.synapse);
+    this.show_pairs = false;
+
+    var self = this;
+
+    this.colors = Nengo.make_colors(6);
+    this.color_func = function(d, i){return self.colors[i%6]};
```
jgosmann: This line could use some more whitespace like so
`this.color_func = function(d, i) { return self.colors[i % 6] };

nengo_gui/static/components/spa_similarity.js
```diff
@@ -0,0 +1,202 @@
...
+};
+
+Nengo.SpaSimilarity.prototype.update_legend = function(new_labels){
+    // Should figure out how to mix recs and text into one
+
+    var self = this;
+    this.pointer_labels = this.pointer_labels.concat(new_labels);
+
+    // expand the svg
+    this.legend_svg.attr("height", 20*this.pointer_labels.length);
```
jgosmann: space around `*`

jgosmann: Where does the factor of 20 come from?

jgosmann: It also appears below, so it should probably be at least a variable to be able to adjust it in one place.

Also unsure about the other magic numbers in there. I haven't tested it yet, but I'm worried that it might break for different font sizes.

nengo_gui/static/components/spa_similarity.js
```diff
@@ -0,0 +1,202 @@
...
+             .attr('class', 'line')
+             .style('stroke', this.color_func)
+             .attr('d', self.line);
+    // remove any lines that aren't needed anymore
+    this.path.exit().remove();
+
+    /* update the legend text */
+    if(this.legend_svg && shown_data[0].length !== 0){
+        // get the most recent similarity
+         var latest_simi = [];
```
jgosmann: indentation

nengo_gui/static/components/value.js
```diff
@@ -58,13 +62,12 @@ Nengo.Value.prototype.on_message = function(event) {
     var data = new Float32Array(event.data);
     data = Array.prototype.slice.call(data);
     var size = this.n_lines + 1;
+    /** since multiple data packets can be sent with a single event,
+    make sure to process all the packets */
```
jgosmann: Same comment as above about docstring comments applies.

jgosmann: Also the second line should probably be indented to make it visually distinct from the next actual code line.

nengo_gui/static/datastore.js
```diff
@@ -1,15 +1,7 @@
-/**
- * Storage of a set of data points and associated times.
- * @constructor
- *
- * @param {int} dims - number of data points per time
- * @param {Nengo.SimControl} sim - the simulation controller
- * @param {float} synapse - the filter to apply to the data
- */
```
jgosmann: What is the reason to remove this comment?

nengo_gui/static/datastore.js
```diff
@@ -146,3 +139,177 @@ Nengo.DataStore.prototype.get_last_data = function() {
...
+        },
+        set: function(dim_val){
+            // throw a bunch of errors if bad things happen
+            // assuming you can only grow dims and not shrink them
+            if(this._dims < dim_val){
+                for (var i=0; i < dim_val - this._dims; i++) {
+                    this.data.push([]);
+                }
+            } else if(this._dims > dim_val) {
+                console.log("OH HELL NO");
```
jgosmann: This should be something informative ...
Would it be better to raise an exception?

nengo_gui/static/datastore.js
```diff
@@ -146,3 +139,177 @@ Nengo.DataStore.prototype.get_last_data = function() {
...
+            }
+        }
+    }
+}
+
+/**
+ * Return just the data that is to be shown
+ */
+Nengo.VariableDataStore.prototype.get_shown_data = function() {
+     var offset = this.get_offset();
```
jgosmann: indentation

nengo_gui/static/nengo.js
```diff
@@ -57,7 +57,7 @@ Nengo.create_websocket = function(uid) {
  */
 Nengo.make_colors = function(N) {
     //Color blind palette with blue, green, red, magenta, yellow, cyan
-    var palette=["#1c73b3","#039f74","#d65e00","#cd79a7","#f0e542","#56b4ea"];
+    var palette = ["#1c73b3","#039f74","#d65e00","#cd79a7","#f0e542","#56b4ea"];
```
jgosmann: It would be nice to have spaces also after `,`

nengo_gui/static/nengo.js
```diff
@@ -57,7 +57,7 @@ Nengo.create_websocket = function(uid) {
  */
 Nengo.make_colors = function(N) {
     //Color blind palette with blue, green, red, magenta, yellow, cyan
```
jgosmann: Space after `//`

nengo_gui/static/test/datastore_test.js
```diff
@@ -0,0 +1,201 @@
...
+        assert.deepEqual(data_store.data, [[3.1, 4.1], [3.2, 4.2], [4.3]]);
+        data_store.dims = 4;
+        data_store.push([1.5, 5.1, 5.2, 5.3, 5.4]);
+        data_store.update();
+        assert.deepEqual(
+            data_store.data,
+            [[3.1, 4.1, 5.1], [3.2, 4.2, 5.2], [4.3, 5.3], [5.4]]);
+    });
+
+    // OH FUCK
```
jgosmann: This has to be censored. ;)

jgosmann:
Did a first pass, but I have to go over some parts again in more detail (and actually test it).

Do we aim to follow a certain style in the JS? The placement and existence of spaces in some constructs varies wildly. :/

There are a lot of comments that ask questions like "should I do this in this way or better another way?" It would be nice to reduce them as much as possible by making decisions on those issues. If that's currently not possible they probably should be marked with `TODO`, `FIXME` or `XXX ` to make it easy to find those later on (but such comments tend to stay around and never get fixed, so this is not my preferred solution). Maybe some of those can be converted to issues, too (especially if they are not directly related to a specific piece of code).

jgosmann:
Getting an exception. Steps to reproduce:
1. Open `examples/spa.py`
2. Add similarity graph for `b`
3. Run for a while
4. Pause
5. Activate show pairs for the graph
6. Run for a while (the lines for the pairs seem to fill in from the left, but I think they should be filling in from the right, this might be another bug)
7. Pause
8. Add similarity graph for `c`

The exception:
```
  File "/home/jgosmann/Documents/projects/nengo_gui/nengo_gui/page.py", line 454, in runner
    self.sim.step()
  File "/home/jgosmann/.pyenv/versions/2.7.10/lib/python2.7/site-packages/nengo/simulator.py", line 187, in step
    step_fn()
  File "/home/jgosmann/.pyenv/versions/2.7.10/lib/python2.7/site-packages/nengo/builder/node.py", line 39, in step
    y = fn(t_sig.item(), *args) if t_in else fn(*args)
  File "/home/jgosmann/Documents/projects/nengo_gui/nengo_gui/components/spa_similarity.py", line 67, in gather_data
    pair_similarity = np.dot(vocab.vector_pairs, x)
TypeError: unsupported operand type(s) for *: 'NoneType' and 'float'
```

nengo_gui/static/components/spa_similarity.js
```diff
@@ -0,0 +1,202 @@
...
+            latest_simi.push(shown_data[i][shown_data[i].length - 1]);
+        }
+
+        // update the text in the legend
+        var texts = this.legend_svg.selectAll("text").data(this.pointer_labels);
+
+        texts.attr("x", 15)
+              .attr("y", function(d, i){ return i *  20 + 9;})
+              .html(function(d, i) {
+                    var sign = "&nbsp;&nbsp;&nbsp;";
```
jgosmann: How difficult would it be to use an invisible + instead of non-breaking spaces? I'm still experiencing some jitter.

jgosmann:
The exception seems to also triggered by toggling show pairs  three times on a single plot.

nengo_gui/components/spa_similarity.py
```diff
@@ -0,0 +1,102 @@
+from .pointer import Pointer
+from nengo_gui.components.component import Component
+
+import numpy as np
+import nengo
+
+import struct
+import copy
```
jgosmann: `struct` and `copy` are unused imports

nengo_gui/components/spa_similarity.py
jgosmann: pylint told me that `target_key` is unused, so all four lines can be removed.

nengo_gui/components/spa_similarity.py
```diff
@@ -0,0 +1,102 @@
...
+                # if we're starting to show pairs, track pair length
+                self.old_pairs_length = len(vocab.key_pairs)
+            else:
+                # Hmmm... an OH HELL NO was triggered...
+                vocab.include_pairs = False
+                self.data.append(
+                    '["show_pairs_toggle", "%s"]' %(
+                        '","'.join(vocab.keys)))
+
+        if(self.old_vocab_length != len(vocab.keys)):
```
jgosmann: unnecessary parens after if

jgosmann: Coding too much JS? ;)

nengo_gui/components/spa_similarity.py
```diff
@@ -0,0 +1,102 @@
...
+            # pass all the missing keys
+            legend_update = []
+            legend_update.append(vocab.keys[-1])
+            self.old_vocab_length = len(vocab.keys)
+            # and all the missing pairs if we're showing pairs
+            if self.config.show_pairs:
+                legend_update += vocab.key_pairs[self.old_pairs_length:]
+                self.old_pairs_length = len(vocab.key_pairs)
+
+            self.data.append('["update_legend", "%s"]' %('","'.join(legend_update)))
```
jgosmann: Line too long

nengo_gui/components/spa_similarity.py
```diff
@@ -0,0 +1,102 @@
...
+                    pointer_labels=self.labels)
+        json = self.javascript_config(info)
+        return 'new Nengo.SpaSimilarity(main, sim, %s);' % json
+
+    def add_nengo_objects(self, page):
+        with page.model:
+            output = self.obj.outputs[self.target][0]
+            self.node = nengo.Node(self.gather_data,
+                                   size_in=self.vocab_out.dimensions)
+            self.conn = nengo.Connection(output, self.node, synapse=0.01)
```
jgosmann: `self.node` and `self.conn` should be defined in `__init__` (you can set them to `None`)

jgosmann:
Taking `examples/spa.py` and adding a pointer called 'A,b' to the vocab gives me a legend showing to entries for 'D', but none for that vector. (Admitted 'A,b' is not a normal semantic pointer name, so potentially an exception should already be thrown when adding it to the vocab.) Another bad name is 'A"b'

Adding a a pointer 'E' it doesn't show up in the legend, but an additional line is shown.

jgosmann:
Resetting the model and adding a pointer 'X' worked fine, but the pausing the model and removing the manipulation of the vocab broke. When restarting the simulation 'X' did not disappear from the legend and two (!) pointers 'D' appeared in the legend with value `NaN`.

(That's probably enough ways to break the legend for now.)

New plots should probably have an empty legend instead of showing 'A' and 'C'.

jgosmann:
One step to fix these issues is probably to escape the names of the pointers (they should be considered untrusted user input in a sense)

nengo_gui/components/spa_similarity.py
```diff
@@ -0,0 +1,102 @@
+from .pointer import Pointer
+from nengo_gui.components.component import Component
+
+import numpy as np
+import nengo
+
+import struct
+import copy
+
+class SpaSimilarity(Pointer):
```
jgosmann: It seems this class overrides almost all of `Pointer`'s method. Thus, I would suggest to not derive from `Pointer`. If there is code worth to share, it might be better to create an abstract class `SpaPlot` (or another appropriate name) containing the shared code and have `Pointer` and `SpaSimilarity` derive from it.

nengo_gui/components/spa_similarity.py
```diff
@@ -0,0 +1,102 @@
...
+                # Hmmm... an OH HELL NO was triggered...
+                vocab.include_pairs = False
+                self.data.append(
+                    '["show_pairs_toggle", "%s"]' %(
+                        '","'.join(vocab.keys)))
+
+        if(self.old_vocab_length != len(vocab.keys)):
+            # pass all the missing keys
+            legend_update = []
+            legend_update.append(vocab.keys[-1])
```
jgosmann: This seems to assume that at most one new key is added at a time? Is this always True? It seems to me that multiple keys at once could be added.

nengo_gui/components/spa_similarity.py
```diff
@@ -0,0 +1,102 @@
...
+        key_similarity = np.dot(vocab.vectors, x)
+        simi_list = ['{:.2f}'.format(simi) for simi in key_similarity]
+        if self.config.show_pairs:
+            pair_similarity = np.dot(vocab.vector_pairs, x)
+            simi_list += ['{:.2f}'.format(simi) for simi in pair_similarity]
+
+        self.data.append( '["data_msg", %g, %s]' %(t, ",".join(simi_list) )  )
+        self.previous_pairs = self.config.show_pairs
+
+    def update_client(self, client):
```
jgosmann: This is identical to the parent class. No need to override.

nengo_gui/static/components/spa_similarity.js
```diff
@@ -0,0 +1,202 @@
...
+}
+
+Nengo.SpaSimilarity.prototype.data_msg = function(push_data){
+
+    var data_dims = push_data.length - 1;
+
+    if(data_dims !== this.n_lines){
+      this.data_store.dims = data_dims;
+      this.n_lines = data_dims;
+    }
```
jgosmann: Might it be better to have the resizing of the dims handled in the `VariableDataStore` so that a call to `push` automatically adjusts dimensions? Or is this risky because it might be used in places where this automatical adjustment shouldn't happen?

nengo_gui/static/components/spa_similarity.js
```diff
@@ -0,0 +1,202 @@
...
+    var self = this;
+    this.pointer_labels = this.pointer_labels.concat(new_labels);
+
+    // expand the svg
+    this.legend_svg.attr("height", 20*this.pointer_labels.length);
+
+    // Data join
+    var recs = this.legend_svg.selectAll("rect").data(this.pointer_labels);
+    var texts = this.legend_svg.selectAll("text").data(this.pointer_labels);
+    // nothing to update
```
jgosmann: This comment is slightly confusing. Why is there nothing to update in a function called `update_legend`? It is also not clear what code it refers to ... (I guess the non-existing code that would update stuff if there were stuff to update?)

nengo_gui/static/components/spa_similarity.js
```diff
@@ -0,0 +1,202 @@
...
+          .append("text")
+          .attr("x", 15)
+          .attr("y", function(d, i){ return i *  20 + 9;})
+          .html(function(d, i) {
+                return self.pointer_labels[i];
+           });
+
+};
+
+/* there a three types of messages that can be recieved:
```
jgosmann: typo in received

nengo_gui/static/datastore.js
```diff
@@ -146,3 +139,177 @@ Nengo.DataStore.prototype.get_last_data = function() {
     }
     return shown;
 }
+
+
+Nengo.VariableDataStore = function(dims, sim, synapse){
```
jgosmann: Would be `ResizeableDataStore` be a better name?
`VariableDataStore` might make one think of a `DataStore` for variables.

jgosmann: Actually `EnlargeableDataStore` might be better (as it doesn't allow shrinking).
(I also used `GrowableArray` for [something similar with Numpy arrays before](https://github.com/jgosmann/goppy/blob/master/goppy/growable.py).

jgosmann:
I think there is too much code duplication in `DataStore` and `VariableDataStore`. (One function seems to be even identical and could be deleted as `VariableDataStore` derives from `DataStore` if I am not mistaken.) To me it seems that `VariableDataStore` could be used as a substitute for `DataStore`. So it might be best to get rid of `DataStore` and just keep `VariableDataStore` (which then could be named `DataStore`).

nengo_gui/static/test/datastore_test.js
```diff
@@ -0,0 +1,201 @@
...
+var sim = {
+            time_slider:{
+                first_shown_time:0,
+                shown_time:0,
+                last_time:0,
+                kept_time:2
+            }
+}
+
+// I should initialise the data_store before each test or does it do that already?
```
jgosmann: Should be figured out before the merge ...

nengo_gui/static/test/datastore_test.js
```diff
@@ -0,0 +1,201 @@
...
+describe("DataStore", function() {
+    var data_store = new Nengo.DataStore(2, sim, 0);
+
+    beforeEach(function() {
+        data_store = new Nengo.DataStore(2, sim, 0);
+    });
+
+    it("accepts data", function() {
+        data_store.push([0.0, 1.1, 1.2]);
+        assert.deepEqual(data_store.data, [[1.1], [1.2]]);
```
jgosmann: Why do we only get back 1.1 and 1.2 if also 0.0 gets pushed to the data store?

nengo_gui/static/test/datastore_test.js
```diff
@@ -0,0 +1,201 @@
...
+    });
+
+    it("orders data", function() {
+        data_store.push([1.0, 3.1, 3.2]);
+        data_store.push([0.5, 2.1, 2.2]);
+        assert.deepEqual(data_store.data, [[2.1], [2.2]]);
```
jgosmann: Why as `data` only that one row with 2.x? (I suppose, the first value is the time?)

nengo_gui/static/components/spa_similarity.js
```diff
@@ -0,0 +1,202 @@
...
+    this.div.appendChild(this.legend);
+    this.legend_svg = Nengo.draw_legend(this.legend, args.pointer_labels, this.color_func);
+};
+
+Nengo.SpaSimilarity.prototype = Object.create(Nengo.Value.prototype);
+Nengo.SpaSimilarity.prototype.constructor = Nengo.SpaSimilarity;
+
+Nengo.SpaSimilarity.prototype.show_pairs_toggle = function(new_labels){
+    // clear the database and make a new one
+    self.data_store = new Nengo.VariableDataStore()
```
jgosmann: There seems to be a reset function on `DataStore`. Why are you not using it?

nengo_gui/static/test/datastore_test.js
```diff
@@ -0,0 +1,201 @@
...
+        assert.deepEqual(data_store.data, [[], []]);
+    });
+
+    it("throws away old data", function() {
+        data_store.push([0.5, 2.1, 2.2]);
+        data_store.push([1.0, 3.1, 3.2]);
+        data_store.push([1.5, 4.1, 4.2]);
+        data_store.sim.time_slider.last_time = 3;
+        data_store.update();
+        assert.deepEqual(data_store.data, [[3.1, 4.1], [3.2, 4.2]]);
```
jgosmann: Unclear how `last_time` relates to the lower time bound and thus what data is expected to remain in the data store (and this might potentially change?)

nengo_gui/static/test/datastore_test.js
```diff
@@ -0,0 +1,201 @@
...
+    it("accepts data", function() {
+        data_store.push([0.0, 1.1, 1.2])
+        assert.deepEqual(data_store.data, [[1.1], [1.2]])
+    });
+
+    it("accepts jagged data", function() {
+        data_store.push([0.0, 1.1, 1.2])
+        data_store.dims = 3;
+        assert.equal(data_store.dims, 3)
+        assert.equal(data_store._old_dims, 2)
```
jgosmann: Is the underscore in `_old_dims` supposed to indicate a protected attribute as in Python? In that case it should be considered an implementation detail and not be tested.

Why is it stored anyways? It seems to be only set, but never read except in these tests.

jgosmann:
Finished my review for now. I (or someone else) have to take another look after the issues have been addressed.

nengo_gui/components/spa_similarity.py
Seanny123: So the timing issue is that when a person toggles `show_pairs`, they won't see an update until the simulation starts running. This is a bit awkward because the user also doesn't get any indication that their choice has been registered. Do you think this is a legitimate problem? Should it wait until the more hardcore version of this legend is finished?

Seanny123: After talking to @tcstewar, we've decided that blanking out the plot and updating the legend upon a `show_pairs` toggle is advisable.

Seanny123:
I'm going to start working on fixing a few things (zooming, show_pairs update) in this pull request, which might make it hard to view and understand some of the old comments, so I'm going to write down things I still need to investigate here:
- [ ] According to Google's JavaScript style-guide, is there a difference between doc-string like comments and normal comments?
- [ ] Make an invisible negative sign in front of the labels to reduce jitter
- [ ] I've named the two SPA visualizations "Semantic Pointer Cloud" and "Semantic Pointer Plot". This should be discussed at the next Nengo meeting.

jgosmann:
> Make an invisible negative sign in front of the labels to reduce jitter

A in invisible plus sign might be better. It is supposed to have the same width as a minus sign, but would semantically be correct (which primarily might make the code less confusing; I don't think we need to care about screen readers for visually impaired people, but we don't have to make it unnecessarily harder for those).

Seanny123:
@jgosmann one of the ways you mention breaking the legend is by adding various semantic pointers to the vocabulary. How are you going about this? When I try to add the aforementioned pointers via the Semantic Pointer Cloud modal dialog, they are rejected. They are also rejected when I try to add them via code.

Seanny123:
The code is ready for review again. I wasn't able to solve the jitter that Jan is seeing. I'll have to look at his computer to confirm.

jgosmann:
> @jgosmann one of the ways you mention breaking the legend is by adding various semantic pointers to the vocabulary. How are you going about this?

I was using `Vocabulary.add(...)`

tcstewar:
Looking nifty!

I managed to get one crash when clicking on "Hide pairs" while a model was running:

```
  File "c:\users\terry\documents\github\nengo_gui\nengo_gui\components\spa_similarity.py", line 59, in gather_data
    pair_similarity = np.dot(vocab.vector_pairs, x)
TypeError: unsupported operand type(s) for *: 'NoneType' and 'float'
```

I'll try to figure out a way to replicate it, but my guess is it's a timing thing, since vocab.vector_pairs can be None briefly in there.  So the solution is probably just to catch that TypeError and do nothing.

tcstewar:
And the bug where new SPA lines appear from left to right is still there.  

To replicate:
 - Open up tutorials/19-spa.py
 - Make a semantic pointer plot
 - Start the model running
 - Use the semantic pointer cloud to "Set value..." to A (this works fine)
 - Let the model continue running
 - Now use the semantic pointer cloud to "Set value..." to B

The new B line will start drawing from the left, even though the data being plotted should be on the right.

tcstewar:
Here's a screenshot to show what I mean:

image

Seanny123:
I'm a bit stuck trying to solve the bug that Terry's noted. I have the offset of each line, so this should be theoretically easy to correct, but I don't know how to make this mesh with the D3.js `line` object [as elaborated upon here](http://stackoverflow.com/questions/32555442/d3-js-line-object-keep-track-of-lines-processed).

I could also just embed the offset value, but that would mean duplicating the offset value for each point of data, which feel as bit inefficient. Does anyone have any other ideas?

Seanny123:
@tcstewar I think I fixed everything you commented on. Let me know if there's anything else.

tcstewar:
Looking good!  

New bug (which I think is one you mentioned to me):

Start the model running with a spa plot showing.  Now add a blank line in the code and press play again: the spa plot will stop updating.

Seanny123:
Fixed the reset error. It's now ready for another review.

nengo_gui/components/pointer.py
```diff
@@ -8,17 +7,15 @@
...
-        self.data = collections.deque()
+        super(Pointer, self).__init__(obj, **kwargs)
+        # the semantic pointer value as set by the user in the GUI
+        # a value of 'None' means do not override
         self.override_target = None
-        self.target = kwargs.get('args', 'default')
-        self.vocab_out = obj.outputs[self.target][1]
         self.vocab_in = obj.inputs[self.target][1]
 
     def attach(self, page, config, uid):
```
tcstewar: Why is attach being overridden here?  This looks like it's just redoing exactly what the base class already does.

nengo_gui/static/ace.js
```diff
@@ -22,9 +22,9 @@ Nengo.Ace = function (uid, args) {
     this.ws.onmessage = function(event) {self.on_message(event);}
 
     this.current_code = '';
-    var code_div = document.createElement('div')
-    code_div.id = 'editor'
-    $('#rightpane').append(code_div);
+    this.code_div = document.createElement('div')
+    this.code_div.id = 'editor'
+    $('#rightpane').append(this.code_div);
```
tcstewar: Why is this being changed in this PR?

nengo_gui/static/components/spa_similarity.js
```diff
@@ -0,0 +1,205 @@
+/**
+ * Line graph showing semantic pointer decoded values over time
+ * @constructor
+ *
+ * @param {dict} args - A set of constructor arguments (see Nengo.Component)
+ * @param {int} args.n_lines - number of decoded values
+ * @param {Nengo.SimControl} args.sim - the simulation controller
```
tcstewar: This list of parameters doesn't match the ones for the function

nengo_gui/static/components/spa_similarity.js
```diff
@@ -0,0 +1,205 @@
...
+          .append("text")
+          .attr("x", 15)
+          .attr("y", function(d, i){ return i *  20 + 9;})
+          .html(function(d, i) {
+                return self.pointer_labels[i];
+           });
+
+};
+
+/* there a three types of messages that can be received:
```
tcstewar: a -> are

nengo_gui/static/components/value.js
```diff
@@ -58,13 +62,12 @@ Nengo.Value.prototype.on_message = function(event) {
...
     while (data.length >= size) {
         this.data_store.push(data.slice(0, size));
         data = data.slice(size);
     }
-    if (data.length > 0) {
-        console.log('extra data: ' + data.length);
-    }
```
tcstewar: The "packing all data into one message" way of sending data is still new enough that I like having this check and error message in the code.

nengo_gui/static/components/value.js
```diff
@@ -123,7 +121,6 @@ Nengo.Value.prototype.generate_menu = function() {
     items.push(['Set range...', function() {self.set_range();}]);
 
     // add the parent's menu items to this
-    // TODO: is this really the best way to call the parent's generate_menu()?
```
tcstewar: Are we now sure this is the best way to call the parent's generate_menu()?

nengo_gui/static/datastore.js
```diff
@@ -146,3 +148,210 @@ Nengo.DataStore.prototype.get_last_data = function() {
...
+
+    /** return the visible slice of the data */
+    var shown = [];
+    for (var i = 0; i < this._dims; i++) {
+        if(last_index - offset[i] >= 0){
+            shown.push(this.data[i][last_index - offset[i]]);
+        }
+    }
+    return shown;
+}
```
tcstewar: Hmm...  lots of duplicate code here between DataStore and GrowableDataStore

Seanny123: Right, I made a separate issue #556 for that, since the optimisation of that code is kind of outside the scope of this pull request. In this pull request I just needed my own jagged datastore.

Seanny123:
Ready for review again.

Seanny123:
Nvm. I somehow broke showing pairs in the pointer cloud in the process of writing this pull request.

Seanny123:
Fixed again and ready for review.

tcstewar:
New bug:  when I click "hide pairs", it stops showing me the numbers in the legend.  

The same thing happens when there are no terms in the vocabulary when the graph is shown.

nengo_gui/static/components/value.js
tcstewar: Just to make it clear, the above comment is me indicating that I don't think those three lines should have been deleted.

nengo_gui/examples/tutorial/23-spa-binding.py.cfg
```diff
@@ -1,27 +1,27 @@
-_viz_0 = nengo_gui.components.PointerTemplate(model.memory)
```
tcstewar: Did you mean to update this .cfg file?

nengo_gui/static/components/pointer.js
```diff
@@ -182,11 +181,12 @@ Nengo.Pointer.prototype.update = function() {
 
     var items = [];
 
+    // display the text in proportion to similarity
     for (var i=0; i < data.length; i++) {
         var size = parseFloat(data[i].substring(0,4));
         var span = document.createElement('span');
         span.innerHTML = data[i].substring(4);
-        this.pdiv.appendChild(span);
+        this.   pdiv.appendChild(span);
```
tcstewar: whitespace?

nengo_gui/static/components/value.js
```diff
@@ -3,6 +3,7 @@
  * Line graph showing decoded values over time
  * @constructor
  *
+ * @param {DOMElement} args.parent - the element to add this component to
```
Seanny123: remove "args"

Seanny123:
Ready for review again. Thanks for all the help with git Terry. `git commit -p` is a godsend.

tcstewar:
Looks good to me!  Very nice functionality to have!

I'll let other people stare at it an then merge tomorrow, just in case.  :)

Seanny123:
Fixed the jitter that became apparent from different fonts.

tcstewar:
That looks good to me....

tcstewar:
Nice work on this!

PR #543, opened 2015-08-26
Got rid of disposable_js

### LINKED ISSUE #539: Rename `disposable_js.py`
Would it be a good idea to clarify the name of `disposable_js.py` to reflect the function that it serves in collaboration with `modal.js`? I was thinking of something like `modal_response.py`. Does anyone else have other suggestions?

Seanny123:
Fixes #539

### DIFF
--- a/nengo_gui/static_plots.py
+++ b/nengo_gui/static_plots.py
@@ -1,0 +1,71 @@
+import numpy as np
+
+import nengo
+from nengo.utils.ensemble import response_curves, tuning_curves
+
+
+class PlotInfo(object):
+    def __init__(self, title, plot="none"):
+        self.title = title
+        self.plot = plot
+        self.warnings = []
+        self.x = None
+        self.y = None
+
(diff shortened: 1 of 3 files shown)


tbekolay:
The commit message is a bit misleading; looks like you're just deleting `disposable_js`. Perhaps `
Split disposable_js into modal_js, static_plots`

nengo_gui/static_plots.py
```diff
@@ -0,0 +1,70 @@
+import numpy as np
+
+import nengo
+from nengo.utils.ensemble import response_curves, tuning_curves
+
+class PlotInfo(object):
```
tbekolay: PEP8 says two blank lines before class and function definitions

tbekolay:
Aside from the inline comment, the rearranging looks good! Note that I haven't checked it out and run it, just looked at the code.

tcstewar:
LGTM  :)

PR #544, opened 2015-08-27
Pass interactive server argument properly

Seanny123:
Need this for setting up `nengo_gui` for testing with Selenium


tcstewar:
Oops...  good catch.  The fix looks good to me!

PR #546, opened 2015-08-29
Create unit-tests for Nengo GUI

Seanny123:
Started creating unit-tests to run with `py.test` and `selenium webdriver`. This is definitely a first stab at the issue and there's still a lot of things to improve:
- [ ] Still a ton of tests to be written
- [ ] Need to check if the GUI is being reset properly between tests
- [ ] Need to integrate with Travis.CI
- [ ] Probably isn't running as fast as it could be

Note: Is only going to run FireFox until we have our own test server

To test, go to the `nengo_gui/static/components/tests` folder and run `py.test test_basic_netgraph.py`.

Any help writing unit tests would be greatly appreciated! If you have any trouble getting started with Selenium, please let me know. I would be happy to help. (:

P.S. for exploring unit-tests I like to start an ipython shell with a driver already started by running the script below
```
import nengo_gui
import threading
import traceback as tb
import logging
import ipdb

from selenium import webdriver

logging.basicConfig(filename='example.log',level=logging.DEBUG)
port = 8800
model_path= "model_basic.py"

gui = None
def do_gui():
    logging.info("Launching nengo_gui (port=%d)", port)
    #print("Launching nengo_gui (port=%d)", port)
    gui = nengo_gui.GUI(model_path, interactive=False)
    try:
        gui.start(port=port, browser=False)
    except:
        logging.debug("Server raised exception, suppressed:\n%s",
                      tb.format_exc(()))

# start the server in it's own thread
threading.Thread(target=do_gui).start()

print("webdriver")
d = webdriver.Firefox()
d.get("http://localhost:%s" %port)

ipdb.set_trace()
```

PR #548, opened 2015-09-01
Fix broken editor due to refactoring in beb6638

tcstewar:
I'm rushing through this fix, as right now the editor is broken in master (it doesn't execute the script as you type)


Seanny123:
Oh dear... Did I mention that I started a pull request with `py.test` fixtures for easy unit-testing with Selenium #546

Issue #363, opened 2015-06-09
Nodes and direct mode xy-plots can be confusing


tbekolay:
It's easy to see that an xy-plot is working for an ensemble, because the noisy representation moves a bit making the line visible. That's not the case with nodes and ensembles in direct mode; if the value is static, it looks like the xy-plot isn't even working. You can't even really see the blue dot where the value actually is.

I'm not sure the best resolution to this, though, because not being able to see it is, basically, correct. But perhaps for direct mode and node xy-plots, we can make the line thicker? Or, actually, maybe a better resolution is to keep the line the same size, but draw a big blue at the value at the current timestep (the line would just be trail or where the value was recently).

tbekolay:
Then, I suppose the question would be, should that circle still be drawn on spiky ensembles?

bjkomer:
We could have circle/line thickness a configurable option for those plots. It would default to the regular way for spiking ensembles, and default to a bigger dot for direct mode.

tcstewar:
I like the idea of a larger dot at the end all the time (for spiky ensembles as well as Direct and LIFRate modes)...  one thing that'd do is help indicate what the most recent data is when it's paused (or a screen shot)

tcstewar:
Closed by #646

Issue #271, opened 2015-06-02
xy-plot axis lines should be at 0, not at half range


tbekolay:
Setting a range of -1,2 does not work as expected. However, the values of the X and Y plot should still be identical.

tcstewar:
Thanks to #646 the axes now draw correctly.

PR #552, opened 2015-09-04
Solve issues with XY plots

### LINKED ISSUE #551: XYPlot fails with 1 dimension
**Scenario:**
1. Create a node with at least 2 dimensions
2. Create XY-plot
2. Switch node back to one dimension

**Result:**
Resizing anything fails catastrophically.

**Proposed Solution:**
Shade out plot and change the label to "invalid dimensions", which informs the user of what is wrong, without making the decision to remove the plot for them.

I'm aiming for something that kind of looks like this, but feedback is appreciate on how to make this less visually harsh:
image

Seanny123:
Should resolve #551 , #363 and #271

### DIFF
--- a/nengo_gui/components/xyvalue.py
+++ b/nengo_gui/components/xyvalue.py
@@ -8,6 +8,7 @@ from nengo_gui.components.component import Component
 
 
 class XYValue(Component):
+    # TODO: What is index_x and index_y used for?
     config_defaults = dict(max_value=1, min_value=-1, index_x=0, index_y=1,
                            **Component.config_defaults)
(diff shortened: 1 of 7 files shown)


tcstewar:
Cool!   I like the look.  I might be tempted to put the green circle behind the blue line, just to see what that's like.

Seanny123:
Current status:
- Still get awkward error when switching from 2D to 1D
- I somehow lost the litte circle that used fix the direct-mode problem
- The axes still don't cross at zero

So basically nothing is fixed.

Issue #559, opened 2015-09-21
Creating semantic pointer cloud on a spa.Compare network causes error


Seanny123:
If you try to create a Semantic Pointer Cloud on a spa.Compare network it causes an error, because `pointer.py` can't figure out the input vocabulary, because it has multiple inputs. However, this raises a few other issues. Additionally, it should be possible to get a value plot from it's output.

Firstly, neither SPA related plots should be available for that component, since it's output is a scalar and nothing useful can be shown from either the Cloud or the Plot.

Secondly, this same error will be triggered for `spa.Bind` once it's merged to master. In the `spa.Bind` case, I could fix this bug. In other words, I would recommend allowing a Semantic Pointer Cloud and even setting an override value. What do other people think?

tcstewar:
Wow, that's been broken for a long time.  There's no reason for pointer.py to even look at obj.inputs.  It should do everything on the output, I think.   And you are right that it should ignore scalar outputs.

Seanny123:
> There's no reason for pointer.py to even look at obj.inputs

I think the reason `pointer.py` accesses the inputs is to enable over-writing the value. That's what it says in the git blame. Is it possible to do everything on the output instead? I ask, since you're the one who wrote this code in the first place and I don't totally understand how the value is being over-ridden by `gather-data`. (:

tcstewar:
Yes, it's totally possible to do it on the output, since the output is just a Node which can be used as an input.  However, it'll probably take some tweaking of the strength of the overriding input value.

Seanny123:
[This is how I currently understands it works](https://docs.google.com/drawings/d/16e_0rWObGFP_cVN7vG1FS7hD5fZAZ9K6I1e69QHIvAU/edit?usp=sharing). How do you recommend over-riding the output other than this method?

tcstewar:
That's not how it works.

How it currently works is on the left in the picture below.  I've drawn in the Node and the two Connections that are made in pointer.py::add_nengo_objects()

![image](https://cloud.githubusercontent.com/assets/2380772/10003612/40ea2a7e-607b-11e5-9831-2c07e679f539.png)

That approach only works if there's one natural input and one natural output from a component, and if the component is meant to just represent that value.  So, it basically only works for spa.State and spa.AssociativeMemory.  

[...]

PR #561, opened 2015-09-23
Better spa override

Seanny123:
Resolves part of #559


nengo_gui/components/pointer.py
```diff
@@ -53,11 +69,12 @@ def gather_data(self, t, x):
...
         if self.override_target is None:
-            return self.vocab_in.parse('0').v
+            return self.default_override_val
         else:
             v = (self.override_target.v - x) * 3
-            if self.vocab_in is not self.vocab_out:
-                v = np.dot(self.vocab_out.transform_to(self.vocab_in), v)
+            if self.loop_in:
+                if self.vocab_in is not self.vocab_out:
+                    v = np.dot(self.vocab_out.transform_to(self.vocab_in), v)
```
tcstewar: That call to ```transform_to``` is going to be incredibly expensive to compute every timestep.  If we do something like this, it'd be a huge speed improvement to do it outside the loop, I think.

I'm also unsure that we even need this complexity.  The one case where it could be used is an AssociativeMemory with a different vocab for the input and output, and I'd be tempted to just remove that case from the whitelist.

nengo_gui/components/pointer.py
```diff
@@ -16,17 +15,34 @@ def __init__(self, obj, **kwargs):
         # the semantic pointer value as set by the user in the GUI
         # a value of 'None' means do not override
         self.override_target = None
-        self.vocab_in = obj.inputs[self.target][1]
+        loop_in_whitelist = [spa.Buffer, spa.Memory, spa.State, spa.AssociativeMemory]
```
tcstewar: All of these should only be whitelisted for the case when ```self.target == 'default'```

nengo_gui/components/pointer.py
```diff
@@ -16,17 +15,34 @@ def __init__(self, obj, **kwargs):
...
+
+        if type(obj) in loop_in_whitelist:
+            self.vocab_in = obj.inputs[self.target][1]
+            self.loop_in = True
+            self.default_override_val = self.vocab_in.parse('0').v
```
tcstewar: since we have ```size_out``` here, this would be clearer as ```np.zeros(self.size_out)```

nengo_gui/components/pointer.py
tcstewar: Although before I suggest removing it from the whitelist, how well did it work when you tried it?  What sort of things did you try?

nengo_gui/components/pointer.py
Seanny123: That's how `self.target` is defined in `SpaPlot`.

tcstewar: > That's how self.target is defined in SpaPlot.

No it's not.  The line in SpaPlot says ```self.target = kwargs.get('args', 'default')``` which means that it *defaults* to ```'default'```, but if someone passes it in to the constructor something like ```Pointer(target='A')``` then it'll display that output instead.  Admittedly, we don't currently use that argument, but the fact that it's there and we have specific code in the constructor to handle setting it means that we shouldn't assume that it's always set to ```'default'```.

nengo_gui/components/pointer.py
Seanny123: I tried it with a normal clean-up memory and it worked fine, but today I realised that it didn't work with an associative memory where the input_keys don't match the output_keys.

tcstewar: Ah, good point.  Yup, I'd remove that from the whitelist.

Seanny123:
Ready for review

Seanny123:
Hey Terry, don't know if you missed my last ping, but this is the last bit that needs review before we can make a Nengo GUI release.

nengo_gui/components/pointer.py
```diff
@@ -16,17 +15,35 @@ def __init__(self, obj, **kwargs):
         # the semantic pointer value as set by the user in the GUI
         # a value of 'None' means do not override
         self.override_target = None
-        self.vocab_in = obj.inputs[self.target][1]
+        loop_in_whitelist = [spa.Buffer, spa.Memory, spa.State]
```
tcstewar: There should be a comment indicating what this whitelist is, and what conditions there are if someone were to add something new onto the whitelist (and also why we have this whitelist in the first place)

nengo_gui/components/pointer.py
```diff
@@ -16,17 +15,35 @@ def __init__(self, obj, **kwargs):
...
+
+        self.vocab_in = None
+        if type(obj) in loop_in_whitelist and self.target == 'default':
+            self.vocab_in = obj.inputs[self.target][1]
```
tcstewar: Given the whitelist, ```self.vocab_in``` is always identical to ```self.vocab_out```, so this variable isn't needed.

nengo_gui/components/pointer.py
```diff
@@ -16,17 +15,35 @@ def __init__(self, obj, **kwargs):
...
+            self.loop_in = True
+            self.size_out = self.vocab_in.dimensions
```
tcstewar: Due to the comment above, we don't need ```size_out``` either, as we can just always use ```self.vocab_out.dimensions```.

tcstewar:
Just to check, what was your reasoning for removing AssociativeMemory from the whitelist?  I'm not sure I disagree with that approach, but I'm just making sure you didn't accidentally do it...  :)

Seanny123:
I removed it on purpose. The approach of looping-in when there was a
different vocab on the input and ouput, even if the connection between them
tried to transform between the two vocabs, didn't work, so I removed it
from the whitelist.

tcstewar:
> I removed it on purpose. The approach of looping-in when there was a
> different vocab on the input and ouput, even if the connection between them
> tried to transform between the two vocabs, didn't work, so I removed it
> from the whitelist.

Fair enough.  I assume that also means that you felt that checking for the (common) use-case where the input vocab and output vocab are the same wasn't worth it?

Seanny123:
Exactly.

Seanny123:
Fixed all your comments. If you want to explore the network for yourself, this is what I tested with.
```
import nengo
from nengo import spa

D = 32
vo = spa.Vocabulary(D)
vo.parse("A + B + C")
model = spa.SPA(vocabs=[vo])


with model:
    model.in_A = spa.State(D)
    model.in_B = spa.State(D)
    model.bind = spa.Bind(D, invert_b=True)
    model.ass = spa.AssociativeMemory(vo)
    model.out = spa.State(D)
    model.cortical = spa.Actions(
        'bind_A = in_A',
        'bind_B = in_B',
        'ass = bind',
        'out = ass'
    )
    model.c_act = spa.Cortical(model.cortical)
```

nengo_gui/components/pointer.py
```diff
@@ -16,17 +15,31 @@ def __init__(self, obj, **kwargs):
...
         # a value of 'None' means do not override
         self.override_target = None
-        self.vocab_in = obj.inputs[self.target][1]
+        # The white list indicates the networks whose user-defined
+        # over-ride value can be inserted on the input
+        # thus the value loops in from the output to the input.
+        # All other networks have their value inserted on the output
+        self.loop_in_whitelist = [spa.Buffer, spa.Memory, spa.State]
+
+        self.obj = obj
```
tcstewar: this is already being set in the parent class

nengo_gui/components/pointer.py
```diff
@@ -16,17 +15,31 @@ def __init__(self, obj, **kwargs):
         # the semantic pointer value as set by the user in the GUI
         # a value of 'None' means do not override
         self.override_target = None
-        self.vocab_in = obj.inputs[self.target][1]
+        # The white list indicates the networks whose user-defined
+        # over-ride value can be inserted on the input
+        # thus the value loops in from the output to the input.
+        # All other networks have their value inserted on the output
```
tcstewar: Add in a mention that the whole point of looping in on the input is so that the actual neural activity of this module will be changed by setting its value, rather than just the output of the module.

nengo_gui/components/pointer.py
```diff
@@ -16,17 +15,31 @@ def __init__(self, obj, **kwargs):
...
         with page.model:
             output = self.obj.outputs[self.target][0]
-            input = self.obj.inputs[self.target][0]
             self.node = nengo.Node(self.gather_data,
                                    size_in=self.vocab_out.dimensions,
-                                   size_out=self.vocab_in.dimensions)
+                                   size_out=self.vocab_out.dimensions)
             self.conn1 = nengo.Connection(output, self.node, synapse=0.01)
-            self.conn2 = nengo.Connection(self.node, input, synapse=0.01)
+            if(type(self.obj) in self.loop_in_whitelist
```
tcstewar: spacing

Seanny123:
Fixed most recent comments

nengo_gui/components/pointer.py
```diff
@@ -1,32 +1,48 @@
...
             output = self.obj.outputs[self.target][0]
-            input = self.obj.inputs[self.target][0]
             self.node = nengo.Node(self.gather_data,
                                    size_in=self.vocab_out.dimensions,
-                                   size_out=self.vocab_in.dimensions)
+                                   size_out=self.vocab_out.dimensions)
             self.conn1 = nengo.Connection(output, self.node, synapse=0.01)
-            self.conn2 = nengo.Connection(self.node, input, synapse=0.01)
+            if (type(self.obj) in self.loop_in_whitelist
+                and self.target == 'default'):
```
tcstewar: This line does not distinguish itself from the indented line below, making it very hard to parse.

Note that this is exactly the one annoying case in Python indenting (an if statement with a parenthesis continuation, meaning that the natural amount you'd want to indent the second part of the first line is exactly four spaces).  I normally just rewrite the logic a bit to avoid having to do the line continuation.

Seanny123: Now that you've reminded me, I do recall you mentioning this before. I've
switched the logic and will remember this from now on.

tcstewar:
Looks good to me!

PR #567, opened 2015-10-04
Create plots for the input and the output of the SPA Basal Ganglia

### LINKED ISSUE #411: Custom plots for Thalamus and Basal Ganglia outputs
This came up in Terry's SPA tutorial on Friday. One of the most helpful things when debugging a SPA network is seeing which actions are being selected by the Basal Ganglia and Thalamus. At this point when plotting the values, you get the standard plot with no labels. It's hard to translate those colours into selected values. A legend or maybe an entirely new plot would be helpful.

Seanny123:
Also fixes a bug where the jQuery selector for finding the largest label for proper legend SVG sizing needed to be adjusted.

Partially resolves #411

### DIFF
--- a/nengo_gui/components/value.py
+++ b/nengo_gui/components/value.py
@@ -14,7 +14,7 @@ class Value(Component):
                          show_legend=False, legend_labels=[],
                          **Component.config_defaults)
 
-    def __init__(self, obj):
+    def __init__(self, obj, n_lines=0):
         super(Value, self).__init__()
         # the object whose decoded value should be displayed
         self.obj = obj
@@ -23,7 +23,11 @@ class Value(Component):
         self.data = []
 
         # the number of data values to send
-        self.n_lines = int(obj.size_out)
+        self.n_lines = 0
+        if n_lines == 0:
+            self.n_lines = int(obj.size_out)
+        else:
+            self.n_lines = n_lines
 
         # the binary data format to sent in.  In this case, it is a list of
         # floats, with the first float being the time stamp and the rest
@@ -40,7 +44,7 @@ class Value(Component):
         # data we want to show while the model is running.
         with page.model:
             self.node = nengo.Node(self.gather_data,
-                                   size_in=self.obj.size_out)
+                                   size_in=self.n_lines)
             self.conn = nengo.Connection(self.obj, self.node, synapse=0.01)
 
     def remove_nengo_objects(self, page):
--- a/nengo_gui/components/value.py
+++ b/nengo_gui/components/value.py
@@ -10,8 +10,8 @@ class Value(Component):
     """The server-side system for a Value plot."""
 
     # the parameters to be stored in the .cfg file
-    config_defaults = dict(max_value=1,
-                         min_value=-1, 
+    config_defaults = dict(max_value=1, min_value=-1,
+                         show_legend=False, legend_labels=[],
                          **Component.config_defaults)
 
     def __init__(self, obj):
--- a/nengo_gui/static/components/netgraph_item.js
+++ b/nengo_gui/static/components/netgraph_item.js
@@ -362,6 +362,7 @@ Nengo.NetGraphItem.prototype.generate_menu = function () {
         items.push(['Semantic pointer plot',
             function() {self.create_graph('SpaSimilarity', self.sp_targets[0]);}])
     }
+    // TODO: Enable input and output value plots for basal ganglia network
     items.push(['Details ...', function() {self.create_modal();}]);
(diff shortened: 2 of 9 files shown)


tcstewar:
Looks like you forgot to include ```components/bg_plot.py```  and possibly a ```.js``` file as well...  :)

Seanny123:
:blush: fixed it now

Seanny123:
This pull request is ready for review, however a few caveats before you dig in:
- It both adds legends to value plots and creates bg plots. Should this be split up into two separate pull requests?
- The bg output plot is still present. I remember talking during a meeting about removing it and/or replacing it with something else. Can that be part of another pull request or does it need to be a part of this one?

Seanny123:
Hmmm... Doesn't seem to be responding to edited actions on reload.

Seanny123:
Now responds to edits, but just gives the default values labels, which isn't good.

Seanny123:
This pull request is ready for review again!

tcstewar:
Here's an interesting bug with how the labels are generated:

- start with examples/tutorial/21-spa-sequence.py
- show the BG input plot (or output plot).  This will have labels based on the condition part of each rule.
- comment out one rule that's NOT the last one.  So something like:

```
    actions = spa.Actions(
        'dot(memory, A) --> memory=B',
        'dot(memory, B) --> memory=C',
        #'dot(memory, C) --> memory=D',
        'dot(memory, D) --> memory=E',
        'dot(memory, E) --> memory=A',
        )
```

The remaining lines will be labelled ```dot(memory, A), dot(memory, B), dot(memory, C), dot(memory, D)``` rather than ```dot(memory, A), dot(memory, B), dot(memory, D), dot(memory, E)```

nengo_gui/components/bg_plot.py
```diff
@@ -0,0 +1,68 @@
...
+        # default legends to show
+        self.def_legend_labels = args["legend_labels"]
+
+        # the item to connect to
+        self.probe_target = args["probe_target"]
+
+        self.label = "bg " + self.probe_target
+
+    def attach(self, page, config, uid):
+        super(Value, self).attach(page, config, uid)
```
tcstewar: If this isn't doing anything other than calling the constructor, I'm not sure it should be defined at all

nengo_gui/components/value.py
```diff
@@ -95,6 +95,8 @@ def default_output(obj):
         if isinstance(obj, spa.module.Module):
             if 'default' in obj.outputs.keys():
                 output = obj.outputs['default'][0]
+            elif isinstance(obj, spa.BasalGanglia):
+                output = obj.output
```
tcstewar: I think this should be the first thing checked in the if, just in case in the future someone defines a default output for the BasalGanglia and is surprised that the plots suddenly stop working.

nengo_gui/static/components/netgraph_item.js
```diff
@@ -32,7 +32,13 @@ Nengo.NetGraphItem = function(ng, info, minimap, mini_item) {
         this.g_networks = ng.g_networks_mini;
         this.g_items = ng.g_items_mini;
     }
-    
+
+    // SPA network specific parameter
```
tcstewar: Not sure what this means

nengo_gui/static/components/netgraph_item.js
```diff
@@ -32,7 +32,13 @@ Nengo.NetGraphItem = function(ng, info, minimap, mini_item) {
...
+    this.sp_targets = info.sp_targets;
+    // Basal Ganglia network specific parameters
```
tcstewar: Also not sure what this means.

nengo_gui/components/bg_plot.py
Seanny123: The difference between this calling of the constructor and not defining it at all, is that in this call, I don't assign a label. If this were not defined, a label, which I do not want, would be assigned.

Seanny123:
I figure out the bug causing the weird label stuff.

The legend labels are being saved in the config and being reloaded regardless of their relevance, because they're saved as a list and not as a hash. So items of at the end of the list are deleted regardless of the position of the inputs are removed. To fix this, they should be saved as a hash where bg inputs are associated to specific legend labels. To accomplish this, I need to create a new class for bg-plot in JavaScript and I have to extract the legend management code from Value.js. Consequently, I won't be getting around to this for a while.

tcstewar:
> Consequently, I won't be getting around to this for a while.

No problem!  We can put this off till 0.3...   :)

PR #568, opened 2015-10-05
Fixed calculation of simulation speed

tcstewar:
The speed calculation in the bottom-left is often hysterically wrong (most noticeably when running nengo_spinnaker).  This fixes the calculation and means that when using nengo_spinnaker the speed very clearly is at 1.00x realtime.  :)

### DIFF
--- a/nengo_gui/components/sim_control.py
+++ b/nengo_gui/components/sim_control.py
@@ -48,16 +48,17 @@ class SimControl(Component):
         self.page.finish()
 
     def control(self, t):
+        self.actual_model_dt = t - self.time
         self.time = t
         self.sim_ticks += 1
 
-        now = time.time()
+        now = time.clock()
         if self.last_tick is not None:
             dt = now - self.last_tick
             if dt == 0:
                 self.skipped += 1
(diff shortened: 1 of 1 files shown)


nengo_gui/components/sim_control.py
```diff
@@ -48,16 +48,17 @@ def finish(self):
         self.page.finish()
 
     def control(self, t):
+        self.actual_model_dt = t - self.time
         self.time = t
         self.sim_ticks += 1
 
-        now = time.time()
+        now = time.clock()
```
Seanny123: Apparently, `time.clock()` [is deprecated in Python 3.3](http://stackoverflow.com/questions/85451/python-time-clock-vs-time-time-accuracy)

tcstewar: I see....  looks like ```timeit.default_timer()``` is the way to go...  I'll make that change.

tcstewar:
Updated to not use the deprecated ```time.clock()``` call.

Seanny123:
LGTM.

PR #570, opened 2015-10-05
Fix page initialization

tcstewar:
This ensures nengo_gui doesn't crash if Page.execute()
is never called (as is the case when manually starting
up the gui with a pre-defined model)


Seanny123:
LGTM

PR #572, opened 2015-10-13
Fix completely broken page update due to #570

tcstewar:
I broke nengo_gui.  It was in a state where it would never notice that it had successfully executed a script, other than the first time it ran.


tcstewar:
Since this is a show-stopper bug, and it's such a small fix, I'm merging this right away.

Seanny123:
*facepalm*

PR #574, opened 2015-10-14
Added ability to change color palettes

Seanny123:
Added a modal dialogue to choose between a variety of palettes


Seanny123:
@tcstewar @jgosmann if any of you are in the reviewing mood, start here

Seanny123:
Nvm, in my efforts to rebase, I accidentally lost some code.

Seanny123:
Ready for review again if someone is feeling ambitious/bored.

Seanny123:
Current issues:
- Label setting isn't working totally as intended
- Label setting should have help text
- Palette selection should show the palette to the right of the choice

nengo_gui/components/bg_plot.py
```diff
@@ -0,0 +1,52 @@
...
+            else:
+                self.conn = nengo.Connection(self.obj.output, self.node, synapse=0.01)
+
+    def javascript(self):
+        # generate the javascript that will create the client-side object
+        info = dict(uid=id(self), label=self.label, 
+                    n_lines=self.n_lines, synapse=0)
+
+        if getattr(self.config, "legend_labels") == []:
+            json = self.javascript_config(info, override={"legend_labels":self.def_legend_labels})
```
tcstewar: ????   Why not just do

```
info = dict(uid=id(self), label=self.label, 
                   n_lines=self.n_lines, synapse=0,
                   legend_labels=self.def_legend_labels)
```

and get rid of this whole override thing?

nengo_gui/components/component.py
```diff
@@ -82,13 +82,16 @@ def remove_nengo_objects(self, page):
...
         """Convert the nengo.Config information into javascript.
         
         This is needed so we can send that config information to the client.
         """
         for attr in self.config._clsparams.params:
-            cfg[attr] = getattr(self.config, attr)
+            if attr not in override:
+                cfg[attr] = getattr(self.config, attr)
+            else:
+                cfg[attr] = override[attr]
```
tcstewar: See the comment in bgplot.py; I'm pretty sure that this override thing isn't needed at all, since you can just directly put that stuff in the dictionary you're creating and passing in here.

bjkomer:
I'm seeing the basal ganglia plots disappear when making a change in the editor, they might not be getting saved properly in the config.

nengo_gui/components/bg_plot.py
Seanny123: Because the behaviour of `javascript_config` is to override everything with the defaults. It doesn't look to see if the defaults have been set to anything else. Is that it's intended behaviour or should the code be changed?

tcstewar: I don't think I understand...   all javascript_config does is go through the config entry for the item and just into a json output:

```python
        for attr in self.config._clsparams.params:
            cfg[attr] = getattr(self.config, attr)
        return json.dumps(cfg)
```

what do you mean by overriding everything with the defaults?

Seanny123: I'll break it down why it over-writes by going through the above code with the example of `legend_labels` being set in `dict` argument to `["A", "B", "C"]`.

`cfg` is the `dict` I've passed in as an argument. The for-loop takes values from the default config, which is `self.config._clsparams` and adds it to `cfg`. The default config contains `legend_labels = []`, so when the for-loop runs, regardless of what the value that `cfg` has for `legend_labels`, it will be over-written by the default config. That is why `["A","B","C"]` will be over-written to `[]`. Does that make sense now?

Seanny123: It's basically equivalent to running this code:
```
cfg = dict(label="pants", n_lines=4, legend_labels=["A", "B", "C"])

default_conf = dict(min=-1, max=1, legend_labels=[])

for attr in default_conf.keys():
    cfg[attr] = default_conf[attr]
print(cfg)
```

PR #579, opened 2015-10-16
Added check for exiting items in NameFinder

### LINKED ISSUE #563: Weird connection error and behaviour when modifying Thalamus
I'm still investigating how to isolate this example, but I'm going to document how I found it here for now.

Using the code and config file found in [this Gist](https://gist.github.com/Seanny123/8b7b38ec9a7d39c16267). Expand the network called `channel_4_memory` in the Thalamus network.

By toggling line line 17 of the code between:
`'dot(memory, E) --> memory=speech'`
and
`'dot(memory, E) --> memory=speech*vision'`

The connection won't be made due to multiple JavaScript errors. The workaround is to just collapse and re-expand the network.

Seanny123:
This was preformed for lists of objects, but not for single objects.

Amazingly, this resolves #563

### DIFF
--- a/nengo_gui/namefinder.py
+++ b/nengo_gui/namefinder.py
@@ -31,7 +31,8 @@ class NameFinder(object):
                             n = '%s.%s[%d]' % (net_name, k, i)
                             self.known_name[obj] = n
                 elif isinstance(v, classes):
-                    self.known_name[v] = '%s.%s' % (net_name, k)
+                    if v not in self.known_name:
+                        self.known_name[v] = '%s.%s' % (net_name, k)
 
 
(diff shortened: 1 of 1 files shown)


tcstewar:
Argh!  Damnit, nice find.  :)

tcstewar:
After getting a chance to do some testing, this fix also resolves a few weirdnesses I'd never tracked down before.  Thank you for finding it!

However, it doesn't seem to resolve all of #563.  The initial error still seems to happen.  It does resolve the connection rendering bug that @ikajic reported, though.  Can you check again whether it resolves it for you?

PR #566, opened 2015-10-04
Updating docs and fixing typos

Seanny123:
This is the pull request I'm going to use for fixing little typos and adding missing docs. That way, I don't clog up my other pull requests with little fixes that drive reviewers crazy. Should be merged before milestone.

Addresses #545 and #558

PR #581, opened 2015-10-20
Added plot for Compare networks

### LINKED ISSUE #559: Creating semantic pointer cloud on a spa.Compare network causes error

Seanny123:
Before regular SPA options were given which would crash the simulation if chosen.

Resolves #559

### DIFF
--- a/nengo_gui/components/__init__.py
+++ b/nengo_gui/components/__init__.py
@@ -11,6 +11,7 @@ from .ace_editor import AceEditor
 from .spa_similarity import SpaSimilarity
 from .htmlview import HTMLView
 from .spike_grid import SpikeGrid
+from .comp_plot import CompPlot
 
 # Old versions of the .cfg files used Templates which had slightly different
 # names than the Components currently use.  This code allows us to
--- a/nengo_gui/components/__init__.py
+++ b/nengo_gui/components/__init__.py
@@ -11,7 +11,6 @@ from .ace_editor import AceEditor
 from .spa_similarity import SpaSimilarity
 from .htmlview import HTMLView
 from .spike_grid import SpikeGrid
-from .comp_plot import CompPlot
 
 # Old versions of the .cfg files used Templates which had slightly different
 # names than the Components currently use.  This code allows us to
--- a/nengo_gui/components/netgraph.py
+++ b/nengo_gui/components/netgraph.py
@@ -9,6 +9,7 @@ from nengo import spa
 import json
 
 from nengo_gui.components.component import Component
+from nengo_gui.components.value import Value
 from nengo_gui.modal_js import infomodal
 import nengo_gui.user_action
 import nengo_gui.layout
@@ -470,8 +471,8 @@ class NetGraph(Component):
         elif isinstance(obj, nengo.Ensemble):
             info['dimensions'] = int(obj.size_out)
             info['n_neurons'] = int(obj.n_neurons)
-        elif isinstance(obj, spa.Compare):
-            info['scalar_out'] = True
+        elif Value.default_output(obj):
+            info['default_output'] = True
 
         info['sp_targets'] = (
             nengo_gui.components.pointer.Pointer.applicable_targets(obj))
--- a/nengo_gui/components/netgraph.py
+++ b/nengo_gui/components/netgraph.py
@@ -471,7 +471,7 @@ class NetGraph(Component):
         elif isinstance(obj, nengo.Ensemble):
             info['dimensions'] = int(obj.size_out)
             info['n_neurons'] = int(obj.n_neurons)
-        elif Value.default_output(obj):
+        elif Value.default_output(obj) is not None:
             info['default_output'] = True
 
         info['sp_targets'] = (
--- a/nengo_gui/components/netgraph.py
+++ b/nengo_gui/components/netgraph.py
@@ -5,6 +5,7 @@ import collections
 import threading
 
 import nengo
+from nengo import spa
 import json
 
 from nengo_gui.components.component import Component
@@ -469,8 +470,12 @@ class NetGraph(Component):
         elif isinstance(obj, nengo.Ensemble):
             info['dimensions'] = int(obj.size_out)
             info['n_neurons'] = int(obj.n_neurons)
+        elif isinstance(obj, spa.Compare):
+            info['scalar_out'] = True
+
         info['sp_targets'] = (
             nengo_gui.components.pointer.Pointer.applicable_targets(obj))
+
         return info
 
     def send_pan_and_zoom(self, client):
--- a/nengo_gui/components/pointer.py
+++ b/nengo_gui/components/pointer.py
@@ -96,6 +96,11 @@ class Pointer(SpaPlot):
     @staticmethod
     def applicable_targets(obj):
         if isinstance(obj, Module):
-            return list(obj.outputs.keys())
+            targets = []
+            output_names = obj.outputs.keys()
+            for i, output in enumerate(obj.outputs.values()):
+                if output[0].size_out > 1:
+                    targets.append(output_names[i])
+            return targets
         else:
             return []
--- a/nengo_gui/components/pointer.py
+++ b/nengo_gui/components/pointer.py
@@ -95,12 +95,9 @@ class Pointer(SpaPlot):
 
     @staticmethod
     def applicable_targets(obj):
+        targets = []
         if isinstance(obj, Module):
-            targets = []
-            output_names = obj.outputs.keys()
-            for i, output in enumerate(obj.outputs.values()):
-                if output[0].size_out > 1:
-                    targets.append(output_names[i])
-            return targets
-        else:
-            return []
+            for target_name, (obj, vocab) in obj.outputs.items():
+                if vocab is not None:
+                    targets.append(target_name)
+        return targets
--- a/nengo_gui/components/tests/test_pointer.py
+++ b/nengo_gui/components/tests/test_pointer.py
@@ -1,0 +1,17 @@
+import pytest
+import nengo
+from nengo import spa
+
+from nengo_gui.components.pointer import Pointer
+
+def test_applicable_targets():
+    # Only return non-scalar targets
+    model = spa.SPA()
+
+    with model:
+        model.comp = spa.Compare(4)
+        assert Pointer.applicable_targets(model.comp) == []
+
+        model.state = spa.State(4)
+        res = Pointer.applicable_targets(model.state)
+        assert res == model.state.outputs.keys()
--- a/nengo_gui/components/tests/test_pointer.py
+++ b/nengo_gui/components/tests/test_pointer.py
@@ -14,4 +14,4 @@ def test_applicable_targets():
 
         model.state = spa.State(4)
         res = Pointer.applicable_targets(model.state)
-        assert res == model.state.outputs.keys()
+        assert res == list(model.state.outputs.keys())
--- a/nengo_gui/components/value.py
+++ b/nengo_gui/components/value.py
@@ -23,13 +23,12 @@ class Value(Component):
 
         # grab the output of the object
         self.output = obj
-        default_out = self.default_output(self.obj)
-        if(default_out):
+        default_out = Value.default_output(self.obj)
+        if default_out is not None:
             self.output = default_out
 
-
         # the number of data values to send
-        self.n_lines = self.output.size_out
+        self.n_lines = int(self.output.size_out)
 
         # the binary data format to sent in.  In this case, it is a list of
         # floats, with the first float being the time stamp and the rest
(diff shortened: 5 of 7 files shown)


nengo_gui/components/__init__.py
```diff
@@ -11,6 +11,7 @@
 from .spa_similarity import SpaSimilarity
 from .htmlview import HTMLView
 from .spike_grid import SpikeGrid
+from .comp_plot import CompPlot
```
tcstewar: I assume this file was missed when you checked things in?

Seanny123: ARRRRGHHH.

Fixed it.

Seanny123:
Terry thinks sub-classing the value network is a bad idea and should be just replaced with an if-statement checking for SPA related networks. I'm thinking that maybe just checking for an "output" node might not be a bad idea and would allow for the accommodation of Ensemble Arrays and other things more easily.

tcstewar:
> I'm thinking that maybe just checking for an "output" node might not be a bad idea and would allow for the accommodation of Ensemble Arrays and other things more easily.

Seems fair to me.  I think logic like this might work:

- if self.obj is a Node or Ensemble, use that
- if self.obj is a spa.Module and it has self.outputs['default'] set, use that
- if self.obj is a Network and it has self.output and self.output is a Node or Ensemble, use that

I guess technically we don't need the middle step in there, but it seems right to me to have the nengo_gui aware of the output default system, seeing as how it already exists.

Seanny123:
Fixed the implementation as discussed.

Seanny123:
@bjkomer this pull request is basic enough that I think you should be able to understand it. Basically, I enabled `netgraph.py` to tell `netgraph_item.js` that the `spa.Compare` object should have it's own special set of menus. Then I told `value.py` how to deal with a `spa.Compare` object.

nengo_gui/components/value.py
```diff
@@ -14,7 +15,7 @@ class Value(Component):
                          min_value=-1, 
                          **Component.config_defaults)
 
-    def __init__(self, obj):
+    def __init__(self, obj, n_lines=0):
```
tcstewar: Why is n_lines being added as an option here?  Is this used anywhere?

nengo_gui/components/value.py
```diff
@@ -22,8 +23,20 @@ def __init__(self, obj):
...
+        self.output = self.obj
+        if isinstance(self.obj, spa.module.Module):
+            self.output = self.obj.outputs['default'][0]
+        elif isinstance(self.obj, nengo.network.Network):
+            if self.obj.output != None:
+                self.output = self.object.output
+
         # the number of data values to send
-        self.n_lines = int(obj.size_out)
+        self.n_lines = 0
```
tcstewar: n_lines doesn't need to be set multiple times like this

nengo_gui/components/value.py
```diff
@@ -22,8 +23,20 @@ def __init__(self, obj):
         # the pending data to be sent to the client
         self.data = []
 
+        # grab the output of the object
+        self.output = self.obj
+        if isinstance(self.obj, spa.module.Module):
+            self.output = self.obj.outputs['default'][0]
+        elif isinstance(self.obj, nengo.network.Network):
+            if self.obj.output != None:
```
tcstewar: Won't this crash if ```self.obj``` is a Network that doesn't have an attribute ```output```?  Did you mean ```if hasattr(self.obj, 'output'):```?

nengo_gui/components/value.py
```diff
@@ -22,8 +23,20 @@ def __init__(self, obj):
         # the pending data to be sent to the client
         self.data = []
 
+        # grab the output of the object
+        self.output = self.obj
+        if isinstance(self.obj, spa.module.Module):
+            self.output = self.obj.outputs['default'][0]
```
tcstewar: This will crash if the module doesn't have a ```default``` output (there aren't any right now that don't but we may as well check for it now)

nengo_gui/components/netgraph.py
```diff
@@ -469,8 +470,12 @@ def get_extra_info(self, obj):
         elif isinstance(obj, nengo.Ensemble):
             info['dimensions'] = int(obj.size_out)
             info['n_neurons'] = int(obj.n_neurons)
+        elif isinstance(obj, spa.Compare):
+            info['scalar_out'] = True
```
tcstewar: I'm okay with this special-case code for now, but we should add a note here that if we start having a lot of cases added, we should make this more generic.  For example, we could just do the same check as in value.py:

```
elif ((isinstance(obj, spa.module.Module) and obj.outputs.haskey('default')) or
      (isinstance(obj, spa.Network) and hasattr(obj, 'output'))):
             info['has_default_output'] = True
```

Hmm, and I'd also say that ```has_default_output``` may be clearer than ```scalar_out```, especially since there's nothing in this version of the code that forces it to be a scalar.....

tcstewar: Hmm, but on second thought, that more generic approach would cause ```spa.State``` to also have this line, which we don't want.  So let's stick with this special-case for now and worry about it later.   I still think the name of the flag should be changed, though.  :)

nengo_gui/components/value.py
```diff
@@ -22,8 +23,13 @@ def __init__(self, obj):
         # the pending data to be sent to the client
         self.data = []
 
-        # the number of data values to send
-        self.n_lines = int(obj.size_out)
+        # grab the output of the object
+        self.output = self.obj
+        if isinstance(self.obj, spa.module.Module):
+            self.output = self.obj.outputs['default'][0]
```
tcstewar: My earlier comment on this line seems to have disappeared; here it is again:

> This will crash if the module doesn't have a default output (there aren't any right now that don't but we may as well check for it now)

nengo_gui/components/value.py
```diff
@@ -68,7 +74,8 @@ def update_client(self, client):
     def javascript(self):
         # generate the javascript that will create the client-side object
         info = dict(uid=id(self), label=self.label,
-                    n_lines=self.n_lines, synapse=0)
+                        n_lines=self.n_lines, synapse=0)
```
tcstewar: spacing.  :)

nengo_gui/static/components/netgraph_item.js
```diff
@@ -357,10 +358,14 @@ Nengo.NetGraphItem.prototype.generate_menu = function () {
         }
     }
     if (this.sp_targets.length > 0) {
-        items.push(['Semantic pointer cloud',
-                    function() {self.create_graph('Pointer', self.sp_targets[0]);}])
-        items.push(['Semantic pointer plot',
-            function() {self.create_graph('SpaSimilarity', self.sp_targets[0]);}])
+        if(!this.scalar_out) {
```
tcstewar: I'm uncomfortable with this check.  Whatever is generating sp_targets shouldn't be giving us invalid data in the first place.

Seanny123: So if sp_targets didn't return anything for scalar outputs, this check wouldn't be needed. Would that be an acceptable solution?

tcstewar: Exactly, since those aren't sp_targets as they aren't semantic pointers.   Although rather than checking for scalars, I'd do the check so that it doesn't include anything where ```vocab=None```, since if there's no vocab, it's not a semantic pointer.  (we might have future situations like ```spa.Compare``` but where there's a 2-dimensional output, for example)

Seanny123:
Made all the changes discussed. Ready for review again.

nengo_gui/components/value.py
```diff
@@ -76,3 +83,15 @@ def code_python_args(self, uids):
         # generate the list of strings for the .cfg file to save this Component
         # (this is the text that would be passed in to the constructor)
         return [uids[self.obj]]
+
+    @staticmethod
+    def default_output(obj):
+        """Find default output object for the input object if it exists"""
+        output = False
```
tcstewar: This should be ```None```, not ```False```

nengo_gui/components/netgraph.py
```diff
@@ -469,8 +471,12 @@ def get_extra_info(self, obj):
         elif isinstance(obj, nengo.Ensemble):
             info['dimensions'] = int(obj.size_out)
             info['n_neurons'] = int(obj.n_neurons)
+        elif Value.default_output(obj):
```
tcstewar: I think this is clearer as ```elif Value.default_output(obj) is not None:```

nengo_gui/components/pointer.py
```diff
@@ -96,6 +96,11 @@ def message(self, msg):
     @staticmethod
     def applicable_targets(obj):
         if isinstance(obj, Module):
-            return list(obj.outputs.keys())
+            targets = []
+            output_names = obj.outputs.keys()
+            for i, output in enumerate(obj.outputs.values()):
+                if output[0].size_out > 1:
+                    targets.append(output_names[i])
```
tcstewar: I'm not sure about this ```size_out > 1``` check.  ```output[1] is not None``` seems much more robust to me.

tcstewar: This enumerate logic also feels a bit hard to follow.  How about

```python
def applicable_targets(obj):
    targets = []
    if isinstance(obj, Module):
        for target_name, (obj, vocab) in obj.outputs.items():
            targets.append(target_name)
    return targets
```

or even

```python
def applicable_targets(obj):
    if isinstance(obj, Module):
        for target_name, (obj, vocab) in obj.outputs.items():
             yield target_name
```

nengo_gui/components/value.py
```diff
@@ -22,8 +21,15 @@ def __init__(self, obj):
         # the pending data to be sent to the client
         self.data = []
 
+        # grab the output of the object
+        self.output = obj
+        default_out = self.default_output(self.obj)
```
tcstewar: I'd prefer this to be called as ```Value.default_output(self.obj)```

nengo_gui/components/value.py
```diff
@@ -22,8 +21,15 @@ def __init__(self, obj):
...
+        if(default_out):
```
tcstewar: spacing and parentheses

nengo_gui/components/value.py
```diff
@@ -22,8 +21,15 @@ def __init__(self, obj):
...
+            self.output = default_out
+
+
         # the number of data values to send
-        self.n_lines = int(obj.size_out)
+        self.n_lines = self.output.size_out
```
tcstewar: Why get rid of the ```int``` here?  That's needed because some versions of numpy give size_outs that can't be serialized.....

nengo_gui/static/components/netgraph_item.js
```diff
@@ -353,14 +358,14 @@ Nengo.NetGraphItem.prototype.generate_menu = function () {
             items.push(['XY-value', function() {self.create_graph('XYValue');}]);
         }
         if (this.html_node) {
-            items.push(['HTML', function() {self.create_graph('HTMLView');}])
+            items.push(['HTML', function() {self.create_graph('HTMLView');}]);
```
tcstewar: These changes and the ones below it are not part of this PR; they're part of #566

Seanny123:
Fixed all the syntax errors. I've also surrendered myself to the merge conflicts that merging this branch will cause.

nengo_gui/components/tests/test_pointer.py
```diff
@@ -0,0 +1,17 @@
...
+    # Only return non-scalar targets
+    model = spa.SPA()
+
+    with model:
+        model.comp = spa.Compare(4)
+        assert Pointer.applicable_targets(model.comp) == []
+
+        model.state = spa.State(4)
+        res = Pointer.applicable_targets(model.state)
+        assert res == model.state.outputs.keys()
```
tcstewar: this should be ```list(model.state.outputs.keys())``` so that it passes on Python 3 (```keys()``` in Python 3 returns an iterator, not a list)

Seanny123: Fixed.

tcstewar:
> I've also surrendered myself to the merge conflicts that merging this branch will cause.

I suppose....  :) but please try to get into the habit of doing a ```git diff <filename.py>``` to look at your changes before committing them, and doing a ```git commit -p <filename.py>``` whenever there are changes that aren't part of whatever you're working on....  :)

PR #587, opened 2015-10-20
Rebuild Components correctly for collapsed Networks

### LINKED ISSUE #586: Graphs don't get rebuilt if their components aren't shown in the NetGraph
When parameters change on a graph (the number of lines, for example), they get rebuilt.  However, this code is currently dependent on the component that is being graphed currently being shown.  If it is inside a Network that has been collapsed, it never gets rebuilt.  

When this happens to a Value plot, it starts getting very confused, since it is expecting N data values, but starts being given M data values.  This causes all sorts of weird display issues.

Example code to trigger this:

```python
import nengo
import numpy as np

D = 3

model = nengo.Network()
with model:
    subnet = nengo.Network()
    with subnet:
        node = nengo.Node(np.linspace(-.9, .9, D))
```

Plot the data in the node, collapse the subnet, and start varying D.  It gets very upset and the graph starts doing weird things.  Now expand the subnet, and it should work fine.

The problem is here:

https://github.com/nengo/nengo_gui/blob/master/nengo_gui/components/netgraph.py#L168

That's where we check if a Component should get rebuilt.  It does this by looking at all the things this component depends on and checking if they're in the ```rebuilt_objects``` list.  That's the list of Nodes, Ensembles, and Connections that have been changed (due to dimensionality, etc).  Unfortunately, that list is not exhaustive, as it only includes displayed items.

I think the solution is to keep track of both ```rebuilt_objects``` and ```non_rebuilt_objects```.  If a Component depends on something that is not in either list, then it should go and check whether that thing has changed.

tcstewar:
Fixes #586

### DIFF
--- a/nengo_gui/components/netgraph.py
+++ b/nengo_gui/components/netgraph.py
@@ -162,9 +162,41 @@ class NetGraph(Component):
         orphan_components = []
         rebuild_components = []
 
+        # items that are shown in components, but not currently displayed
+        #  in the NetGraph (i.e. things that are inside collapsed
+        #  Networks, but whose values are being shown in a graph)
+        collapsed_items = []
+
         removed_items = list(removed_uids.values())
         for c in self.page.components[:]:
             for item in c.code_python_args(old_default_labels):
+                if item not in self.uids.keys() and item not in collapsed_items:
+                    # this Component depends on an item inside a collapsed
+                    #  Network, so we need to check if that component has
+                    #  changed or been removed
+                    try:
+                        old_obj = eval(item, old_locals)
+                    except:
+                        continue
+                    try:
+                        new_obj = eval(item, self.page.locals)
+                    except:
+                        new_obj = None
+
+                    keep_object = True
+                    if new_obj is None:
+                        removed_items.append(item)
+                    elif not isinstance(new_obj, old_obj.__class__):
+                        rebuilt_objects.append(item)
+                    elif (self.get_extra_info(new_obj) != 
+                          self.get_extra_info(old_obj)):
+                        rebuilt_objects.append(item)
+
+                    # add this to the list of collapsed items, so we
+                    # don't recheck it if there's another Component that
+                    # also depends on this
+                    collapsed_items.append(item)
+
                 if item in rebuilt_objects:
                     self.to_be_sent.append(dict(type='delete_graph',
                                                 uid=c.original_id,
--- a/nengo_gui/components/netgraph.py
+++ b/nengo_gui/components/netgraph.py
@@ -76,7 +76,7 @@ class NetGraph(Component):
 
     def _reload(self, code=None):
 
-        old_locals = self.page.locals
+        old_locals = self.page.last_good_locals
         old_default_labels = self.page.default_labels
 
         if code is None:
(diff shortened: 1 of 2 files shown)


nengo_gui/components/netgraph.py
```diff
@@ -162,9 +162,41 @@ def _reload(self, code=None):
...
         for c in self.page.components[:]:
             for item in c.code_python_args(old_default_labels):
+                if item not in self.uids.keys() and item not in collapsed_items:
+                    # this Component depends on an item inside a collapsed
+                    #  Network, so we need to check if that component has
+                    #  changed or been removed
+                    try:
+                        old_obj = eval(item, old_locals)
+                    except:
+                        continue
```
Seanny123: Why would evaluating this cause an exception? If I understood why, I would feel more comfortable with behaviour of continuing regardless of exceptions.

Also, if you do continue after an exception, isn't that going to cause an error in line Line 188 where you do `not isinstance(new_obj, old_obj.__class__)` since `old_obj` won't exist?

tcstewar: > Also, if you do continue after an exception, isn't that going to cause an error in line Line 188 where you do not isinstance(new_obj, old_obj.__class__) since old_obj won't exist?

That's not what ```continue``` does.  You're thinking of ```pass```, which would cause the problem you're indicating.  ```continue``` causes python to go on to the next item in the ```for``` loop.

> Why would evaluating this cause an exception? If I understood why, I would feel more comfortable with behaviour of continuing regardless of exceptions.

I should definitely add a comment about this.  Basically, the list of things returned by ```c.code_python_args``` can contain things that aren't items.  It includes all the arguments that should be passed into the Component, including things like ```target='default'``` for the semantic pointer plots, for example.  Evaluating those will cause an Exception (since it's not a valid expression), so we should skip those.

That said, I should probably change this to be a bit more robust for different cases of things that might be in ```c.code_python_args```, as now that you've pointed it out I can think of a few cases that would break this logic (but they're all cases that we don't use at the moment).

tcstewar:
I've reworked the Component argument checking code to handle the generic case, so it no longer needs that ```continue``` hack (which was basically it giving up on anything that was too hard to parse).  

I also fixed a bug where it was using the locals from the immediately previous run of the script for comparison, rather than the previous *successful* run of the script (which is the last time the netgraph got updated, since it only gets updated when there's no errors).  That might have been causing all sorts of other weird bugs.....

nengo_gui/components/netgraph.py
```diff
@@ -162,9 +162,52 @@ def _reload(self, code=None):
...
+                    old_obj = eval(safe_eval % item, old_locals)
+
+                    try:
+                        new_obj = eval(safe_eval % item, self.page.locals)
+                    except:
+                        # the object this Component depends on no longer exists
+                        new_obj = None
+
+                    keep_object = True
+                    if new_obj is None:
```
Seanny123: Since this code is identical to the code in line 128, I'd like to throw the whole `keep_object` decision tree into a function.

tcstewar: It's not identical.  The code around 128 doesn't need the distinction between removed_items and rebuilt_items, so it doesn't do that.  It'd certainly be possible and a good idea to refactor this stuff in general, such that it'd be possible to have identical code here, but I think that's outside the scope of this bugfix.

That said, the keep_object=True line should be removed...  :)

Seanny123: Sounds good to me.

Seanny123:
LGTM.

PR #588, opened 2015-10-20
Update in tutorials: Renamed Memory and Buffer to State

ikajic:
I've replaced all occurrences `spa.Buffer` and `spa.Memory` with `spa.State`, including some other minor changes to the code, docstrings and fixed typos. Also, the last example `25-spa-parse.py` is now very similar to the second example in the SPA tutorial (still it's not the same). @tcstewar could you take a look at this? Thanks!


nengo_gui/examples/tutorial/02-two-neurons.py
```diff
@@ -19,7 +19,7 @@
...
     ens = nengo.Ensemble(n_neurons=2, dimensions=1, seed=5)
     nengo.Connection(stim, ens)
-    
-    
+
+
     import time
-    slowdown = nengo.Node(lambda t: time.sleep(0.001)) 
\ No newline at end of file
+    slowdown = nengo.Node(lambda t: time.sleep(0.001))
```
Seanny123: Ideally, this line shouldn't exist, but until then, maybe add a comment?

Seanny123:
This all looks good to me.

nengo_gui/examples/tutorial/02-two-neurons.py
ikajic: This line is also there in examples 00 and 01. My assumption is that the code is shown in case anyone needs to use this if graphics rendering is ahead of the simulation. I could add a comment in `00-intro.py` and remove these lines from 01 and 02?

Seanny123: That makes sense to me

tcstewar: > This line is also there in examples 00 and 01. My assumption is that the code is shown in case anyone needs to use this if graphics rendering is ahead of the simulation. I could add a comment in 00-intro.py and remove these lines from 01 and 02?

I'd really prefer to remove them entirely.  Does any other computer have the problem you're seeing with these being too fast?  Does it happen on both Chrome and Firefox?  If this is a very rare problem I'd prefer to not add this complexity.

nengo_gui/examples/tutorial/20-spa-actions.py
```diff
@@ -20,18 +22,19 @@
...
+# to set the speech State to the vector for BARK ("speech=BARK").
 #
 # Try running the model.  Use "Set value" on the vision graph to set the
 # input to DOG.  The speech output should become BARK.  Try other inputs,
 # such as CAT or COW.  You can also try a combination, such as
 # RAT+0.5*COW.  The system will (almost always) select a single action
 # to perform, and that will be the action with the highest utility at that
-# moment.
+# moment. If there is no input, or the input contains just noise, the speech
+# State is set to zero ("speech=0").
```
tcstewar: I'm not sure about the "noise" comment.  How about "If there is no input, or the input is not sufficiently similar
to any of the four animals it knows, then the speech State is set to zero ("speech=0")"

nengo_gui/examples/tutorial/25-spa-parse.py
```diff
@@ -4,53 +4,67 @@
...
-# things at the same time by using the + operation.  Overall, this shows how
-# we can use a single input modality (vision) and treat the information
-# in different ways as appropriate.
-#
+# action looks for verbs (WRITE or SAY) and sends them into the memory for
+# verbs. The second action looks for nouns (YES or NO or HI or BYE or OK) and
+# sends them to the memory for nouns.  The noun stored in the noun memory binds
+# with the semantic pointer NOUN and the verb stored in the verb memory binds
+# with the semantic pointer VERB in the phrase state.  This binding is
+# implemented with cortical actions. Notice that the actions can looks for
```
tcstewar: "looks" should be "look"

nengo_gui/examples/tutorial/25-spa-parse.py
```diff
@@ -4,53 +4,67 @@
...
-    model.memory = spa.Memory(D, synapse=0.2)
-    model.speech = spa.Buffer(D)
-    model.hand = spa.Buffer(D)
+    # perceptual input
+    model.vision = spa.State(D)
+
+    # linguistic components
+    model.phrase = spa.State(D)
+    model.verb = spa.State(D, feedback=1, feedback_synapse=0.01)
+    model.noun = spa.State(D, feedback=1, feedback_synapse=0.01)
```
tcstewar: I'd expect this to be better with a ```feedback_synapse=0.1``` (the default).  Did that work for you?

nengo_gui/examples/tutorial/25-spa-parse.py
```diff
@@ -4,53 +4,67 @@
...
+# should fill with WRITE*VERB.  Now change the vision input to HI.  The phrase
+# should now fill with NOUN*HI.  If you get rid of the visual input (by setting
+# the value to nothing), it shuold sent HI to the hand.  If you try the same
 # thing with SAY it should send the result to speech.
 
 import nengo
 import nengo.spa as spa
 
-D = 32  # the dimensionality of the vectors
+D = 64  # the dimensionality of the vectors
```
tcstewar: If it's possible to get away with 32 dimensions, I'd prefer it, as then it'll work okay on slower computers....

nengo_gui/examples/tutorial/21-spa-sequence.py
```diff
@@ -11,7 +11,7 @@
 
 model = spa.SPA()
 with model:
-    model.memory = spa.Memory(D)
+    model.memory = spa.State(D, feedback=1)
```
tcstewar: I think this should have ```feedback_synapse=0.01``` as well

nengo_gui/examples/tutorial/22-spa-sequence-controlled.py.cfg
```diff
@@ -1,36 +1,36 @@
...
 _viz_config[_viz_2].max_value = 1
 _viz_config[_viz_2].min_value = -1
-_viz_config[_viz_2].height = 0.361262
-_viz_3 = nengo_gui.components.ValueTemplate(model.thalamus.actions.output)
-_viz_config[_viz_3].label_visible = 1
-_viz_config[_viz_3].width = 0.227926
-_viz_config[_viz_3].x = 0.621798
-_viz_config[_viz_3].y = 1.31364
+_viz_config[_viz_2].height = 0.384941
+_viz_3 = nengo_gui.components.Value(model.bg.output)
```
tcstewar: This is the wrong output graph.  It should be showing model.thalamus.actions.output

nengo_gui/examples/tutorial/22-spa-sequence-controlled.py
```diff
@@ -32,8 +38,6 @@
         'dot(memory, D) --> memory=E',
         'dot(memory, E) --> memory=vision',
         )
-
```
tcstewar: Hmm...  how do you feel about tossing in an extra action that is ```'0.5 --> memory=vision',```?  That seems to avoid a few cases for me where it gets stuck not doing anything....

nengo_gui/examples/tutorial/23-spa-binding.py.cfg
```diff
@@ -1,62 +1,60 @@
...
 _viz_config[model.memory].expanded=False
 _viz_config[model.memory].has_layout=False
-_viz_config[model.memory.state].expanded=False
-_viz_config[model.memory.state].has_layout=False
-_viz_config[model.shape].pos=(0.08471786321417869, 0.1569041118557128)
-_viz_config[model.shape].size=(0.0968667431223761, 0.07580458343675023)
+_viz_config[model.memory.state_ensembles].expanded=False
+_viz_config[model.memory.state_ensembles].has_layout=False
+_viz_config[model.shape].pos=(0.48323371886886723, 1.2739592850585906)
+_viz_config[model.shape].size=(0.4, 0.21052631578947367)
```
tcstewar: Did you mean to reorganize the layout like this?  The spa clouds are now far away from the components they are plotting, which seems strange to me.

nengo_gui/examples/tutorial/02-two-neurons.py
ikajic: On the machine I'm working on it happens with both Chrome and Firefox. I also tried on my laptop, the same happens with Firefox, Chrome works slightly better but it's still lagging. I will then remove these lines, but I think it would be useful to mention this elsewhere. Maybe a FAQ or just verbally when introducing tutorials.

nengo_gui/examples/tutorial/25-spa-parse.py.cfg
```diff
@@ -1,50 +1,48 @@
-_viz_0 = nengo_gui.components.PointerTemplate(model.vision)
-_viz_config[_viz_0].label_visible = 1
+_viz_0 = nengo_gui.components.Pointer(model.vision,target='default')
```
tcstewar: I like this new layout in general, but the ```phrase``` cloud seems placed on the opposite side from the ```phrase``` network, and it'd be good to have the BG input and the thalamus.actions.output plots.

nengo_gui/examples/tutorial/25-spa-parse.py
```diff
@@ -4,53 +4,67 @@
...
-        'dot(memory, VERB*WRITE) - dot(vision, WRITE+SAY+A+B+C+D+E)'
-            '--> hand=memory*~NOUN',
-        'dot(memory, VERB*SAY) - dot(vision, WRITE+SAY+A+B+C+D+E)'
-            '--> speech=memory*~NOUN',
+        'dot(vision, WRITE+SAY) --> verb=vision',
+        'dot(vision, YES+NO+HI+BYE+OK) --> noun=vision',
+        'dot(phrase, VERB*WRITE) - dot(vision, WRITE+SAY+YES+NO+HI+BYE+OK)'
+            '--> hand=phrase*~NOUN',
+        'dot(phrase, VERB*SAY) - dot(vision, WRITE+SAY+YES+NO+HI+BYE+OK)'
+            '--> speech=phrase*~NOUN',
```
tcstewar: I still find this model to sometimes fail for me.  Looking at the thalamus and bg plots, it looks like it's sometimes selecting the last two actions, even when there is a visual input.  A quick way to fix this is to penalize it even more, but multiplying by two:

```dot(phrase, VERB*WRITE) - 2*dot(vision, WRITE+SAY+YES+NO+HI+BYE+OK)```

That seems to make it very reliable for me.

nengo_gui/examples/tutorial/02-two-neurons.py
tcstewar: Fair enough.  Let's add in the comment that you've got here, and include the comment and the code in examples 00, 01, and 02.

ikajic:
I've included the changes mentioned in the discussion above. The last example `25-spa-parse.py`seems also to work with 32 dimensions, but to be on the safe side I would appreciate if anyone else could try testing the model with a few inputs, thanks!

tcstewar:
Seems to work for me...  nicely done!

PR #590, opened 2015-10-21
Made import relative so that it works in Python3 again

tcstewar:
Python3 support was broken since Aug 26 due to
38192511b22bbcb35ebabb3904cee039e4a54021


tcstewar:
Actually, since this is a small change and a show-stopper for anyone on Python 3, I'll just merge this right away.

PR #592, opened 2015-10-28
Fixed thalamus routing switch errors

Seanny123:
Two errors were happening with the thalamus routing:
- A JavaScript error was being thrown about labels
- #563 

This pull request resolves both those issues

### DIFF
--- a/nengo_gui/components/netgraph.py
+++ b/nengo_gui/components/netgraph.py
@@ -125,12 +125,15 @@ class NetGraph(Component):
             if new_uid != uid:
                 new_item = None
 
+            same_class = (isinstance(new_item, old_item.__class__) or
+                          isinstance(old_item, new_item.__class__))
+
             # find reasons to delete the object.  Any deleted object will
             # be recreated, so try to keep this to a minimum
             keep_object = True
             if new_item is None:
                 keep_object = False
-            elif not isinstance(new_item, old_item.__class__):
+            elif same_class:
                 # don't allow changing classes
                 keep_object = False
             elif self.get_extra_info(new_item) != self.get_extra_info(old_item):
--- a/nengo_gui/components/netgraph.py
+++ b/nengo_gui/components/netgraph.py
@@ -125,8 +125,11 @@ class NetGraph(Component):
             if new_uid != uid:
                 new_item = None
 
-            same_class = (isinstance(new_item, old_item.__class__) or
-                          isinstance(old_item, new_item.__class__))
+            same_class = False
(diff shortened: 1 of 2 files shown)


nengo_gui/components/netgraph.py
```diff
@@ -125,12 +125,15 @@ def _reload(self, code=None):
...
+            same_class = (isinstance(new_item, old_item.__class__) or
+                          isinstance(old_item, new_item.__class__))
+
             # find reasons to delete the object.  Any deleted object will
             # be recreated, so try to keep this to a minimum
             keep_object = True
             if new_item is None:
                 keep_object = False
-            elif not isinstance(new_item, old_item.__class__):
+            elif same_class:
```
tcstewar: Yikes...   really good catch finding this.  That logic was definitely wrong.....  

Do you think something like this might be more robust?

```
for cls in [nengo.Ensemble, nengo.Node, nengo.Network, nengo.Connection]:
    if isinstance(new_item, cls) and isinstance(old_item, cls):
        same_class = True
        break
else:
    same_class = False
```

nengo_gui/static/components/netgraph.js
```diff
@@ -307,9 +307,6 @@ Nengo.NetGraph.prototype.on_message = function(event) {
         var item = this.svg_objects[data.uid];
         item.set_label(data.name);
 
-        var item_mini = this.minimap_objects[data.uid];
-        item_mini.set_label(data.name);
-
```
tcstewar: Was that label ever used for anything?  I'm not sure I understand why this was causing an issue...  was it a side effect of the above error?

Seanny123: This was something I corrected while debugging. It was probably a side-effect of the above error. This label was never actually used for anything.

nengo_gui/components/netgraph.py
Seanny123: I like that logic way better. I'll make the change.

Seanny123:
Made the changes. Ready for review again.

nengo_gui/static/components/netgraph.js
tcstewar: Hmm..  I'm still not sure I'm following the logic here right.  @studywolf , could you quickly confirm that it's okay to remove this setting of the label?

studywolf: ahhh yeah i think that might just be a holdover from me copying all the svg_objects code for the minimap when i was getting it started, i'm pretty sure i never used the label for anything

tcstewar:
Looks great -- thanks for finding this!

Issue #599, opened 2015-11-11
Graphs and menu elements get selected


celiasmith:
As in this screen shot, menus and graph elements often come up selected, somewhat randomly.  Apparently a Mac Chrome issue. (The grey is blue in the GUI, it's a screen cap colour change)

![untitled](https://cloud.githubusercontent.com/assets/1984851/11095747/845efdfa-8864-11e5-819f-945e264e31eb.png)

tbekolay:
Fixed in #600

PR #600, opened 2015-11-11
Fixes accidental selection in browser

celiasmith:
Addresses #599 

This works in Chrome and Safari on Mac; Please check other OSs.  Be sure to check that sliders are still editable, and that ctrl/cmd-a works in the editor window.

On Mac, cmd-a used to select many elements of the GUI that we didn't want selected.


Seanny123:
Works for me on Linux and Chrome

tcstewar:
Works for me too

PR #605, opened 2015-11-13
GUI style and docs fixups

Seanny123:
I'm creating another branch for all the little problems I find while editing the code so that I don't clog up the other pull requests.

### DIFF
--- a/.jscs.json
+++ b/.jscs.json
@@ -1,0 +1,88 @@
+{
+    "requireCurlyBraces": [
+        "if",
+        "else",
+        "for",
+        "while",
+        "do",
+        "try",
+        "catch"
+    ],
+    "requireOperatorBeforeLineBreak": true,
+    "maximumLineLength": {
+      "value": 80,
+      "allExcept": ["comments", "regex"]
+    },
+    "validateIndentation": 4,
+
+    "disallowMultipleLineStrings": true,
+    "disallowMixedSpacesAndTabs": true,
+    "disallowTrailingWhitespace": true,
+    "disallowSpaceAfterPrefixUnaryOperators": true,
+    "disallowMultipleVarDecl": true,
+    "disallowKeywordsOnNewLine": ["else"],
+
+    "requireSpaceAfterKeywords": [
+      "if",
+      "else",
+      "for",
+      "while",
+      "do",
+      "switch",
+      "return",
+      "try",
+      "catch"
+    ],
+    "requireSpaceBeforeBinaryOperators": [
+        "=", "+=", "-=", "*=", "/=", "%=", "<<=", ">>=", ">>>=",
+        "&=", "|=", "^=", "+=",
+
+        "+", "-", "*", "/", "%", "<<", ">>", ">>>", "&",
+        "|", "^", "&&", "||", "===", "==", ">=",
+        "<=", "<", ">", "!=", "!=="
+    ],
+    "requireSpaceAfterBinaryOperators": true,
+    "requireSpacesInConditionalExpression": true,
+    "requireSpaceBeforeBlockStatements": true,
+    "requireSpacesInForStatement": true,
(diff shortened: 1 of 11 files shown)


nengo_gui/namefinder.py
```diff
@@ -22,9 +22,12 @@ def find_names(self, net):
         classes = (nengo.Node, nengo.Ensemble, nengo.Network,
                    nengo.Connection)
 
+        private = False
+        in_lists = False
```
tcstewar: Why are these being initialized outside the loop?

Seanny123: Because I forgot that it doesn't matter in Python if you initialise variables inside or outside a for-loop. http://stackoverflow.com/questions/407255/difference-between-declaring-variables-before-or-in-loop

Seanny123: Because it's bad to initialise variables inside a loop? I can't find a reference for this however...

Seanny123: Bizarrely, based off my own tests, defining things inside the loop is faster...

```
import timeit

def outside(num_list):
    total = 0
    even = False
    for num in num_list:
        even = num % 2 == 0
        if even:
            total += 1
    return total


def inside(num_list):
    total = 0
    for num in num_list:
        even = num % 2 == 0
        if even:
            total += 1
    return total

if __name__ == '__main__':
    res1 = timeit.timeit("outside([n for n in range(50)])", setup="from __main__ import outside", number=15)
    print(res1)
    res2 = timeit.timeit("inside([n for n in range(50)])", setup="from __main__ import inside", number=15)
    print(res2)
```

tcstewar: > Because it's bad to initialise variables inside a loop? I can't find a reference for this however...

Hmm, i don't think I share that intuition.  Mostly, for me, it's about clarity -- when I read this code I got confused looking for the reason why it was initialized like that, especially for a variable that's only used inside the loop.  Since it was defined outside it looked like you needed a default value for if the loop executed zero times....

Seanny123: Fixed.

jgosmann: On my machine it is virtually no difference where the variable is defined.

jgosmann: Also, clarity > efficiency as long as it is not a part of the code which runs really a lot.

.jscs.json
```diff
@@ -0,0 +1,88 @@
+{
```
tcstewar: Is this file supposed to be in here?  I assume it's a config for the javascript style checker, but there's no documentation or indication as to what it's for....  did you mean to include it?

tcstewar: Nevermind, I found the mention in workflow.rst

nengo_gui/namefinder.py
```diff
@@ -23,8 +23,9 @@ def find_names(self, net):
                    nengo.Connection)
 
         for inst_attr in dir(net):
-            if (not inst_attr.startswith('_')
-                and inst_attr not in base_lists + all_lists):
+            private = inst_attr.startswith('_')
+            in_lists = inst_attr in base_lists + all_lists
+            if (not private and not in_lists):
```
tcstewar: Remove the parentheses!

.jscs.json
Seanny123: It's addressed in line 27 of workflow.rst https://github.com/nengo/nengo_gui/pull/605/files#diff-ff11f4d5436e9c5c3e9fdb37dd16c379R27

tcstewar:
Looks good other than the parens thing...  If you rebase it to fix the conflicts, I can do some final testing and merging.

nengo_gui/namefinder.py
Seanny123: I thought because of the `and` the parenthesis were allowed. My bad. When should I use them?

PR #611, opened 2015-11-17
Added ability to add list and dict to config

Seanny123:
Used for saving legend_labels that are set by the user

### DIFF
--- a/nengo_gui/config.py
+++ b/nengo_gui/config.py
@@ -47,10 +47,11 @@ class Config(nengo.Config):
                 lines.append('%s = %s' % (uid, obj.code_python(uids)))
                 for k in obj.config_defaults.keys():
                     v = getattr(self[obj], k)
-                    if isinstance(v, bool):
-                        val = '%s' % v
-                    else:
+                    try:
                         val = '%g' % v
+                    except TypeError:
+                        val = '%s' % v
+
                     lines.append('_viz_config[%s].%s = %s' % (uid, k, val))
 
         return '\n'.join(lines)
--- a/nengo_gui/config.py
+++ b/nengo_gui/config.py
@@ -57,11 +57,9 @@ class Config(nengo.Config):
                                          "values that can be successfully "
                                          "evaluated are allowed." % (val))
 
-                    try:
-                        assert recovered_v == v
-                    except AssertionError:
+                    if recovered_v != v:
                         raise ValueError("Cannot save %s to config, recovery "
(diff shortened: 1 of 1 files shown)


nengo_gui/config.py
```diff
@@ -47,10 +47,11 @@ def dumps(self, uids):
                 lines.append('%s = %s' % (uid, obj.code_python(uids)))
                 for k in obj.config_defaults.keys():
                     v = getattr(self[obj], k)
-                    if isinstance(v, bool):
-                        val = '%s' % v
-                    else:
+                    try:
                         val = '%g' % v
```
tcstewar: If ```v``` is True, this will result in the string ```'1'```, rather than ```True```.  That's why the isinstance check was there.

Seanny123: Fixed.

tcstewar:
So I'm a bit worried about this, in that the functionality you've added here breaks horribly for lots of different inputs.  For example, if I set a config value to a numpy array, it'll save fine, and then fail when trying to load the value.  In the old version, it could only save things that we know can be loaded.

What is your goal here with this change?  To allow lists and dicts?  Why not just allow those, rather than allowing *anything*?  And what things do you want to allow inside those lists?  what about lists of numpy arrays?  Should that be allowed?

I can see an argument for allowing *anything*, and leaving it up to the programmer to only do things that can be reloaded.  But if that's the plan, why not just replace this whole code with ```val = repr(v)```?

tcstewar:
Oh, and I forgot to mention setting a config option to a string....  that could cause all sorts of strange bugs with this change.

Seanny123:
I did consider only allowing lists and dicts of strings, but that seemed like an annoying amount of code to write.

I would vote for allowing anything and just using `val = repr(v)`. The alternative has too many corner cases to cover and is going to cause code bloat.

Should this be discussed in the meeting or should I go ahead and make the change?

Seanny123:
Fixed now

Seanny123:
Even better now. (:

Seanny123:
If I put the `assert` in a separate exception and explained when it failed that the value couldn't be retrieved properly, would that be better?

tcstewar:
> If I put the assert in a separate exception and explained when it failed that the value couldn't be retrieved properly, would that be better?

I think so.  :)

Seanny123:
Amended

tcstewar:
Looks good.  :)

PR #616, opened 2015-11-21
Fix label svg sizing

Seanny123:
Current method doesn't work for long labels, because I forgot to include the text as part of the label class...

### DIFF
--- a/nengo_gui/static/nengo.js
+++ b/nengo_gui/static/nengo.js
@@ -101,6 +101,7 @@ Nengo.draw_legend = function(parent, labels, color_func, uid) {
               .append("text")
               .attr("x", 15)
               .attr("y", function(d, i){ return i *  20 + 9;})
+              .attr("class", "legend-label")
               .html(function(d, i) {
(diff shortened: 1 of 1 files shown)


tcstewar:
What can I do to see the incorrect behaviour that this is trying to fix?  I've mucked with making long labels, and it seems to work for the things I've tried (mostly opening a SPA plot and a SPA cloud and seeing what happens when I create new semantic pointer inputs with long or short text names).

Seanny123:
You'd only see it on the other branches and on reload of a network with a saved config including long labels.

tcstewar:
So this only affects things that can't happen on the master branch?  Hmm, that does make it hard/impossible to test, but it is a pretty small change, so I guess it's okay.  Technically it should be done in those other branches, but whatever.  :)

tcstewar:
Again, if  you fix the conflicts, I can merge...  :)

tcstewar:
Btw, what method are you using to update the branch to master to fix these conflicts?

PR #624, opened 2015-12-09
Fix Python 3 `zip` error

mundya:
Fixes an error that occurs when, e.g., running `25-spa-parse.py` with Python 3.

Coincidentally, I can't select or copy text from the error pane below the code in Nengo GUI (Firefox 42.0, Arch Linux); is that a bug?

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -19,6 +19,7 @@ to the file like so:
 Please keep this list sorted alphabetically by first name.
 
 - Aaron Voelker <arvoelke@gmail.com>
+- Andrew Mundy <andrew.mundy@manchester.ac.uk>
 - Brent Komer <brent.komer@gmail.com>
 - Chris Eliasmith <celiasmith@uwaterloo.ca>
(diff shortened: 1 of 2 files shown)


CONTRIBUTORS.md
```diff
@@ -19,6 +19,7 @@ to the file like so:
 Please keep this list sorted alphabetically by first name.
 
 - Aaron Voelker <arvoelke@gmail.com>
+- Andrew Mundy <andrew.mundy@manchester.ac.uk>
```
mundya: I'm totally happy to not add my name to this list for such a small change.

Seanny123:
LGTM.

tcstewar:
> Fixes an error that occurs when, e.g., running 25-spa-parse.py with Python 3.

Nice way to fix that...  thanks!

> Coincidentally, I can't select or copy text from the error pane below the code in Nengo GUI (Firefox 42.0, Arch Linux); is that a bug?

Huh, I never tried selecting from there before, but yeah, it looks like that's been disabled in all browsers.  Should be an easy thing to turn back on if an issue gets made for it....

PR #646, opened 2016-02-05
Better xy

### LINKED ISSUE #551: XYPlot fails with 1 dimension

Seanny123:
Fixing a bunch of problems with XY plots:
- Direct mode is awkward for XY plots
- Axes don't cross at zero if you don't have uneven ranges on the x and y axis
- When you reduce dimensions from 2 to 1 and you have an xy-plot affected, bad things happen

Resolves #551 , #363 and #271

### DIFF
--- a/nengo_gui/static/components/2d_axes.js
+++ b/nengo_gui/static/components/2d_axes.js
@@ -5,10 +5,8 @@
  * @param {DOMElement} parent - the element to add this component to
  * @param {float} args.width - the width of the axes (in pixels)
  * @param {float} args.height - the height of the axes (in pixels)
- * @param {bool} args.center_x_axis - should the x-axis be in the middle
- * @param {bool} args.center_y_axis - should the y-axis be in the middle
- * @param {float} args.min_value - minimum value on y-axis
- * @param {float} args.max_value - maximum value on y-axis
+ * @param {array} args.x_range - minimum and maximum value on x-axis
+ * @param {array} args.y_range - minimum and maximum value on y-axis
  */
 
 Nengo.Axes2D = function(parent, args) {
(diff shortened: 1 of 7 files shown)


nengo_gui/static/components/2d_axes.js
```diff
@@ -26,17 +24,15 @@ Nengo.Axes2D = function(parent, args) {
...
     /** spacing between the graph and the outside edges (in pixels) */
     this.set_axes_geometry(args.width, args.height);
-    
+
     /** define the x-axis */
     this.axis_x = d3.svg.axis()
         .scale(this.scale_x)
         .orient("bottom")
-        .ticks(2);
+        this.axis_x.ticks(2);
```
tcstewar: why was this changed?

nengo_gui/static/components/2d_axes.js
```diff
@@ -68,7 +64,7 @@ Nengo.Axes2D.prototype.set_axes_geometry = function(width, height) {
 /**
  * Adjust the graph layout due to changed size
  */
-Nengo.Axes2D.prototype.on_resize = function(width, height) {
+Nengo.Axes2D.prototype.on_resize = function(width, height, x_offset, y_offset) {
```
tcstewar: Why pass in x_offset and y_offset and not use them?

nengo_gui/static/components/xyvalue.js
```diff
@@ -73,20 +79,62 @@ Nengo.XYValue.prototype.on_message = function(event) {
...
-        .x(function(d, i) {
-            return self.axes2d.scale_x(
-                shown_data[self.index_x][i]);
-            })
-        .y(function(d) {return self.axes2d.scale_y(d);})
-    this.path.data([shown_data[this.index_y]])
-             .attr('d', line);
+    /** update the lines if there is data with valid dimensions */
+    var good_idx = self.index_x < self.n_lines && self.index_y < self.n_lines
+    if (this.data_store.data.length > 1 && good_idx) {
```
tcstewar: This should just be ```if(good_idx)``` (so you can plot 0,0 on a 1-D, if you want to)

tcstewar:
The white box in the background when the error message is there gets a bit odd when the window is resized.  Perhaps get rid of the background white box and just mark the text as having background and a border in HTML?

tcstewar:
Circle should be 50% opacity, and @jgosmann thinks it should be blue.  (I concur, but really shouldn't have a vote).

nengo_gui/static/components/xyvalue.js
```diff
@@ -73,20 +79,62 @@ Nengo.XYValue.prototype.on_message = function(event) {
...
+          .attr("fill", "white")
+          .attr("stroke-width", 1)
+          .attr("stroke", "rgb(0,0,0)")
+          .attr("fill-opacity", 1);
+
+        // create the HTML text element
+        this.warning_text = document.createElement('div');
+        this.div.appendChild(this.warning_text);
+        this.warning_text.innerHTML = "Change<br>Dimension<br>Indices";
+        this.warning_text.style = "color:#a94442; position: absolute; text-align: center; margin-left: auto; margin-right: auto; width:100%; top:37%; display:box; box-align:center; box-pack:center;"
```
tcstewar: This should be in the style sheet

tcstewar:
Looks good to me!

PR #647, opened 2016-02-05
Enable console text selection

### LINKED ISSUE #634: Text in console is not selectable.
This makes it hard to copy and paste errors.

s72sue:
Fixes #634

### DIFF
--- a/nengo_gui/components/raster.py
+++ b/nengo_gui/components/raster.py
@@ -21,6 +21,10 @@ class Raster(Component):
         if n_neurons is None:
             n_neurons = min(self.max_neurons, 10)
         self.n_neurons = n_neurons
+        self.conn = None
+        self.node = None
+
+
 
     def attach(self, page, config, uid):
         super(Raster, self).attach(page, config, uid)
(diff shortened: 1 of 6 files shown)


jgosmann:
You should change all of the `*-select` lines. Also, this exacty reverts 85c8e31c7ade5deb237e1d644cb4a544919fa439. We probably should decide whether we want to
1. allow text selection in general
2. disallow text selection in general
3. only allow text selection in certain areas (like the console)

s72sue:
Hmm...I didn't know about this other issue. I think it would make sense to
allow text selection in certain areas like the console. We definitely don't
want text getting randomly selected in the visualizer since that can be
confusing (and looks weird).

s72sue:
I have made the change to enable text selection in console while disabling accidental selection in the visualizer. 

jgosmann This seemed to be the most sensible combination to me, but let me know if people want some other combination for allowing/disallowing text selection in certain areas.

tcstewar:
Looks good to me.  :)

That said, watch out for including extra files into the PR that aren't part of it.  They greatly complicate the review process and mean that when we look back in the history to find out why certain lines were added, we get a lot of extraneous information to sift through.

s72sue:
Oh I apologize ..didn't realise that those files got added to the PR. Will
definitely watch out in the future.

Issue #375, opened 2015-06-09
Persist configuration options to disk


tbekolay:
#372 reminded me of an issue we should track (and probably needs some discussion). We have some configuration options now, accessible through the gear in the top toolbar. Right now, it's all client-side, so it works for the current session and that's it. It should work forever! To do this, we need to persist something to disk, which means that we need to let Python know about the configuration stuff and save it somewhere.

A few discussion points.

1. Where should the configuration be saved?
2. What format should we use?

I think the `.ini` style configuration that we use in Nengo is probably a decent enough format to use. We could also use where Nengo saves configuration:

```ipython
In [2]: nengo.utils.paths.config_dir
Out[2]: '/Users/Trevor/.config/nengo'
```

I'm okay with using it directly, but it could be the `nengo_gui` equivalent too.

jgosmann:
I would vote for separate configuration files. Seems cleaner to me.

tbekolay:
Oh yes! I agree, separate configuration files (not using `nengorc` for sure), just using the same format and storing the file (`nengo_gui.conf` or something?) in the same directory (`~/.config/nengo`).

mundya:
FWIW `nengo_spinnaker` config sneaks itself into `~/.config/nengo` but will follow whatever convention is set here...

jgosmann:
> just using the same format and storing the file (nengo_gui.conf or something?) in the same directory (~/.config/nengo)

I'm ok with that.

rfernand2:
Persisting the font size will make it easier for Nengo instructors to keep the font large enough for their audiences.  :)

Seanny123:
+1 for same directory as `nengo`

tcstewar:
This was solved in #649 by persisting to the javascript localstorage!  Much nicer than figuring out where to store it ourselves...

PR #649, opened 2016-02-05
Persist configuration options to local storage

tbekolay:
The bulk of the logic is in `static/config.js`. So far we only persist the 4 configuration options accessible through the gear icon in the top toolbar, but we can add more things easily.

Addresses issue #375.

### DIFF
--- a/nengo_gui/static/ace.js
+++ b/nengo_gui/static/ace.js
@@ -4,7 +4,7 @@
  *
  * @param {string} uid - A unique identifier
  * @param {dict} args - A set of constructor arguments (see Nengo.Component)
- * Ace function is written into HTML by server and called when the 
+ * Ace function is written into HTML by server and called when the
  * page is loaded.
  */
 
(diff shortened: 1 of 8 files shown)


tbekolay:
Updated this to also persist some options for the code editor.

The only thing that I think could persist that doesn't is the backend used... which I'm not sure it should be persisted. Thoughts?

tcstewar:
> The only thing that I think could persist that doesn't is the backend used... which I'm not sure it should be persisted. Thoughts?

I'm okay with that not being persisted, especially since you can already set it on the command line...

tbekolay:
Cool cool. I think maybe we should think about moving that somewhere other than through the Configuration Options modal window eventually, but definitely not a priority for now!

nengo_gui/static/ace.js
```diff
@@ -174,68 +195,56 @@ Nengo.Ace.prototype.on_message = function (event) {
     }
 }
 
+Nengo.Ace.prototype.on_resize = function() {
+    this.max_width = $(window).width() - 100;
+    if (this.width > this.max_width) {
+        this.width = this.max_width;
+    }
+    console.log(this.max_width);
```
tcstewar: I assume this console.log should get removed

tcstewar:
Very very nifty.  And other than the one console.log comment, I didn't find any other issues.  :)

nengo_gui/static/ace.js
tbekolay: You are correct! I'll amend it.

tbekolay:
console.log statement removed!

Issue #644, opened 2016-02-02
option to disable auto-update


tcstewar:

Seanny123:
Resolved via #676

Issue #639, opened 2016-01-22
shift-enter to run from editor


studywolf:
It would be nice to be able to hit shift-enter from the editor portion of the window to run the model. Who has time to reach for the mouse??

tcstewar:
+1

PR #651, opened 2016-02-05
toggle for auto-updating from code editor

bjkomer:
Adding the option to disable automatic synchronization of the model and the code in the editor.

Which shortcut keys to use is still up in the air, currently using Ctrl-Shift-1 to toggle auto-updating, and Ctrl-1 to trigger an update. The toggle can also be set in the config, and the trigger with a tool bar button (which also doubles as an indicator of whether or not the code differs from the model).
Addresses #644 and also provides a way around #444 #550 

Allowed Shift-Enter to run the model from the editor window (also works in the other windows). Addresses #639


tcstewar:
Nifty!  And I like how clean the implementation is....  :)

I'm still not sure what the right hot-key should be for updating the netgraph.  Any suggestions?

I'm also not sure about the icon, but I'm having trouble finding a better one...

tcstewar:
Argh; this needs to get updated for the persistent storage stuff that got merged already.

tcstewar:
> I'm also not sure about the icon, but I'm having trouble finding a better one...

Maybe glyphicon-circle-arrow-left ?

tcstewar:
Ah, we have a wiki page for more icons!

https://github.com/nengo/nengo_gui/wiki/Icon-Libraries

tcstewar:
Grr, looks like I screwed up the rebase on this one.  :(  I'm going to make a new branch with just the actual content of this PR.  Sorry about complicating things!

tcstewar:
I've made a new PR for this based on current master: #676

tcstewar:
This got merged as #676 (same content, but with a cleaner history than this one, due to a mess I made)

Issue #388, opened 2015-06-10
Semantic pointer plot should take into account label


tbekolay:
Right now, the text in the pointer plot can overlap with the label at the top of the graph. We should offset the top to take into account the label (or figure out some other way of ensuring that they don't overlap)

Example:

![screen shot 2015-06-10 at 12 26 31 pm](https://cloud.githubusercontent.com/assets/203709/8087958/f7d9f4b4-0f6b-11e5-851b-a2449b7970c8.png)

PR #653, opened 2016-02-05
Pointer cloud fixed to not overlap with label

pblouw:
Addresses issue #388 by shifting and rescaling the semantic pointer cloud.


tcstewar:
Looks good to me.  I've tred it on all the tutorial/spa examples and seems to look good to me!

(Note, this isn't a perfect solution, but it's no more hacky than the current solution, so it feels like an improvement to me).

PR #655, opened 2016-02-05
Added self.node and self.conn to __init__

### LINKED ISSUE #553: Nengo Objects in Component Declaration
`self.conn` and `self.node` should be declared in `__init__()`

s72sue:
fixes #553


tcstewar:
Thanks for this!  :)  Looks like this covers all the cases.

s72sue:
Yup, it covers all component objects!

Issue #10, opened 2015-01-20
Save as pdf


celiasmith:
Make nice ways to save the layout to pdf/svg

studywolf:
it would also be nice to be able to save the response curves in the neuron details as a pdf easily

jgosmann:
Duplicate of #52?

Seanny123:
@jgosmann I think they refer to different scopes. One refers to saving the whole layout while the other refers to saving individual plots.

However, just to be the devil's advocate, the reason why we don't want the person to rely on screenshot software is that the screenshot might be too low resolution right?

Seanny123:
Also, in terms of flow, I'm assuming that people want a context menu for individual plots. For saving the layout, should that be a context menu in the background or a top-menu item?

studywolf:
yeah the screenshot is passable but it would be really nice to get a pdf /
svg so it looks clean after manipulating it for a presentation

jgosmann:
It should be possible to get the SVG parts as an SVG without additional libraries (which you could convert with an external tool to PDF or whatever you need).

tcstewar:
PR #656 did the SVG export, and you can convert that to pdf by opening it in a decent browser and exporting to PDF.

PR #656, opened 2016-02-06
Export to SVG/PDF

ikajic:
Opening this PR to notify folks I'm working on #10 . Still work in progress!

Exporting to SVG works but the XML ignores CSS so the exported images look crappy, working on embedding CSS properly.

### DIFF
--- a/nengo_gui/static/components/netgraph.css
+++ b/nengo_gui/static/components/netgraph.css
@@ -1,4 +1,5 @@
 .netgraph {
+    font-family: 'Helvetica Neue', Helvetica, Arial, sans-serif;
 }
 
 .ensemble{
--- a/nengo_gui/static/components/netgraph.js
+++ b/nengo_gui/static/components/netgraph.js
@@ -138,6 +138,24 @@ Nengo.NetGraph = function(parent, args) {
     this.g_items = this.createSVGElement('g');
     this.svg.appendChild(this.g_items);
 
+    /** Reading netgraph.css file as text and embedding it within def tags,
+     *  this is needed for saving the SVG plot to disk*/
+
+    /** load contents of the CSS file as string */
+    var file = document.getElementById('netgraphcss');
+    var css = Array.prototype.map.call(file.sheet.cssRules, function
+            css_text(x) {return x.cssText; } ).join('\n');
+
+    /** embed CSS code into SVG tag */
+    var s = document.createElement('style');
+    s.setAttribute('type', 'text/css');
+    s.innerHTML = "<![CDATA[\n" + css + "\n]]>";
+
+    var defs = document.createElement('defs');
+    defs.appendChild(s);
+
+    this.svg.insertBefore(defs, this.svg.firstChild);
+
     /** connect to server */
     this.ws = Nengo.create_websocket(args.uid);
     this.ws.onmessage = function(event) {self.on_message(event);}
--- a/nengo_gui/static/modal.js
+++ b/nengo_gui/static/modal.js
@@ -55,6 +55,37 @@ Nengo.Modal.prototype.footer = function(type, ok_function, cancel_function){
         $('#confirm_reset_button').on('click', function() {
             toolbar.reset_model_layout();
         });
+    } else if (type === 'confirm_savepdf') {
+        this.$footer.append('<button type="button" ' +
+            'id="confirm_savepdf_button" class="btn btn-primary" data-dismiss="modal">Save</button>');
+        this.$footer.append('<button type="button" ' +
+            'class="btn btn-default" data-dismiss="modal">Close</button>');
+        $('#confirm_savepdf_button').on('click', function() {
+            var svg = $("#main svg")[0];
+
+            // Serialize SVG as XML
+            var svg_xml = (new XMLSerializer).serializeToString(svg);
+            source = '<?xml version="1.0" standalone="no"?>' + svg_xml;
+            source = source.replace("&lt;", "<");
+            source = source.replace("&gt;", ">");
+
+            var svg_uri = 'data:image/svg+xml;base64,' + btoa(source);
+
+            // Extract filename from the path
+            var path = $("#filename")[0].textContent;
+            filename = path.split('/').pop()
+            filename = filename.split('.')[0]
+
+            // Initiate download
+            var link = document.createElement("a");
+            link.download = filename + ".svg";
+            link.href = svg_uri;
+
+            // Adding element to the DOM (needed for Firefox)
+            document.body.appendChild(link);
+            link.click();
+            document.body.removeChild(link);
+        });
     } else if (type === 'refresh') {
         this.$footer.append('<button type="button" ' +
             'id="refresh_button" class="btn btn-primary">Refresh</button>');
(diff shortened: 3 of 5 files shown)


ikajic:
Embedded CSS into the SVG file but only browsers can read the CSS (opening the .svg image in any other viewer will give rubbish). What works now is opening the .svg in browser and printing as .pdf

ikajic:
Ready to be tested! 

Some of the issues I have encountered:
- Print svg as pdf in browser does not work for `Chromium 45.0.2454.85 Built on 8.1, running on Debian 8.2`
- Save button doesn't do anything with `Mozilla Iceweasel 38.2.1` (working on this), but hey at least it prints it nicely as a pdf :)

tbekolay:
This is awesome!! It will save a ton of time making figures in the future! Even if these aren't perfect, they should definitely be in 0.2 as tweaking an existing SVG is way easier than making one from scratch. This worked well on my machine, which is running Mac OS X and Chrome version 49.0.2623.28.

I noticed a few minor issues trying this out:

1. I think the export layout button should be to the left of the settings icon -- I think it's just natural for the settings to be on the far right.
2. I would expect pressing the "save" button the modal dialog window to close the modal dialog. I opened the SVG right after saving it, and was surprised to see the modal dialog still up when I came back to that tab.
3. The font in the SVG was incorrect for me; `nengo_gui` uses `font-family: 'Helvetica Neue', Helvetica, Arial, sans-serif`, but I get no font-family in the saved SVG, which renders for me as Times New Roman. Definitely not a dealbreaker, but something to maybe make into an issue to deal with later?
4. I noticed in the saved SVG that there are two `<defs xmlns...><style type="text/css"><![CDATA[...` tags. They look to have identical content. Looking at it in `nengo_gui` itself, this appears to actually be a problem there too, so it isn't a bug in the SVG serialization itself.

A more long-term thing related to 1. is that our toolbar is getting pretty long at the top. Most modern interfaces have opted for a button that expands into a dropdown with the things that you want quick-but-not-immediate access to (e.g., the hamburger at the top right of Chrome). A long toolbar can be a bit overwhelming for a user. So, I made issue #666 to keep that work separate from this PR.

ikajic:
> The font in the SVG was incorrect for me; nengo_gui uses font-family: 'Helvetica Neue', Helvetica, Arial, sans-serif, but I get no font-family in the saved SVG, which renders for me as Times New Roman. Definitely not a dealbreaker, but something to maybe make into an issue to deal with later?

Are you fine with me changing stuff in `netgraph.css`? Following seems to work:
```css
.netgraph {
    font-family: 'Helvetica Neue', Helvetica, Arial, sans-serif;
}
```

tbekolay:
Seems fine to me! I think the style is currently applied by Bootstrap, so including it in our CSS should only be a good thing.

ikajic:
> I noticed in the saved SVG that there are two <defs xmlns...><style type="text/css"><![CDATA[... tags. They look to have identical content. Looking at it in nengo_gui itself, this appears to actually be a problem there too, so it isn't a bug in the SVG serialization itself.

I don't have a rational explanation of what happened here. I managed to reproduce this once and every other time I saved SVG it had just one tag... do you mind trying again to see if you can reproduce it?

tbekolay:
The new changes fix 1-3 for me! And I figured out how to reproduce 4: it happens if I export the SVG multiple times; it adds a `defs` tag for each time I export. So, a few possible ways to fix it: 1) add the defs tag when the root SVG is created (rather than when it's exported), 2) delete the defs tag once the SVG is exported, or 3) detect if a defs tag exists and only add it if it isn't there.

ikajic:
> And I figured out how to reproduce 4: it happens if I export the SVG multiple times; it adds a defs tag for each time I export.

Aah.. thanks for figuring this out!

>  1) add the defs tag when the root SVG is created (rather than when it's exported)

This seems to be the cleanest solution! I assume [this](https://github.com/nengo/nengo_gui/blob/master/nengo_gui/static/modal.js#L404) is where the SVG object gets created?

tbekolay:
I think it's [this one](https://github.com/nengo/nengo_gui/blob/master/nengo_gui/static/components/netgraph.js#L115)

jgosmann:
Cool stuff. It seems that ImageMagick is able to convert the SVG including the CSS to other formats. But it produces shitty results for me (e.g., circles get polygons). Maybe that can be fixed with appropriate arguments to `convert`, but maybe not.

> 1) add the defs tag when the root SVG is created (rather than when it's exported)

This would technically still leave the CSS in there twice (from the `<link>` tag and in the SVG). It might be better to get rid of the linked CSS and have it only defined in the SVG. We probably can pass the CSS content from Python to the JS-Netgraph constructor.

ikajic:
> Cool stuff. It seems that ImageMagick is able to convert the SVG including the CSS to other formats. But it produces shitty results for me (e.g., circles get polygons). Maybe that can be fixed with appropriate arguments to convert, but maybe not.

I have tried several packages (including convert) for SVG to PDF conversion but it is actually not an easy problem. Usually they would export just pieces of an image which they are programmed to recognize. Until further, the solution to getting a PDF really is to do print from a browser.

> 1) add the defs tag when the root SVG is created (rather than when it's exported)
> This would technically still leave the CSS in there twice (from the <link> tag and in the SVG). It might be better to get rid of the linked CSS and have it only defined in the SVG. We probably can pass the CSS content from Python to the JS-Netgraph constructor.

I don't see a `<link>` tag in the exported SVG, did you mean having a `<link>` in `page.html`? Also what do you mean by "pass the CSS content from Python to the JS-Netgraph constructor"?

jgosmann:
> did you mean having a <link...> in page.html?

yes

> what do you mean by "pass the CSS content from Python to the JS-Netgraph constructor"?

The JS code line invoking the constructor of the Netgraph class gets put together by the Python backend and is inserted into `page.html` at the appropriate place before sending it to the browser. So it would be possible to read the CSS file with Python and insert it into the constructor call generated with Python.

jgosmann:
[Could this help?](https://github.com/shakiba/svgexport)

ikajic:
>
> Could this help? <https://github.com/shakiba/svgexport>
>

Not really. The results with SVG->PNG->PDF were crappy. I would leave the
scope of this PR to focus on exporting the layout to SVG and in the future
address the conversion to PDF. I will rename the  branch to reflect that.

So it would be possible to read the CSS file with Python and insert it into
> the constructor call generated with Python.
>
Why would we want to do that?

jgosmann:
> I would leave the scope of this PR to focus on exporting the layout to SVG and in the future address the conversion to PDF.

Yeah, I never meant to imply, that this should be part of this PR.

> Why would we want to do that?

To only have one place that defines the CSS (e.g., to avoid confusion when debugging).

tcstewar:
This looks good to me, other than a block of code in ```modal.js``` where the only change is some spacing, due to some commits that removed and then replaced that code.  @tbekolay , my feeling is if that's fixed and then we maybe squash some commits together, this should be all set.  Does that make sense to you?  And perhaps the squashing would resolve the merge conflicts...

tbekolay:
Yes, I would definitely remove the spacing change and squash the commits. Dunno if it'll resolve the conflicts altogether, but it should be easier to deal with! LMK if you want me to handle the rebase.

tcstewar:
> Yes, I would definitely remove the spacing change and squash the commits. Dunno if it'll resolve the conflicts altogether, but it should be easier to deal with! LMK if you want me to handle the rebase.

If you could take a stab at that, that'd be great.  :)  Thank you!

tbekolay:
OK, rebased and squashed. I haven't tested it in the squashed state though, so would be good to re-test before merging.

tcstewar:
> OK, rebased and squashed. I haven't tested it in the squashed state though, so would be good to re-test before merging.

Thanks!  I'll do that now....

tcstewar:
Functions great for me, other than a small typo while merging.  I'll wait for the tests to complete and then merge.  Thank you for the rebasing/squashing!

Issue #631, opened 2015-12-14
tab to enter different values when setting range?


studywolf:
I'm not sure if this would be easier, but it feels super unnatural to me to set multiple values in a gui using a comma to separate, this might be more intuitive having different boxes you can tab between?

PR #657, opened 2016-02-06
allowing tabbing to set multiple values for input box

bjkomer:
Addresses #631 

When entering values for sliders or changing their range, you can now optionally press tab to go to the next value instead of typing in the commas. This will automatically put the comma in for you, and will automatically fill in the previous value for an index if you leave it blank and press tab.

### DIFF
--- a/nengo_gui/static/modal.js
+++ b/nengo_gui/static/modal.js
@@ -262,7 +262,7 @@ Nengo.Modal.prototype.single_input_body = function(start_values, label) {
         }
         //Allow tabs to enter in default values
         if ((event.keyCode || event.which) == 9) {
-            var values = $("#singleInput").attr('placeholder').split(",");//start_values.split(", ");
+            var values = $("#singleInput").attr('placeholder').split(",");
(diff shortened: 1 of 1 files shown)


nengo_gui/static/modal.js
```diff
@@ -253,13 +253,47 @@ Nengo.Modal.prototype.single_input_body = function(start_values, label) {
...
+    
+    $(".controls").on('keydown', '#singleInput', function(event) {
+        //Allow the enter key to submit
         if (event.which == 13) {
             event.preventDefault();
             $('#OK').click();
         }
+        //Allow tabs to enter in default values
+        if ((event.keyCode || event.which) == 9) {
+            var values = $("#singleInput").attr('placeholder').split(",");//start_values.split(", ");
```
tcstewar: Should the comment on the end of this line be removed?

bjkomer: Good catch, that shouldn't be there.

tcstewar:
Looks good to me!  Merging...

Issue #423, opened 2015-06-15
Making a slider for a passthrough node gives `NaN`


tbekolay:
If you make a slider from a passthrough node, the slider pops up with the value `NaN`. In a sense, this is correct, but in another sense, pressing play will crash the simulation with `FloatingPointError: invalid value encountered in less` while doing `LIF.step_math`.

Possible solutions:

1. Don't allow sliders for passthrough nodes (this would be a bummer).
2. Pop them up in some kind of disabled state; update them once the simulation starts.
3. Default to 0. Update once the simulation starts.

Seanny123:
I would prefer option 3, but I'm not totally clear why initialising to NaN is a technically correct in the first place?

tbekolay:
Because a passthrough node has output of `None` because its output is not defined until it has some input. `None` is not a number, but Javascript doesn't have a `None`, so it would be `null`, `undefined`, or `NaN`, they're all kind of correct.

Seanny123:
In that case, I find option 2 to be more intuitive. Shall I implement it or
should this be discussed further?

tcstewar:
> In that case, I find option 2 to be more intuitive. Shall I implement it or
> should this be discussed further?

I'd still stick with option 3, as it's *much* simpler, and this aspect about ```None``` is a fairly esoteric implementation detail, so I'd argue that ```0``` is more intuitive than some sort of weird disabling thing that isn't present in any other case.

Seanny123:
I've changed my mind and now agree with the simplicity presented by option
2.

tcstewar:
> I've changed my mind and now agree with the simplicity presented by option
> 2.

Do you mean 3?  3 seems simpler than 2 to me....

Seanny123:
Yeah, I totally meant 3. Gosh darn number dyslexia.

tcstewar:
Fixed by #658

PR #658, opened 2016-02-06
passthrough nodes now have working sliders

celiasmith:
Fixes passthrough nodes to not display NaN and allow you to manipulate them.  See discussion in #423

### DIFF
--- a/nengo_gui/components/netgraph.py
+++ b/nengo_gui/components/netgraph.py
@@ -10,10 +10,10 @@ import json
 
 from nengo_gui.components.component import Component
 from nengo_gui.components.value import Value
+from nengo_gui.components.slider import Slider
 from nengo_gui.modal_js import infomodal
 import nengo_gui.user_action
 import nengo_gui.layout
-
 
 class NetGraph(Component):
     """Handles computations and communications for NetGraph on the JS side.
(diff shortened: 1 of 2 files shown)


tcstewar:
It'd be nice to reorganize this so it doesn't use this new ```node._passthrough```.  Perhaps make a dummy function in Slider and do ```node.output = Slider.passthrough_function``` (and then in Netgraph do ```if node.output is None or node.output is Slider.passthrough_function```).

tcstewar:
Bug: if you make a passthrough slider, and the reload the page, it gets NaN values.

nengo_gui/components/netgraph.py
```diff
@@ -487,6 +487,8 @@ def expand_network(self, network, client):
         for ens in network.ensembles:
             self.create_object(client, ens, type='ens', parent=parent)
         for node in network.nodes:
+            if node.output is None:
+                node.output = Slider.passthrough_fcn
```
tcstewar: Do we need this to happen in the NetGraph?  I'd only change the ones that there are Sliders for, and Slider's already taking care of this.

nengo_gui/components/netgraph.py
```diff
@@ -527,7 +529,7 @@ def get_extra_info(self, obj):
         '''
         info = {}
         if isinstance(obj, nengo.Node):
-            if obj.output is None:
+            if obj.output is Slider.passthrough_fcn:
```
tcstewar: I think this should be ```if obj.output is None or obj.output is Slider.passthrough_fcn```

tcstewar:
Looks good to me!  :)

Issue #91, opened 2015-04-08
Speed throttle


tcstewar:
Have a way to set the desired speed

Seanny123:
Where do you envision this fitting into the user interface? Click on speed to expose a context menu to adjust?

tcstewar:
> Where do you envision this fitting into the user interface? Click on speed to expose a context menu to adjust?

I've been thinking about this a bit, and about various ways to implement it.  In Nengo 1.4, we did something pretty much like the context menu approach, with a few discrete options (fastest (no throttle), 1x, 0.1x, 0.01x, etc).  So that's sorta okay, but it always felt a bit weird to me.

So for this one, I've been kinda imagining an actual throttle.  That is, a slider where all the way to the right is full speed, and all the way to the left is stop, and in between is some monotonic delay system.  I'm just not sure whether it should be on a log-scale or linear.  :)  And the idea in my head is that slider is just always there, right underneath the speed measurement -- that way the slider itself doesn't need any numbers on it, as the speed measurement itself tells you how fast you're going.

PR #661, opened 2016-02-06
Speed throttle

tcstewar:
This should resolve issue #91 .  It is a speed throttle! 

When you start, if you're going faster than 1.0x real time, it'll slow you back down to 1.0.  There's a slider in the bottom bar that lets you adjust from full speed (all the way to the right) to pretty much stopped (all the way to the left).

Things still to do:

- [ ] test on different OSes
- [ ] disable it for nengo_spinnaker
- [ ] check if there's something nicer to do than the busy-loop to accurately slow things down.


nengo_gui/components/sim_control.py
```diff
@@ -73,14 +77,56 @@ def control(self, t):
...
+        # if we have a desired rate, do a simple P-controller to get there
+        if self.target_rate is not None:
+            rate_error = self.rate - self.target_rate
+            delta = rate_error * 0.0000002
+            self.delay_time += delta
+
+        if self.delay_time < 0:
+            self.delay_time = 0
+        if self.delay_time > 0.5:
+            self.delay_time = 0.5
```
tbekolay: Could replace with `self.delay_time = np.clip(self.delay_time, 0, 0.5)`

nengo_gui/components/sim_control.py
```diff
@@ -73,14 +77,56 @@ def control(self, t):
...
+        if self.delay_time > 0:
+            self.busy_sleep(self.delay_time)
+
         # Sleeps to prevent the simulation from advancing
         # while the simulation is paused
         while self.paused and self.page.sim is not None:
             time.sleep(0.01)
             self.last_tick = None
 
+    def busy_sleep(self, delay_time):
```
tbekolay: Is `time.sleep` too imprecise compared to the busy loop? You could just sleep for `delay_time`.

tcstewar: Way too imprecise.  And on Linux, it can't do anything < 10ms.

tcstewar: (my plan for something to try is to use time.sleep(), but accurately measure how much time it actually slept for, and use that to adjust how much time I'll sleep next time.)

tcstewar:
There still seem to be some bugs where if you slow things down, it has difficulty recovering the speed (i.e. going back up to 50%).  This showed up on both OSX machines tested.

tcstewar:
The busy-loop has been replaced with a normal sleep call, but I track the actual amount of time slept and use it to modify future calls to sleep.  This seems fairly robust and way nicer to the processor.

tcstewar:
> There still seem to be some bugs where if you slow things down, it has difficulty recovering the speed (i.e. going back up to 50%). This showed up on both OSX machines tested.

I think this is now fixed with a change to how I'm converting the slider value into a delay time.  There shouldn't be a giant memory effect any more (which was due to computing the delay_time based on the computed rate, which is itself a low-pass filtered value that depends on delay_time....)

tbekolay:
I just tried this on my Debian machine, and it is very smooth! I don't get the issue with being stuck at slow speeds anymore, it speeds up as expected. I'll try this on my Mac in a bit as well.

tbekolay:
Tried this on my Mac. Works great there too! I only tested with a few simple networks, but it definitely frees up the CPU if you're not running full speed, and doesn't get stuck when going down to slow speeds.

tcstewar:
I just tried is with nengo_spinnaker, and I think it's actually fine how it is right now without adding in special-case code to disable it.  Basically, it doesn't affect things at all, unless you drop the speed throttle all the way down, in which case it just slows down the sampling of the data coming back from spinnaker.  Otherwise it just sits there with the throttle all the way to the right (full-speed).  So I'm happy with that for version 0.2.

nengo_gui/static/sim_control.css
```diff
@@ -122,3 +122,24 @@
     padding: 0.2em !important;
     border: none !important;
 }
+
+#speed_throttle {
+    height: 10px;
+    width: 110px;
+    #background: yellow;
```
rsimmons1: Is there any reason to have the # in front of the background property? As far as I can tell its only for IE support.

rsimmons1:
Other than that it all looks good to me. Very useful.

Seanny123:
This pull request should also change the Tutorial Examples 01 to 04 so that they stop using that slowdown node.

nengo_gui/static/sim_control.css
tcstewar: Oops -- that's me forgetting what language I'm using and trying to comment out a line, and then forgetting to remove it before committing.  Oops...  removing that line now....

tcstewar:
> This pull request should also change the Tutorial Examples 01 to 04 so that they stop using that slowdown node.

Good idea.  I've removed it now from all the tutorials (00 to 05).  That seems to work automatically for the most part on my machine, although I wouldn't be surprised if there's still some situations where users have to manually slow the simulation down if the spike plots stop rendering well.

Issue #362, opened 2015-06-09
Add a home directory configuration option


tbekolay:
Right now, it's difficult to explain where your Nengo scripts are and why, because you have to be pretty comfortable with the terminal to realize that when you start `nengo_gui` it starts in a particular directory, which we use.

Instead, it would be nice if it didn't particularly matter where we started `nengo_gui`. So, how about we add a configuration option (accessible through the gear in the top toolbar) to set a directory to be where the file browser starts from. If the user doesn't change this, we can still start from the current directory, but if the user specifies something then clicking the folder should show the files in the specified directory.

PR #662, opened 2016-02-06
Script directory configuration option

tbekolay:
This adds a new option to the Configuration options. Originally, it looks like:

screen shot 2016-02-06 at 11 32 04 am

You can type in an absolute path to some directory.

screen shot 2016-02-06 at 11 32 23 am

The file browser will then change to be based on that directory.

screen shot 2016-02-06 at 11 32 39 am

This persists when you close the browser and reopen it. If you clear the text in the input field in the config options, you get back to a file listing from your current directory.

screen shot 2016-02-06 at 11 33 01 am

A few issues that may or may not be necessary to deal with before merging:

- [ ] No validation is done on the path given. If there are problems with the path, this is what happens:

  screen shot 2016-02-06 at 11 41 50 am

  It's a kind of okay error message, but it could be better. In particular, it would be good to style the window better.
- [ ] Typing in an absolute path is kind of annoying. Unfortunately, as far as I can determine from a bunch of stackoverflow threads, you can't get access to a single directory through a typical HTML5 file browser input. So, we would essentially have to do a separate jQueryFileTree to get access to the right directory. Could be better to do that, but it's hard to think about how that would look in the modal window...

This is based on the `local-storage` branch, so #649 should be merged before this PR is merged.

Addresses issue #362.


tcstewar:
Very nice.  This looks good to me, and I think it's worth merging and leaving improving the error message as a separate issue.

tcstewar:
Other than the merge conflicts, this looks good to go.  Should I take a stab at those?

tbekolay:
I can give it a rebase!

tbekolay:
OK, rebased but not tested.

nengo_gui/static/modal.js
```diff
@@ -267,22 +274,35 @@ Nengo.Modal.prototype.main_config = function() {
...
+        Nengo.config.scriptdir = sd;
+    });
+
     $('#config-backend').change(function () {
         sim.set_backend($('#config-backend').val());
     });
 
     //Allow the enter key to submit
-    $("#config-fontsize").keypress(function(event) {
+    submit = function(event) {
```
tcstewar: This should be "var submit = ..." so we don't get a global variable here

tcstewar:
> OK, rebased but not tested.

Just tested it; seems to be working for me!  Just that one tiny global variable thing I just noticed and commented on above.  :)

PR #663, opened 2016-02-06
Control time kept

bjkomer:
Allows the time kept on the time slider to be configurable and saved to the config. Addresses #643


nengo_gui/static/sim_control.js
```diff
@@ -332,6 +332,35 @@ Nengo.TimeSlider.prototype.reset = function() {
...
+    if (this.first_shown_time < this.last_time - this.kept_time) {
+        this.first_shown_time = this.last_time - this.kept_time;
+    }
+
+    x = this.kept_scale(this.first_shown_time);
+    Nengo.set_transform(this.shown_div, x, 0);
+
+    /** update the time axis display */
+    this.axis_g
+        .call(this.axis);
```
Seanny123: This is a bit silly to have on two lines.

Seanny123:
@tcstewar this feature is ready for review now that it's been brought up to date with master

tcstewar:
The kept_time value doesn't seem to persist.  If I change the kept_time parameter and then reload the page, it goes back to the old 4s value.  However, the config menu still has the changed value.....

nengo_gui/static/components/netgraph.js
```diff
@@ -70,6 +70,17 @@ Nengo.NetGraph = function(parent, args) {
...
+        get: function() {
+            return Nengo.config.kept_time;
+        },
+        set: function(val) {
+            if (val === this.kept_time) { return; }
+            Nengo.config.kept_time = val;
+            sim.time_slider.update_kept_time(val);
+        }
+    });
+
```
tcstewar: Why is this in Nego.NetGraph?  It isn't used by any of the NetGraph components...

Issue #641, opened 2016-02-01
Lorenz tutorial Simulation


rsimmons1:
Very randomly I've found that the equations in the lorenz simulation become stable (the lines on the graph stop moving) and the x-y graph shrinks into a point, is this suppose to be possible? I'm not sure how to recreate this issue at the moment.

tcstewar:
We should probably set a seed for this one, as yeah, it'll happen sometimes....

tcstewar:
Fixed by #664

PR #664, opened 2016-02-06
added seed to lorenz and a test

celiasmith:
Addresses #641.  The tutorial lorenz example now has a seed.  It is also tested to ensure it doesn't fade to a stable point.


Seanny123:
Took the liberty of fixing some formatting. Looks good to me now.

tcstewar:
Looks good to me, too.  I'll also note that the updated .cfg file gives a better layout for the tutorial.

Issue #633, opened 2016-01-05
Backspace Navigation


rsimmons1:
Hitting backspace and accidentally redirecting the page is a tedious issue. It occurs when the user has last clicked on the netgraph, but then hits backspace to edit their code without realizing that the IDE was not selected. This issue occurs on chrome but not firefox. In general the GUI does not seem friendly to browser navigation versus internal GUI navigation so I think it would be best if backspace as navigation was disabled.

Seanny123:
Resolved via merging #665

PR #665, opened 2016-02-06
disabling backspace navigation

bjkomer:
Addresses #633


Seanny123:
Works for me on Chrome and Ubuntu. Firefox doesn't even have that keyboard shortcut in the first place.

PR #669, opened 2016-02-06
stopping sim control from stealing all of the events

### LINKED ISSUE #667: Hotkeys don't work while focus is on the play button
If you click the play button, and then try some hotkey (such as Ctrl-M for minimap) the hotkey will not work until the focus leaves the button.

bjkomer:
Fixes #667


Seanny123:
LGTM.

Issue #438, opened 2015-06-17
Changing the .vectors attribute of a vocabulary appears to break the semantic pointer display


teslar:
When I create a vocabulary, and then change the vectors in it by directly messing with its .vectors attribute, displaying the semantic pointer in a buffer that uses this vocabulary gets really funky:
1. When running the model, it displays a bunch of keys even though no input has been given
2. When an input is given:
   - It does not set itself to that input
   - Some of the pointers in the display are now preceded by integers
     - Said integers fluctuate between some values between 0 and 9 (e.g. the digit preceding YELLOW in the screenshot below will not always be a 7; it just happened to be when I took the screenshot)
     - They only seem to appear at the beginning of the semantic pointer's key

Example code that does this, and screenshot:
![screen shot 2015-06-17 at 19 14 06](https://cloud.githubusercontent.com/assets/7680856/8220934/b7a25dbc-1526-11e5-837c-e7587d97e52e.png)

``` python
import nengo
import nengo.spa as spa
import numpy as np

D = 32

set_noise = 10

vocab = spa.Vocabulary(D)

vocab.parse("RED")
vocab.parse("BLUE")
vocab.parse("GREEN")
vocab.parse("YELLOW")
vocab.parse("PURPLE")
vocab.parse("MAGENTA")
vocab.parse("CYAN")
vocab.parse("BLACK")
vocab.parse("WHITE")
vocab.parse("GREY")
vocab.parse("PINK")

for x in vocab.keys:
    idx = vocab.keys.index(x)

    if not set_noise == 0:
        vocab.vectors[idx] = vocab.vectors[idx]  + np.random.normal(loc=0, scale=set_noise, size=D) 

model = spa.SPA()

with model:
    model.j = spa.Buffer(D, vocab=vocab)
```

pblouw:
A pull request resolving the issue of integers appearing unexpectedly has been submitted (the problem arises here because the norms of the modified vectors have become very large in comparison to standard use cases). The problem of the spa module not taking on value of the provided input is due to the fact that vocab objects store semantic pointer information in two places, while modifying the vocab's .vectors attribute only changes this information in one place. A pending change to vocab objects that allows for consistent modifications of the vector underlying a particular SP will be used to solve this problem.

krisfrenette:
I believe that this is fixed, if not please reopen

PR #670, opened 2016-02-06
Prevents random numbers from appearing in SP cloud plot.

pblouw:
If the .vectors attribute of a vocab object is modified to include one or more vectors with a very large norm, the dot products used to calculate the similarities depicted in semantic pointer plots can get big. Currently, the gui assumes these dot products are single digit, and accordingly gets confused and makes plots with extra numbers in them if they aren't. This pull request fixes this problem (issue #438) by clipping any dot products above 9.99.

### DIFF
--- a/nengo_gui/components/pointer.py
+++ b/nengo_gui/components/pointer.py
@@ -67,7 +67,7 @@ class Pointer(SpaPlot):
                                np.array(vocab.key_pairs)[over_threshold])
             matches = itertools.chain(matches, pair_matches)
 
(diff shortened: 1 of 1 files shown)


Seanny123:
Seems reasonable to me, especially since the whole point of the cloud is to give an approximate value.

Seanny123:
What code should I run to test it out on my machine?

pblouw:
For testing, the code sample posted in issue #438 should longer produce the error mentioned in that issue.

Issue #361, opened 2015-06-08
__file__ is not defined


drasmuss:
When I open `nengo_gui` via

``` python
import nengo_gui
nengo_gui.Viz(__file__).start()
```

it opens up the GUI fine, and the model loads properly.  But then if I trigger a reload (e.g. by changing the file), I get this error:

```
Traceback (most recent call last):
  File "...\nengo_gui\nengo_gui\components\netgraph.py", line 64, in _reload
    exec(code, locals)
  File "<string>", line 77, in <module>
NameError: name '__file__' is not defined
```

So it seems like the global `__file__` doesn't exist when you `exec` the code like that.  This is probably something specific to my setup (it seems like it would have been noticed before if it wasn't).

Edit: to be clear, line 77 is the `nengo_gui.Viz(__file__).start()` line

tbekolay:
I've gotten this error in the past, and the fix was to put this in a `if __name__ == '__main__'` block. There was a PR to change this though... I think the `exec` is in a `try / except` in which the `except` has no qualification so the `NameError` should be caught by that. I'll try to reproduce.

xchoo:
Does passing the local locals into the constructor fix the problem?

```
nengo_gui.Viz(__file__, locals=locals(), interactive=False).start()
```

tcstewar:
Weird...  it's supposed to be filling in ```__file__``` automatically for exactly this reason:

https://github.com/nengo/nengo_gui/blob/master/nengo_gui/viz.py#L203

Oh, oops, I see...  it's not doing that in the ```netgraph.py``` reload case.  That same line should be added there.

tcstewar:
This has been working for a while (since the server refactoring), so I'm closing this....  :)

Issue #554, opened 2015-09-11
Refactor nengo_gui.gui.GUI ?


Seanny123:
While creating the flow documents for the presentation I'm making in September, I noticed some weird code-flow in `nengo_gui.gui.GUI`.

Basically, this is what I understand of the start-up so far:
- main.py parses arguments and starts instance of nengo_gui.gui.GUI
- GUI starts nengo_gui.server.Server which inherits from SWI
- SWI then calls various server functions (`prepare_server` and `begin_lifecycle`) from GUI

Why does `nengo_gui.gui.GUI` need to act as a wrapper around the methods `prepare_server` and `begin_lifecycle`? Should this be refactored to eliminate the circular-like function calls? I understand that these calls are also made by the iPython startup, but it's not like the `server` attribute of the `GUI` is private.

tcstewar:
I totally agree...  indeed, I think this is a duplicate of #456

jgosmann:
Resolved in #673.

PR #673, opened 2016-02-07
Server refactoring

### LINKED ISSUE #456: Refactor server

jgosmann:
### Todo list:

* [x] open browser at start
* [x] Check IPython integration (probably needs some refactoring as startup of server changed somewhat)
* [x] Documentation, especially `InteractiveGUI` docstring is outdated
* [x] disable log messages, but make sure we get exceptions
* [x] user authentication
* [x] SSL encryption for the user authentication (it's 2016 we should not transmit passwords in plain text)
* [x] remaining TODOs and FIXMEs in the code
* [x] Make sure command-line arguments like backend work properly
* [x] Make sure that `interactive` attribute on GUI object isn't need anywhere. Same for other changed attributes.
* [x] `ws_default` in the request handler is a bit longish and could potentially be refactored
* [x] Remove old `server.py` and `swi.py` and move files in  `backend` to a more appropriate place (?)

### Todo list for later:

* [x] Python 3 compatibility: It seems that the test suite itself cannot be run with Python 3 at the moment (e.g., Py 2.7 style print in `conftest.py`). These issues should be fixed first and are not in the scope of this PR.
* [ ] Transmitting POST data as multipart. (Not sure, if we actually need this, but at least it would make the implementation complete)
* [ ] More tests?

### Some general comments

* The code is still based on the Python standard `BaseHTTPServer` stuff. That simplifies the implementation somewhat, but there are also points where it is a bit awkward because when we deal with WebSockets (they have to be registerd with the server to be able to close them properly on server shutdown). I still think that the refactoring improves things because it more clearly delineates what the server class and request handler do. Previously starting the server was implemented in the request handler and there where other weird cross-references.
* I'm trying to remove interactive from the server and move differences in interactive/non-interactive mode outside of the server.
* The new request handler does not pass arguments to the submethods, but sets class attributes. I don't necessarily think that this is the best style, but it is more consistent with what the `BaseHTTPRequestHandler` is doing anyways.
* There are probably a bunch of other decisions I made for (hopefully) good reasons, but can't remember right now.

Resolves #456 and #554


jgosmann:
I think, this could be ready for review. Not sure if I will work on the remaining to-dos in the next few days. But I think all the important features are implemented (except Python 3 compatibility). The history might need some squashing and cleanup.

jgosmann:
Rebased to current master and fixed compatibility to Python 3.

jgosmann:
Rebased to current master.

nengo_gui/tests/test_basic_functionality.py
```diff
@@ -17,6 +17,7 @@ def test_basic_functionality(driver, test_file):
 		#Test page response by clicking the reset button and applying new code to ace-editor
 
 		tt.reset_page(driver)
+		return
```
tcstewar: Why this return?

nengo_gui/gui.py
```diff
@@ -1,111 +1,103 @@
...
+    port : int
+        Port to listen on.
+    password : str, optional
+        Password required to connect to the backend.
+    """
+
+    def start(self):
+        protocol = 'https:' if self.server.settings.use_ssl else 'http:'
+        print("Starting nengo server at %s//%s:%d" %
+              (protocol, self.server.server_name, self.server.server_port))
```
tcstewar: On my windows install, ```server_name``` ends up being the name of the computer, but it doesn't end up resolving (it tries to go to ```http://amygdala:8080/``` and I get a "This site can't be reached" message.)

just going to localhost should be much more reliable, as it's always going to be the local computer.

tcstewar:
When I run this in windows, I get two errors.  First, it tries to connect to ```http://COMPUTER_NAME:8080``` which doesn't work.  If I change that to ```http://localhost:8080``` then I get:

```
Starting nengo server at http://amygdala:8080
WARNING:nengo_gui.server:Error response (403): Forbidden
  File "c:\users\terry\documents\github\nengo_gui\nengo_gui\server.py", line 233, in do_GET
    self.handle_upgrade()
  File "c:\users\terry\documents\github\nengo_gui\nengo_gui\server.py", line 260, in handle_upgrade
    self.upgrade_to_ws()
  File "c:\users\terry\documents\github\nengo_gui\nengo_gui\server.py", line 282, in upgrade_to_ws
    raise Forbidden()
Forbidden: Forbidden
WARNING:nengo_gui.server:Error response (403): Forbidden
  File "c:\users\terry\documents\github\nengo_gui\nengo_gui\server.py", line 233, in do_GET
    self.handle_upgrade()
  File "c:\users\terry\documents\github\nengo_gui\nengo_gui\server.py", line 260, in handle_upgrade
    self.upgrade_to_ws()
  File "c:\users\terry\documents\github\nengo_gui\nengo_gui\server.py", line 282, in upgrade_to_ws
    raise Forbidden()
Forbidden: Forbidden
WARNING:nengo_gui.server:Error response (403): Forbidden
  File "c:\users\terry\documents\github\nengo_gui\nengo_gui\server.py", line 233, in do_GET
    self.handle_upgrade()
  File "c:\users\terry\documents\github\nengo_gui\nengo_gui\server.py", line 260, in handle_upgrade
    self.upgrade_to_ws()
  File "c:\users\terry\documents\github\nengo_gui\nengo_gui\server.py", line 282, in upgrade_to_ws
    raise Forbidden()
Forbidden: Forbidden
WARNING:nengo_gui.server:Error response (403): Forbidden
  File "c:\users\terry\documents\github\nengo_gui\nengo_gui\server.py", line 233, in do_GET
    self.handle_upgrade()
  File "c:\users\terry\documents\github\nengo_gui\nengo_gui\server.py", line 260, in handle_upgrade
    self.upgrade_to_ws()
  File "c:\users\terry\documents\github\nengo_gui\nengo_gui\server.py", line 282, in upgrade_to_ws
    raise Forbidden()
Forbidden: Forbidden
```

nengo_gui/server.py
```diff
@@ -1,171 +1,511 @@
...
+    def set_timeout(self, timeout):
+        self.socket.settimeout(timeout)
+
+    def set_blocking(self, flag):
+        self.socket.setblocking(flag)
+
+    def _read(self):
+        try:
+            self._buf = self._buf + bytearray(self.socket.recv(512))
+        except ssl.SSLWantReadError:
```
tcstewar: This only exists in Python 2.7.9 and above

https://docs.python.org/2.7/library/ssl.html#ssl.SSLWantReadError

tcstewar:
Changing the backend doesn't seem to work.  I get this error:

```
ERROR:root:Error processing: u'backend:nengo_distilled'
  File "c:\users\terry\documents\github\nengo_gui\nengo_gui\guibackend.py", line 236, in _handle_ws_msg
    component.message(msg.data)
  File "c:\users\terry\documents\github\nengo_gui\nengo_gui\components\sim_control.py", line 205, in message
    self.page.settings.backend = msg[8:]
AttributeError: can't set attribute
```

tcstewar:
Sliders current'y don't work.  This line in ```components/slider.py```  needs a small change from

```
        elif page.backend == 'nengo_spinnaker':
```

to 

```
        elif page.settings.backend == 'nengo_spinnaker':
```

nengo_gui/server.py
jgosmann: Any idead what happens in earlier versions of Python?

jgosmann:
> Changing the backend doesn't seem to work

It seems to work for me. At least with the two commits that I added to address two other comments of yours.

nengo_gui/gui.py
jgosmann: It is not clear to me why this happens. Can you check the value of `self.server.settings.listen_addr`?

tcstewar: > Can you check the value of self.server.settings.listen_addr?

I get ```('localhost', 8080)```

jgosmann: Can we just drop support for Windows? It's doing stupid stuff. ;)

tcstewar:
> It seems to work for me. At least with the two commits that I added to address two other comments of yours.

Sorry, I wasn't clear -- I meant changing the backend via the config menu, rather than from the command line.  It looks to me like PageSettings is a named tuple, and so can't be changed....

jgosmann:
Still seems to work for me ...

jgosmann:
Nevermind, it does not work

nengo_gui/server.py
tcstewar: It looks like it's just ```SSLError```, with an ```errno == 2``` to check for that particular error:

https://github.com/slackhq/python-slackclient/pull/30/commits/a1428a30b7bf5d63b3217ed07995f5944e66caf5

tcstewar:
(btw, this code is *way* better than the original.  Thank you for doing this rewrite!)

nengo_gui/gui.py
tcstewar: Sure, as long as I can drop support for OS X as well....  ;)

celiasmith: And, I think we should drop support for linux ... so now all the 
problems are fixed!

tcstewar wrote:

jgosmann: The problem should be fixed with the last commit.

jgosmann:
> When I run this in windows, I get two errors. First, it tries to connect to http://COMPUTER_NAME:8080 which doesn't work. If I change that to http://localhost:8080 then I get:

Should also be fixed now.

nengo_gui/server.py
jgosmann: should also be fixed

jgosmann:
I have to rething the WebSocket `server_name` fix ...

jgosmann:
Note to myself:
* [ ] Check external SSL connections work
* [ ] Problem: listen_addr might be an ampty string in the SSL case, which can lead to problems using it as `server_name` and in the access URL.

jgosmann:
All those problems should now be fixed. I fixed the printed and opened URL to 'localhost'. I'm not completely happy with that because when external access is allowed it would be more useful to have the actual domain name there, but there's probably no way to get this working in all circumstances in Windows.

nengo_gui/guibackend.py
```diff
@@ -0,0 +1,390 @@
...
+    def __init__(
+            self, model_context, server_settings=GuiServerSettings(),
+            page_settings=nengo_gui.page.PageSettings()):
+        if nengo_gui.exec_env.is_executing():
+            raise nengo_gui.exec_env.StartedGUIException()
+        self.settings = server_settings
+
+        server.ManagedThreadHttpServer.__init__(
+            self, self.settings.listen_addr, GuiRequestHandler)
+        if sel.settings.listen_addr[0] != '':
```
tcstewar: Typo: this should be ```self```

nengo_gui/server.py
```diff
@@ -1,171 +1,519 @@
...
+            valid_srv_addrs.append(socket.getfqdn())
 
-        reset_cfg = reset == 'True'
+        try:
+            origin = urlparse(self.headers['Origin'])
+            assert origin.netloc.lower() in valid_srv_addrs
+        except KeyError:
+            raise Forbidden()
+        except AssertionError:
+            raise Forbidden()
```
tcstewar: On windows, this still raises an exception for me.  When it happens, ```origin.netloc``` is ```localhost:8080``` and ```valid_srv_addrs``` is ```['amygdala:8080', 'amygdala:8080']```    ("amygdala" is my computer name)

tcstewar:
Hmm... one strange bug that I'm not sure what to do with.

It looks like the websockets take much longer to form a connection in this version, as opposed to the previous version.  This causes a slower initial display, including about a second where the graph appears at the wrong scale, before it corrects itself.  This isn't too bad, but it also means that I get javascript errors when generating new graphs if I try to move them within a second of creating them.  The error is:

```
component.js:268 Uncaught InvalidStateError: Failed to execute 'send' on 'WebSocket': Still in CONNECTING state.
```

Thoughts?

tcstewar:
Ah, I found the line that's slowing things down for me.  It's this in server.py:

```python
valid_srv_addrs.append('{0}:{1}'.format(socket.getfqdn(), self.server.server_port))
````

That's taking about a second to execute on my machine, which slows everything down and causes the javascript connection problems I mentioned above.

File: nengo_gui/server.py (hunk 5 above)
Comment (jgosmann) [inline nengo_gui/server.py]: That was probably related to the typo. :/

**Comment by jgosmann** [comment]
It's getting too many fixup commits and I will squash them now.

File: nengo_gui/server.py (hunk 5 above)
Comment (tcstewar) [inline nengo_gui/server.py]: Nope -- I had to fix that ```sel``` -> ```self``` typo locally in order to run anything.  The AssertionError is still present in the current version.

**Comment by jgosmann** [comment]
I completely changed how the expected origins for the websockets are determined. It uses the transmitted host header field at login time or localhost if no login is required (external connections are only allowed with password anyways). I hope that this works for all systems (I have only tested Linux and no external connections so far).

I noticed that sometimes threads seem to crash on shutdown which produces ugly output. This seems to happen a lot more with the recent changes. I'll try to figure out why, but not sure if I can. (Because it is at shutdown and in a separate thread I'm not getting a useful traceback).

**Comment by jgosmann** [comment]
It's none of the request threads that is causing the exception.

**Comment by jgosmann** [comment]
Arg ... now I suddenly cannot reproduce it anymore. My suspicion is that it might be the runner thread for the simulation that isn't shut down properly.

**Comment by tcstewar** [comment]
> I hope that this works for all systems (I have only tested Linux and no external connections so far).

Working great on windows now!  The timing problem is gone, and it starts up perfectly.

One new bug under Python3, though.  When I open the file browser, it fails and I get this:

```
  File "c:\users\terry\documents\github\nengo_gui\nengo_gui\server.py", line 235, in do_GET
    self.http_GET()
  File "c:\users\terry\documents\github\nengo_gui\nengo_gui\server.py", line 251, in http_GET
    response = getattr(self, command)()
  File "c:\users\terry\documents\github\nengo_gui\nengo_gui\guibackend.py", line 92, in auth_checked
    return fn(inst)
  File "c:\users\terry\documents\github\nengo_gui\nengo_gui\guibackend.py", line 158, in browse
    b'<a href="#" rel="%s">%s</a></li>' % (ex_tag, ex_html))
TypeError: unsupported operand type(s) for %: 'bytes' and 'tuple'
```

I tried replacing that with a ```b''.join([...])``` and that works great for that line, since everything already bytes, but a few lines below that we need to combine e, ff, and f into the bytes, so just using a join doesn't work....

**Comment by tcstewar** [comment]
Does the server refactoring support ipython notebook integration?  Right now when I try it I get this:

```python
import nengo_gui.ipython
nengo_gui.ipython.IPythonViz(model)
```

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-3-1ef27bce4c06> in <module>()
      1 import nengo_gui.ipython
----> 2 nengo_gui.ipython.IPythonViz(net)

/home/tcstewar/github/nengo_gui/nengo_gui/ipython.py in __init__(self, model, cfg, height)
     36             cfg = get_ipython().mktempfile()
     37 
---> 38         self._server_thread, server = self.start_server(cfg, model)
     39         self.port = server.server.server_port
     40 

/home/tcstewar/github/nengo_gui/nengo_gui/ipython.py in start_server(cls, cfg, model)
     70         page_settings = nengo_gui.page.PageSettings(
     71             filename_cfg=cfg, editor_class=nengo_gui.components.editor.NoEditor)
---> 72         server = nengo_gui.gui.GUI(model_context, page_settings, port=0)
     73         server_thread = threading.Thread(target=server.start)
     74         server_thread.start()

TypeError: __init__() got an unexpected keyword argument 'port'
```

jgosmann:
> My suspicion is that it might be the runner thread for the simulation that isn't shut down properly.

I could verify that it is the runner thread that exits with the exception.

jgosmann:
The problem with the exception on killing the runner thread should be fixed now. I also added a commit to fix the Python compatibility (but not sure if that's working or sufficient, haven't tested it yet).

IPythonViz should be supported, I will have to look into this.

tcstewar:
> I also added a commit to fix the Python compatibility (but not sure if that's working or sufficient, haven't tested it yet).

That fixes that one line, but the two after it are still a problem:

```
  File "/home/tcstewar/github/nengo_gui/nengo_gui/server.py", line 235, in do_GET
    self.http_GET()
  File "/home/tcstewar/github/nengo_gui/nengo_gui/server.py", line 251, in http_GET
    response = getattr(self, command)()
  File "/home/tcstewar/github/nengo_gui/nengo_gui/guibackend.py", line 92, in auth_checked
    return fn(inst)
  File "/home/tcstewar/github/nengo_gui/nengo_gui/guibackend.py", line 171, in browse
    b'<a href="#" rel="%s/">%s</a></li>' % (ff, f))
TypeError: unsupported operand type(s) for %: 'bytes' and 'tuple'
```

And in this case, the ff is a string, not bytes, so we can't just concatenate them, and at this point I get confused by unicode....  :)

tcstewar:
I fixed the python3/bytes encoding issues, and so now everything seems to work in Python3.  Doing some more testing now, but I think the only remaining issue is the IPythonViz thing...

nengo_gui/guibackend.py
```diff
@@ -166,14 +166,19 @@ def browse(self):
 
         for f in sorted(os.listdir(path)):
             ff = os.path.join(path, f)
+            ff = ff.encode('utf-8')
```
jgosmann: I'd prefer to have this in one line by appending the `.encode('utf-8')` to the previous line.

nengo_gui/guibackend.py
```diff
@@ -166,14 +166,19 @@ def browse(self):
...
             if os.path.isdir(os.path.join(path, f)):
+                f = f.encode('utf-8')
                 r.append(b'<li class="directory collapsed">'
-                         b'<a href="#" rel="%s/">%s</a></li>' % (ff, f))
+                         b'<a href="#" rel="' + ff+ b'/">' + f + b'</a></li>')
```
jgosmann: missing a space after `ff`

jgosmann:
IPythonViz should be fixed.
Hopefully the tests will be fixed too.

nengo_gui/components/editor.py
```diff
@@ -0,0 +1,24 @@
...
+        super(Editor, self).__init__(component_order=-8)
+
+    def update_code(self, msg):
+        pass
+
+    def javascript(self):
+        return 'Nengo.disable_editor();'
+
+
+class NoEditor(Editor):
```
Seanny123: Quickly documenting that this is used for iPython would be appreciated.

nengo_gui/gui.py
```diff
@@ -1,111 +1,103 @@
...
+import nengo_gui
+from nengo_gui.guibackend import GuiServer
+
+
+class ServerShutdown(Exception):
+    """Causes the server to shutdown when raised."""
+    pass
 
 
 class GUI(object):
```
Seanny123: Noting that this is what you should use if you want to start the server from the commandline would be helpful. Otherwise the fact that it's a class is a bit surprising.

nengo_gui/components/ace_editor.py
```diff
@@ -1,40 +1,35 @@
...
         self.serve_code = True
 
     def update_client(self, client):
-        if not self.page.gui.interactive:
-            return
         while self.pending_messages:
-            client.write(self.pending_messages.pop())
+            client.write_text(self.pending_messages.pop())
         if self.serve_code:
             i = json.dumps({'code': self.current_code})
```
Seanny123: Merge this line with the next? It's only 71 characters total and I don't think it reduces clarity.

tcstewar: possibly, but slightly extraneous from this PR

nengo_gui/components/editor.py
jgosmann: That's only one use case. There might be more in the future?

nengo_gui/gui.py
jgosmann: > Noting that this is what you should use if you want to start the server from the commandline would be helpful.

But for starting the server from the command line `InteractiveGUI` should be used.

> Otherwise the fact that it's a class is a bit surprising.

Not exactly sure why this is surprising. What exactly was your expaction that is violated?

jgosmann:
@Seanny123 pointed some things out about backwards compatibility we should discuss, @tcstewar.

tcstewar:
> @Seanny123 pointed some things out about backwards compatibility we should discuss, @tcstewar.

Sounds good -- I'm in a meeting at a moment but should be done by 3:30

jgosmann:
Backwards compatibility restored. `GUI` now also has the comment that @Seanny123 requested, `BaseGUI` is still without a comment, but the reason that it exists is for `IPythonViz` which needs to start a server, but doesn't need all the signal handling stuff for shutdown.

tcstewar:
@jgosmann and @tbekolay : when I merge this into master (using the "For maintainers" instructions at https://docs.google.com/document/d/1YJUEAmrB6y_x0fDUYyRBSIPUNFLLsA3VQNKGRz8T8dI/edit#heading=h.4uw2xgy1ldsf ), how much of this history should I keep?  I assume all the "fixup!" commits should get done as fixup commits in the interactive rebase.  Should I be more selective than that?  Or is that good enough for now?

jgosmann:
If you use `--autosquash` with the rebase my `fixup!` commits will be automatically marked for squashing with the appropriate commits. There is still a bunch of other `Fix ...` commits where it might be a bit more difficult to determine the appropriate commit (usually I do a `git blame` on the changed lines for that). There is definitely even more that good be squashed for a better history, but that would probably require me to go through commits one by one and I won't have time before next week and when I have finished my tutorial preparations.

nengo_gui/gui.py
```diff
@@ -118,10 +118,25 @@ class GUI(InteractiveGUI):
...
+        determinined by running the script filename.  Its contents are used
+        to give useful names for the objects in the model.
+    editor : bool
+        Whether or not to show the editor
     """
-    def __init__(self, filename):
+    def __init__(self, filename=None, model=None, locals=None, editor=True):
+        if not editor:
+            ps = nengo_gui.page.PageSettings(
+                    editor_class=nengo_gui.components.editor.NoEditor)
```
jgosmann: @Seanny123 ... and here's suddenly another place where NoEditor is used. ;)

tcstewar:
> If you use --autosquash with the rebase my fixup! commits will be automatically marked for squashing with the appropriate commits. There is still a bunch of other Fix ... commits where it might be a bit more difficult to determine the appropriate commit (usually I do a git blame on the changed lines for that). There is definitely even more that good be squashed for a better history, but that would probably require me to go through commits one by one and I won't have time before next week and when I have finished my tutorial preparations.

Sounds fair...  I'll take a pass at it to squash the obvious ones, and then go ahead and merge.  It won't be the prettiest history, but it'll be better than what I'd been doing before....

nengo_gui/gui.py
tcstewar: Yup...  it's used by ctn_benchmarks if you use the --gui option

Seanny123:
@tcstewar were you planning on merging this before the summer school so it can be tested by all the students?

Seanny123:
As a side note, does this resolve #361

jgosmann:
I don't know, it might if we're lucky. 🍀

jgosmann:
We should get this merged ... one of the recent changes in master apparently broke this branch despite not giving any merge conflicts during rebase.

jgosmann:
Ah sorry, everything still works fine. Somehow parts of an old version got installed despite having checked out this branch.

Seanny123:
This code currently gives the following intermittent error. I'm not sure what causes it, but I thought I would paste it here anyways.
```
ERROR:nengo_gui.guibackend:Error during websocket communication.
  File "c:\nengo_summer_school\nengo_gui\nengo_gui\guibackend.py", line 230, in ws_default
    component.update_client(self.ws)
  File "c:\nengo_summer_school\nengo_gui\nengo_gui\components\netgraph.py", line 378, in update_client
    self.check_for_reload()
  File "c:\nengo_summer_school\nengo_gui\nengo_gui\components\netgraph.py", line 63, in check_for_reload
    self.reload()
  File "c:\nengo_summer_school\nengo_gui\nengo_gui\components\netgraph.py", line 86, in reload
    self._reload(code=code)
  File "c:\nengo_summer_school\nengo_gui\nengo_gui\components\netgraph.py", line 104, in _reload
    self.page.ace_editor.update_code(code)
AttributeError: 'Page' object has no attribute 'ace_editor'
```

jgosmann:
I believe ace_editor had to be changed to editor. I feel like I already fixed that problem ... but maybe I missed a spot.

On June 17, 2016 2:09:20 PM EDT, Sean Aubin <notifications@github.com> wrote:
>This code currently gives the following intermittent error. I'm not
>sure what causes it, but I thought I would paste it here anyways.
>```
>ERROR:nengo_gui.guibackend:Error during websocket communication.
>Traceback (most recent call last):
>File "c:\nengo_summer_school\nengo_gui\nengo_gui\guibackend.py", line
>230, in ws_default
>    component.update_client(self.ws)
>File
>"c:\nengo_summer_school\nengo_gui\nengo_gui\components\netgraph.py",
>line 378, in update_client
>    self.check_for_reload()
>File
>"c:\nengo_summer_school\nengo_gui\nengo_gui\components\netgraph.py",
>line 63, in check_for_reload
>    self.reload()
>File
>"c:\nengo_summer_school\nengo_gui\nengo_gui\components\netgraph.py",
>line 86, in reload
>    self._reload(code=code)
>File
>"c:\nengo_summer_school\nengo_gui\nengo_gui\components\netgraph.py",
>line 104, in _reload
>    self.page.ace_editor.update_code(code)
>AttributeError: 'Page' object has no attribute 'ace_editor'
>```
>
>---
>You are receiving this because you were mentioned.
>https://github.com/nengo/nengo_gui/pull/673#issuecomment-226840923

-- 

tcstewar:
> This code currently gives the following intermittent error. I'm not sure what causes it, but I thought I would paste it here anyways.

Hmm, it looks like the error message includes a line that doesn't exist in this PR.  This PR changes line 104 in components\netgraph.py to ```self.page.editor.update_code(code)```.  Is there a chance you had some sort of combination of branches, @Seanny123 ?

tcstewar:
I'm still trying to sort out what I want to do with the commit log with this PR.  After the reviewing discussion, I think it's pretty important to aim for having every commit in master passing the tests, but this PR is a fundamental reworking of the internals of the system, and so it just does not work unless it's all there.  So I'm strongly tempted to just squash it all into one giant commit.  I just don't see smaller parts that make sense on their own.

Are there any objections to me just squashing this all together?  thoughts, @tbekolay ?

tbekolay:
I don't know, a 3000 line diff in a commit is pretty rough... the commits in this branch

tbekolay:
I don't know, a commit changing 3000 lines is pretty rough, I don't think anyone would be able to read it... the commits in this branch all have useful messages and intents, though some should be fixups rather than standalone commits (e.g. 3e1b7e2fb9d97aa, a126b238f5eedbafc, 9a8c64be3941, 2c4ef93c4e61c1f). With the history as it is, those commits would likely have to be broken apart and squashed into multiple commits, which is difficult / annoying / time consuming, but is also kind of what you have to do if you want a really clean commit history. So it's mostly up to you / @jgosmann to decide whether that's worth it.

If the choice is between keeping the commit history as is or squashing it all into one, I would favour keeping it as is.

tcstewar:
> If the choice is between keeping the commit history as is or squashing it all into one, I would favour keeping it as is.

I'd be quite happy with keeping it as it is (with the fixups that currently exist being squashed, of course).  This is a pretty special PR, so I don't expect this sort of thing to happen often, so I'm not worried about the bad precedent being set.  :)

jgosmann:
I can squash some of the fixup comments, but I don't think I can make every commit pass the tests.

tbekolay:
Agreed; we can play things pretty fast and loose until version 1.0.0 (according to semantic versioning).

tcstewar:
> I can squash some of the fixup comments, but I don't think I can make every commit pass the tests.

That would be wonderful.  Don't worry about making every commit pass.  :)

Seanny123:
In regards to the intermittent error, I may have improperly switched branches.

jgosmann:
I did some cleanup of the history. There are a few commits at the end that should be merged with other commits, but it's hard to figure out with which other commits not knowing which commits make tests fail. So I was planning to run the tests on each commit and figure that out (also make as many commits pass the tests as possible). However, I'm on Firefox 47 which is incompatible with the latest stable Selenium version. [Apparently, one has to switch to another webdriver], but that one is only compatible with dev versions of Firefox (not the stablle 47). Bottom line: I can't run the tests right now without going through a lot of other hoops first.

So do we want to go ahead with the merge or should we get the tests to run first?

jgosmann:
(Also note, that the test coverage is still rather incomplete.)

jgosmann:
What's the timeline for getting this finally merged?

tcstewar:
> What's the timeline for getting this finally merged?

Let's go ahead and do this now.  This is important enough to get merged that I should just go ahead and do it, and it doesn't make sense to be waiting on a perfectly pristine history.   

If it sounds okay to you, @jgosmann , I'll merge this as it is.  The testing that happened over the summer school indicated that this is pretty solid.

jgosmann:
Sounds good to me.

tcstewar:
Done!  Thank you again, @jgosmann !

Issue #276, opened 2015-06-02
Moving and resizing expanded networks is extremely slow


bjkomer:
When you have a network with a bunch of stuff in it (e.g. the network in the net.py example) and it is expanded, it will move very slowly when you drag it around in Firefox. This doesn't happen when the network is collapsed.

jgosmann:
Seems fine on my Mac.

bjkomer:
Hmm... I had it happen on both my laptop and desktop (both linux). The effect is more noticeable when both subnet and subsubnet were expanded.

studywolf:
I saw it on my Linux laptop as well! I seen it!

tbekolay:
Just tried this on my Mac laptop... definitely also happens there. I think that this has a pretty simple explanation... that when you drag an expanded network with multiple things in it, we're updating the transform property of every single element in there. We should look at the snippet of code that does that; probably it can be rewritten to be more performant (since Chrome does an okay, but still not perfect, job of rendering it smoothly; it might just have our implementation slightly more tuned for speed).

tcstewar:
One thing to check for is whether we're updating transform multiple times...  that was the culprit in the past, but I thought I get rid of all the cases where I was accidentally doing that.

Seanny123:
Replicated on Chrome

tbekolay:
One thing that we can do to fix things (at least in some cases) is to set the `shape-rendering` CSS property on our root SVG element to `optimizeSpeed`, which should propagate to all of the shapes. See [this doc for more details](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/shape-rendering).

krisfrenette:
I just tested this in both Firefox, and Chrome and I am not seeing this issue anymore. I am going to close this as it may have been browser related and a lot of updates have been done since the original report of the issue.

PR #674, opened 2016-02-08
Optimize display

drasmuss:
This addresses #276 (lag on firefox when dragging around large networks).  It's still a little bit choppy when dragging a big network around, but significantly improved from what it was like before.

The first improvement comes from only updating the minimap when it is being displayed.  Before we were updating it all the time, so we were effectively doing twice as many updates/redraws as necessary.

The most significant change to the code-base here is that I modified the minimap logic so that the main object takes care of keeping the minimap object up to date (rather than the old method, where the overall netgraph took care of updating both the main objects and the minimap objects).  This means that there is only one place where we need to check if the minimap is open, rather than having to remember to put that check in everywhere we update the object.

The second improvement comes from reducing the number of jquery calls we're making, as those are kind of expensive.  Basically this just involves storing the jquery result in a local variable when we query it (in response to something changing), and then everything else just references that local variable instead of redoing the jquery.  (note: I only did this in `netgraph.js` and `netgraph_item.js`; there are likely other areas of the code that could benefit from the same treatment).

### DIFF
--- a/nengo_gui/static/components/netgraph_item.js
+++ b/nengo_gui/static/components/netgraph_item.js
@@ -745,10 +745,11 @@ Nengo.NetGraphItem.prototype.redraw_size = function() {
 };
 
 Nengo.NetGraphItem.prototype.get_width = function() {
+    if (this.minimap && !this.ng.mm_display) { return 1; }
+    
     if (this.fixed_width !== null) {
         return this.fixed_width;
     }
-
 
     if (this.minimap == false) {
         var w = $(this.ng.svg).width();
@@ -766,6 +767,8 @@ Nengo.NetGraphItem.prototype.get_width = function() {
 }
 
 Nengo.NetGraphItem.prototype.get_height = function() {
+    if (this.minimap && !this.ng.mm_display) { return 1; }
+    
     if (this.fixed_height !== null) {
         return this.fixed_height;
     }
@@ -796,6 +799,8 @@ Nengo.NetGraphItem.prototype.redraw = function() {
 /** determine the pixel location of the centre of the item */
 Nengo.NetGraphItem.prototype.get_screen_location = function() {
     // FIXME this should probably use this.ng.get_scaled_width and this.ng.get_scaled_height
+    if (this.minimap && !this.ng.mm_display) { return [1, 1]; }
+    
     if (this.minimap == false) {
         var w = $(this.ng.svg).width() * this.ng.scale;
         var h = $(this.ng.svg).height() * this.ng.scale;
--- a/nengo_gui/static/components/netgraph_item.js
+++ b/nengo_gui/static/components/netgraph_item.js
@@ -579,6 +579,10 @@ Nengo.NetGraphItem.prototype.remove = function() {
     if (this.depth == 1) {
         this.ng.scaleMiniMap();
     }
+    
+    if (!this.minimap) {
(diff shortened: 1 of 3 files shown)


drasmuss:
Note: depends on #675

nengo_gui/static/components/netgraph_item.js
```diff
@@ -548,15 +593,8 @@ Nengo.NetGraphItem.prototype.get_displayed_size = function() {
...
-        while (parent !== null) {
-            h_scale = h_scale * parent.size[0] * 2;
-            v_scale = v_scale * parent.size[1] * 2;
-            parent = parent.parent;
-        }
-
-        var w = this.size[0] * h_scale;
-        var h = this.size[1] * v_scale;
+    	var w = this.get_nested_width() * h_scale;
+    	var v = this.get_nested_height() * v_scale;
```
tcstewar: I think this should be ```var h``` rather than ```var v```....

drasmuss: Ah yeah, definitely.  The `v_scale` threw me off!

drasmuss:
Hang on on merging/reviewing this one, I just noticed some weird things happen with recurrent connections that I need to look at.

tcstewar:
> Hang on on merging/reviewing this one, I just noticed some weird things happen with recurrent connections that I need to look at.

Will do...  I was going through others and wanted to try this one out, and didn't notice that it wasn't marked as ready for review...  :)

drasmuss:
Actually it was just solved by fixing the thing you already pointed out, so good to go!

tcstewar:
Collapsing subnetworks seems to be broken.  To replicate, open up anything with subnetworks (e.g. ```tutorial/18-networks.py```), expand a network, and then collapse it again.  The subcomponents aren't removed and you get a javascript error of:

```
Uncaught NotFoundError: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
Nengo.NetGraphItem.remove @ netgraph_item.js:578
Nengo.NetGraphItem.remove @ netgraph_item.js:584
Nengo.NetGraphItem.collapse @ netgraph_item.js:505
(anonymous function) @ netgraph_item.js:326
```

drasmuss:
Should be fixed now.  That actually revealed another inefficiency; we were adding all the minimap objects to the `network.children` list, so then when we went to redraw the children we'd redraw all the main objects and minimap objects, and then we'd redraw all the minimap objects again.  So it's even faster now!  Bugs only make me stronger :rage3:.

tcstewar:
Well, I haven't found any other ways to break it, and all the changes look good to me, so I think this is ready for merging!

tcstewar:
(and thanks for going through that code and profiling things and figuring out where some of the slowdowns were!)

drasmuss:
I'd do that just for fun, optimization is like popping bubble wrap :heart_eyes:.

tcstewar:
> I'd do that just for fun, optimization is like popping bubble wrap .

Wow did we hire the right person.  ;)

Seanny123:
What did you use for profiling the code? Would it be worth documenting your methods somewhere?

drasmuss:
Just the built-in Firefox profiler.  Chrome has a built-in profiler as
well, although it isn't quite as fully featured.

PR #675, opened 2016-02-08
Fix remove logic

drasmuss:
This typo was causing us to remove and rebuild every item in the network whenever any object was removed.  In terms of efficiency per character, this has got to be a record breaking optimization :sunglasses:.


tcstewar:
AAAAAAAAAHHHHHHHHHHHHHHHHHH!!!!!!!!!!!!!!!!!!!!


Oh. My. Gods.   That is just....   wow.

Thank you for finding this...

PR #676, opened 2016-02-09
Toggle for auto-updating from code editor

tcstewar:
This is just a version of #651 without the horrible mess of a rebasing I accidentally did to it.

### DIFF
--- a/nengo_gui/static/ace.js
+++ b/nengo_gui/static/ace.js
@@ -50,6 +50,8 @@ Nengo.Ace = function (uid, args) {
     this.console.appendChild(this.console_error);
 
     this.save_disabled = true;
+    this.update_trigger = true; // if an update of the model from the code editor is allowed
+    this.auto_update = true; // automatically update the model based on the text
 
     //Setup the button to toggle the code editor
     $('#Toggle_ace').on('click', function(){self.toggle_shown();});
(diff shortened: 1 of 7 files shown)


tcstewar:
@tbekolay could you take a look at the small fix I made to ```config.js```?  The original code wasn't initializing the variable correctly for the case when default_val is set to true and there's nothing in the localstorage.

Other than that, this code looks good to me!

tcstewar:
(and for clarity, here's @bjkomer 's original message on this:)

Adding the option to disable automatic synchronization of the model and the code in the editor.

Which shortcut keys to use is still up in the air, currently using Ctrl-Shift-1 to toggle auto-updating, and Ctrl-1 to trigger an update. The toggle can also be set in the config, and the trigger with a tool bar button (which also doubles as an indicator of whether or not the code differs from the model).
Addresses #644 and also provides a way around #444 #550

Allowed Shift-Enter to run the model from the editor window (also works in the other windows). Addresses #639

drasmuss:
I think the issue is that if there is nothing in `localStorage` than it uses `default_val`, which is an actual boolean.

tbekolay:
Ah right, I see. I did a bit of googling after and apparently it is supposed to accept other types of data, but no browsers actually do it now.

But you're right, if the default is `true` there's a problem. +1 LGMT :octopus: !

tcstewar:
Yay!  Merging....

PR #677, opened 2016-02-09
Travis ci js

rsimmons1:
Travis CI can now run pure javascript tests via mocha test runner. As well installation files have been created to make it easier to install testing dependencies. **Requires npm being installed**

Linux:
-$ sudo bash install-test.sh 

OSX:
-$ sudo bash install-test-osx.sh 

Javascript test running for both:
-$ npm test


Seanny123:
Works for me on Linux Mint 17.2

rsimmons1:
When it comes to local testing, all of the testing infrastructure is aimed towards Linux and OSX because of their similarity and simplicity, it actually might work on windows, but has not been tested yet.

nengo_gui/conftest.py
```diff
@@ -2,7 +2,6 @@
 import pytest
 from selenium import webdriver
 
-
```
Seanny123: Why remove this line?

requirements.txt
```diff
@@ -1 +0,0 @@
-nengo
```
Seanny123: Why is this removed?

nengo_gui/static/test/datastore_test.js
```diff
@@ -196,4 +196,4 @@ describe("GrowableDataStore", function() {
         assert.deepEqual(data_store.get_last_data(), [5.1, 5.2, 5.3]);
     });
 
-});
+});
```
Seanny123: Just a nitpick, but FYI, you can avoid pushing not-really-helpful changes like this by reviewing the code in your commit with `git commit -p`. Terry showed me to do that and the quality of my pull requests greatly improved. (:

requirements.txt
rsimmons1: this pull request is quite old, nengo is already installed by setup so this was redundant, in fact i can just remove the requirements file

nengo_gui/static/test/datastore_test.js
rsimmons1: Thanks for the tip Sean! I'll start using that, seems really useful.

install-test.sh
```diff
@@ -0,0 +1,3 @@
+sudo python setup.py develop
```
tcstewar: Why does install-test also do a ```sudo python setup.py develop```?  Does it require nengo_gui to be installed in this way?

tcstewar:
On my machine (Linux Mint 17), I get this when I try to run the tests:

```
tcstewar@ctn13 ~/github/nengo_gui $ npm test

> nengo_gui@0.2.0 test /home/tcstewar/github/nengo_gui
> mocha nengo_gui/static/test

npm ERR! weird error 1
npm WARN This failure might be due to the use of legacy binary "node"
npm WARN For further explanations, please read
/usr/share/doc/nodejs/README.Debian
 
npm ERR! not ok code 0
```

install-test-osx.sh
```diff
@@ -0,0 +1,5 @@
+sudo python setup.py develop
+sudo pip install -r requirements-test.txt
+sudo pip install -U pytest
+sudo easy_install selenium
```
jgosmann: Why isn't this using `pip`?

jgosmann:
I'm not sure whether it is a good idea to provide install scripts that use `sudo` per default. It's probably saver to have the user explicitely say `sudo <install script>`. In general I avoid to install Python packages with `sudo pip` as it installs into system location that I wan't to keep clean and in a defined state. I either do just a `pip install` where I use `pyenv` which provides Python installs in my home directory or `pip install --user` where I use a "system" Python. In both cases `sudo` might actually be harmful because file ownership and permissions might be set wrong.

tcstewar:
> I'm not sure whether it is a good idea to provide install scripts that use sudo per default.  It's probably saver to have the user explicitely say sudo <install script>.

I completely agree.

rsimmons1:
I agree as well, this is a pretty old pull request at this point, testing has gotten a lot better and after reviewing it i do not think its contents are necessary to add to nengo_gui at this moment, and it currently over-complicating the testing infrastructure.

tcstewar:
After discussion, we think this PR needs some rethinking, especially since we're finding that there are very few things on the JS side that can be tested on their own.  So we're removing this PR and will reconsider in the future.

PR #645, opened 2016-02-04
Download Simulation Data as CSV

rsimmons1:
Added download button to the toolbar, when clicked it creates a csv that dumps the values and times from Value or XYValue graphs into a well formatted csv. It then automatically downloads the file as "simulation_data.csv"

PR #678, opened 2016-02-10
Simdata download

rsimmons1:
Adds a button to the toolbar that allows users, to download the data from the graphs as a well formatted csv.

### DIFF
--- a/nengo_gui/static/data_to_csv.js
+++ b/nengo_gui/static/data_to_csv.js
@@ -1,7 +1,10 @@
 /**
- * A function that retuns simulation data as a csv
+ * A function that returns simulation data as a csv, only saves data which is present in graphs.
+ * As well it only saves the data in the datastore, which is based on the 
+ * amount of time kept in the simulation.
  *
- * @param {Nengo.Component.components} data_set A list of the graph items in the simulation
+ * @param {Nengo.Component.components} data_set: A list of the graph items in the simulation
+ * 
  */
 
 
--- a/nengo_gui/static/data_to_csv.js
+++ b/nengo_gui/static/data_to_csv.js
@@ -1,0 +1,65 @@
+/**
+ * A function that retuns simulation data as a csv
+ *
+ * @param {Nengo.Component.components} data_set A list of the graph items in the simulation
+ */
+
+
+var data_to_csv = function(data_set){
+
+  var values = [];
+  var dim_values = [];
+  var times = [];
+  var csv = [];
+  var csv_string = "";
+
+  var data_set = data_set.filter(function(data){
+      return (data.constructor === Nengo.Value || data.constructor == Nengo.XYValue)
+  });
+
+  //Extracts all the values from the data_set variable
+  for(var x = 0; x < data_set.length; x++){
+      values.push([]);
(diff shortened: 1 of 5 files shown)


Seanny123:
Does this replace #645 ?

nengo_gui/static/data_to_csv.js
```diff
@@ -0,0 +1,65 @@
...
+
+var data_to_csv = function(data_set){
+
+  var values = [];
+  var dim_values = [];
+  var times = [];
+  var csv = [];
+  var csv_string = "";
+
+  var data_set = data_set.filter(function(data){
```
Seanny123: Maybe add a comment noting that all graphs saving data over time use both of these as constructors.

nengo_gui/static/data_to_csv.js
```diff
@@ -0,0 +1,65 @@
+/**
+ * A function that retuns simulation data as a csv
```
Seanny123: Typo.

nengo_gui/static/modal.js
```diff
@@ -86,7 +86,34 @@ Nengo.Modal.prototype.footer = function(type, ok_function, cancel_function){
             link.click();
             document.body.removeChild(link);
         });
-    } else if (type === 'refresh') {
+    }else if (type === 'confirm_savecsv') {
```
Seanny123: Spacing. Did you check this code with `jscs`?

Seanny123:
Other than a few nitpicks, it LGTM.

nengo_gui/tests/test_data_to_csv.py
```diff
@@ -0,0 +1,59 @@
...
+import pytest
+from nengo_gui import conftest
+from nengo_gui import testing_tools as tt
+import os
+
+def test_data_to_csv(driver):
+	tt.reset_page(driver)
+
+	tt.update_editor(driver,'''
+import nengo
```
Seanny123: For multi-line strings, I would personally prefer if you did string concatenation like [Brent does here](https://github.com/nengo/nengo_gui/pull/663/files#diff-25dd944edac72587df925012e2e3d6f6R168). It sets off fewer "indentation alarms" when I'm reading code.

tcstewar: Hmm..  except this code requires \n line endings, which would have to be put in manually with Brent's approach, so I'm not sure it makes sense here....  anyone else have thoughts on this?

tbekolay: I definitely agree with @Seanny123 for multiline strings in JS, but in Python using the triple quotes is better, I believe. In addition to the \n line endings, the triple quotes here maintains the indentation in the string itself, which is necessary for Python. In JS, the indentation makes it readable in the source, but the generated HTML doesn't have that indentation. Which is fine, because we never look at the generated HTML, but for Python it makes a functional difference.

rsimmons1:
This PR does replace #645, I'm going to implement all the small changes today. Yeah, the string is formatted this way because it has to exactly match the csv, and this is by far the easiest way to do that, without running into issues as mentioned above.

Seanny123:
LGTM now. Thanks for making those little changes Rees!

PR #683, opened 2016-02-16
Fixed Python3 support

tcstewar:
Just a few small changes to get testing running in Python 3

### DIFF
--- a/nengo_gui/conftest.py
+++ b/nengo_gui/conftest.py
@@ -19,8 +19,8 @@ def driver(request):
         assert driver.title != "Problem loading page"
 
     except AssertionError:
-        print "ERROR: The nengo_gui server is not currently running. \
-        Start the server before running tests."
+        print("ERROR: The nengo_gui server is not currently running. \
+        Start the server before running tests.")
         raise
 
     return driver
--- a/nengo_gui/testing_tools.py
+++ b/nengo_gui/testing_tools.py
@@ -107,6 +107,6 @@ def imgur_screenshot(driver):
     PATH = os.path.join(current_folder, 'test_result.png')
     uploaded_image = im.upload_image(PATH, title="Uploaded to Imgur")
     os.remove('test_result.png')
-    print ""
+    print("")
     print(uploaded_image.title)
     print(uploaded_image.link)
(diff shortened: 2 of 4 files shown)


jgosmann:
All the files containing `print` statements should have a `from __future__ import print_function`.

tcstewar:
> All the files containing print statements should have a from __future__ import print_function.

Hmm...  that would touch a lot of files, complicating merging (although hopefully not too much).  And it just makes the error messages more readable, rather than being something that's needed for python 3 to even run.  So I'm tempted to hold off on enforcing print_function until after 0.2 (when I do the flake8 and jscs touchups).

jgosmann:
It should not be a big problem with merging as only one of the first line in the files is touched. But I would be fine if it was only done for files already touched in this PR for now.

tcstewar:
> It should not be a big problem with merging as only one of the first line in the files is touched. But I would be fine if it was only done for files already touched in this PR for now.

Yay compromise!  :)  I've added those imports now.  And I'll make an issue for this for after 0.2....

Seanny123:
This LGTM.

nengo_gui/conftest.py
```diff
@@ -19,8 +20,8 @@ def fin():
         assert driver.title != "Problem loading page"
 
     except AssertionError:
-        print "ERROR: The nengo_gui server is not currently running. \
-        Start the server before running tests."
+        print("ERROR: The nengo_gui server is not currently running. \
```
tbekolay: What does the backslash do here? Should this be a `\n`? If not, then recommend
```python
print("ERROR: The nengo_gui server is not currently running. "
      "Start the server before running tests.")
```

nengo_gui/testing_tools.py
```diff
@@ -107,6 +108,6 @@ def imgur_screenshot(driver):
     PATH = os.path.join(current_folder, 'test_result.png')
     uploaded_image = im.upload_image(PATH, title="Uploaded to Imgur")
     os.remove('test_result.png')
-    print ""
+    print("")
```
tbekolay: Pretty sure this is equivalent to the simpler `print()`

nengo_gui/conftest.py
tcstewar: Oops...  yup, agreed.  :)

nengo_gui/testing_tools.py
tcstewar: Also a good simplification.  :)

PR #685, opened 2016-02-23
Click on file name to save-as

tcstewar:
Here's a first version of save-as.  The current functionality is tied to clicking on the file name, which I don't particularly like, but seems okay for now until we rethink the top toolbar.

If you try to rename to an existing filename, it will give an error message.  If you try to rename to an invalid file name, it will also give an error message.

### DIFF
--- a/nengo_gui/components/ace_editor.py
+++ b/nengo_gui/components/ace_editor.py
@@ -1,4 +1,5 @@
 import json
+import os
 
 from nengo_gui.components.component import Component
 import nengo_gui.exec_env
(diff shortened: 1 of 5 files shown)


rsimmons1:
LGTM, is it tested on windows and OSX? any ideas for selenium testing?

rsimmons1:
If the name of the file highlighted as the mouse scrolls over the file name it would be much clearer that it is a button.

rsimmons1:
Now tested on linux, windows and OSX works great on all 3, including error messages.

rsimmons1:
Created python test for the save as functionality, its pretty generalized, the test should work even if we have to tweak this functionality.

tcstewar:
> It would be much clearer that it is a button if the name of the file highlighted as the mouse scrolls over it. Like the other buttons in the top bar

Good idea!  That's now added.  :)

PR #695, opened 2016-03-14
Fix GUI due to change in Nengo

tcstewar:
A recent merge in Nengo changed the constructor for nengo.params.Parameter.  This causes the entire GUI to crash horribly.

This change fixes the GUI to work.  It is backwards-compatible so will work with both the recent changes and with the old (release) version.

### DIFF
--- a/nengo_gui/config.py
+++ b/nengo_gui/config.py
@@ -10,15 +10,23 @@ class Config(nengo.Config):
         super(Config, self).__init__()
         for cls in [nengo.Ensemble, nengo.Node]:
             self.configures(cls)
-            self[cls].set_param('pos', nengo.params.Parameter(None))
-            self[cls].set_param('size', nengo.params.Parameter(None))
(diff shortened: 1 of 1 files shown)


tcstewar:
This is rather critical, so I'm just going to merge it without review.  I'll set up a separate Issue to do a more detailed review and see if there was a better way to do it.

tbekolay:
Aha, sorry about that, kind of forgot about the Parameter change somewhere along the way!

nengo_gui/config.py
```diff
@@ -4,21 +4,33 @@
 
 import nengo_gui.components
 
+def make_param(name, default):
+    try:
+        # the most recent way of making Parameter objects
+        p = nengo.params.Parameter(name=name, default=default)
+    except:
```
jgosmann: This should be less general and only catch the actual exception raised for a non existing keyword argument.

tcstewar: Good catch...  I've started #696 to collect this and any other changes people suggest

Issue #694, opened 2016-03-14
TypeError: Unconfigurable is not JSON serializable


ikajic:
Something is weird: when I create a new empty file in terminal (e.g. `touch tmp.py`) and try to open in GUI I get:
```
Traceback (most recent call last):
  File "/home/ivana/phd/workspace/nengo_gui/nengo_gui/swi.py", line 276, in handle_request
    text = getattr(self, command)(*args[1:], **db)
  File "/home/ivana/phd/workspace/nengo_gui/nengo_gui/server.py", line 87, in swi
    components = page.create_javascript()
  File "/home/ivana/phd/workspace/nengo_gui/nengo_gui/page.py", line 314, in create_javascript
    component_js = '\n'.join([c.javascript() for c in self.components])
  File "/home/ivana/phd/workspace/nengo_gui/nengo_gui/components/sim_control.py", line 181, in javascript
    js = self.javascript_config(info)
  File "/home/ivana/phd/workspace/nengo_gui/nengo_gui/components/component.py", line 100, in javascript_config
    return json.dumps(cfg)
  File "/usr/lib/python2.7/json/__init__.py", line 243, in dumps
    return _default_encoder.encode(obj)
  File "/usr/lib/python2.7/json/encoder.py", line 207, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib/python2.7/json/encoder.py", line 270, in iterencode
    return _iterencode(o, 0)
  File "/usr/lib/python2.7/json/encoder.py", line 184, in default
    raise TypeError(repr(o) + " is not JSON serializable")
TypeError: Unconfigurable is not JSON serializable
```

Also, I get the same error when resetting the layout? I am using nengo_gui installed from the master (`python setup.py develop`)

ikajic:
Now I get the same error even when I just try to start the gui from terminal :/

studywolf:
i also get this!

ikajic:
Fixed via #695.

tcstewar:
I just did an emergency merge into master to fix this:  https://github.com/nengo/nengo_gui/pull/695

What happened is that the recent changes to nengo.params.Parameter in nengo master changed the constructor, and so nengo_gui wasn't actually setting the default values any more.

PR #696, opened 2016-03-14
Improved the nengo.params fix

tcstewar:
Pull requires #695 was an emergency fix to repair a broken nengo_gui build ( #694 ) due to a change in nengo.  It got merged immediately, but didn't have a code review about whether it was the right way to fix it.  This is the PR for actually reviewing #695 and doing it better.

@jgosmann has already pointed out that it should be ```except TypeError:``` rather than ```except:```, so that change is here.  Any others?

### DIFF
--- a/nengo_gui/config.py
+++ b/nengo_gui/config.py
@@ -8,7 +8,7 @@ def make_param(name, default):
     try:
         # the most recent way of making Parameter objects
         p = nengo.params.Parameter(name=name, default=default)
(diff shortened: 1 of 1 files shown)


jgosmann:
Apart from that LGTM :cake: 
But I did not actually run the code.

Seanny123:
The rest of the code looks fine to me as well. Haven't used it extensively yet though.

Issue #626, opened 2015-12-09
Console autohide-toggle


studywolf:
So I may have missed this option somewhere, but I would like to be able to have the console pop-up thing permanently on display. When I'm adding code and things become valid and invalid that thing is popping up and down and if i'm adding code to the end of the script the code i'm typing in gets hidden and then viewable and it's kind of a blair-witch scenario. 

So it would be nice to be like "OK i'm going to be doing some code editing for a while and let's just make that console box stay visible (even with no error) so I don't have a seizure."

tcstewar:
+1

Or even just having it pause a few seconds before changing states would be nice.  That way it doesn't flicker in and out of existence all the time....

PR #698, opened 2016-03-18
Added console auto-hide toggle

studywolf:
Addresses #626 

- ctrl-shift-h is the shortcut at the moment
- added to hotkeys list
- pops up to 1/4 height of vmiddle

right now it's default is auto-hide, but I would suggest making auto-hide false as default

### DIFF
--- a/nengo_gui/static/hotkeys.js
+++ b/nengo_gui/static/hotkeys.js
@@ -86,6 +86,19 @@ Nengo.Hotkeys = function () {
                 Nengo.ace.update_trigger = true;
                 ev.preventDefault();
             }
+            // toggle auto-hide of console with TODO: pick a good shortcut
+            if (ctrl && ev.shiftKey && key == 'h') {
+                Nengo.ace.auto_hide_console = !Nengo.ace.auto_hide_console;
+                if (Nengo.ace.auto_hide_console == false) {
+                    Nengo.ace.console_height = document.getElementById('vmiddle').clientHeight / 4.0;
+                    $('#console').height(Nengo.ace.console_height);
+                    Nengo.ace.show_console();
+                }
+                else {
+                    Nengo.ace.hide_console();
+                }
+                ev.preventDefault();
+            }
         }
     });
 }
--- a/nengo_gui/static/ace.js
+++ b/nengo_gui/static/ace.js
@@ -52,6 +52,7 @@ Nengo.Ace = function (uid, args) {
     this.save_disabled = true;
     this.update_trigger = true; // if an update of the model from the code editor is allowed
     this.auto_update = true; // automatically update the model based on the text
+    this.auto_hide_console = true; // automatically hide console if no errors in code
 
     //Setup the button to toggle the code editor
     $('#Toggle_ace').on('click', function(){self.toggle_shown();});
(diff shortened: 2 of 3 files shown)


nengo_gui/static/hotkeys.js
```diff
@@ -86,6 +86,19 @@ Nengo.Hotkeys = function () {
...
+            // toggle auto-hide of console with TODO: pick a good shortcut
+            if (ctrl && ev.shiftKey && key == 'h') {
+                Nengo.ace.auto_hide_console = !Nengo.ace.auto_hide_console;
+                if (Nengo.ace.auto_hide_console == false) {
+                    Nengo.ace.console_height = document.getElementById('vmiddle').clientHeight / 4.0;
+                    $('#console').height(Nengo.ace.console_height);
+                    Nengo.ace.show_console();
+                }
+                else {
+                    Nengo.ace.hide_console();
```
tcstewar: Hmm, the logic here isn't quite right -- turning off auto-hide should only hide the console if there is nothing currently being displayed in it.

nengo_gui/static/hotkeys.js
```diff
@@ -86,6 +86,19 @@ Nengo.Hotkeys = function () {
                 Nengo.ace.update_trigger = true;
                 ev.preventDefault();
             }
+            // toggle auto-hide of console with TODO: pick a good shortcut
+            if (ctrl && ev.shiftKey && key == 'h') {
+                Nengo.ace.auto_hide_console = !Nengo.ace.auto_hide_console;
+                if (Nengo.ace.auto_hide_console == false) {
+                    Nengo.ace.console_height = document.getElementById('vmiddle').clientHeight / 4.0;
```
tcstewar: Why change the default height here?  there's already a default height of 100px...

studywolf: just from usability, i changed it because i found that i was resizing it to be larger every time i displayed it. with the not auto display the height doesn't change to automatically display the whole error message, and so i found there wasn't really any error message where i could read the whole thing at the default height

tcstewar:
I very much think we need something like this.  The flickering error box annoyed me a lot over the last week.

I think there are a few options:

1) toggling the auto-hide (which is what this PR does, although the logic for turning auto-hide back on needs some slight tweaking, as noted inline)

2) directly toggling the console.  That is, get rid of auto-hide completely, and just always show the console, unless someone has turned it off.

3) Getting rid of console hiding completely.  Really, resizing it own to being just a single line at the bottom of the screen is pretty close to hiding, and it means we don't even need a hotkey for it.

I have no idea which of those would be best, but I think I'll toss together a quick PR for getting rid of hiding completely, just to see what that's like.

studywolf:
+1 on getting rid of hiding completely

nengo_gui/static/hotkeys.js
tcstewar: Hmm...  I think I'd be more tempted to have that height saved as a config option, so that it's persistent across runs.  So wherever you've moved the console height to stays (as it's a very screen-dependent operation).

nengo_gui/static/hotkeys.js
studywolf: ahah, you're right this is wildly incorrect!

tcstewar: Yeah, getting the logic of this right is part of why I was thinking it'd be easier to just get rid of hiding altogether...  Take a look at #706 and let me know what you think!

tcstewar:
This has been superseded by #706

Issue #703, opened 2016-03-19
can't resize ensembles


studywolf:
So on master of nengo_gui, when i start up nengo_gui I can't resize ensembles. When i got to the console it says 

``` bash
Uncaught TypeError: Cannot read property 'offsetX' of undefined : netgraph_item.js : 228
```

which is 

``` javascript
 var w = pos[0] - event.clientX + this.ng.offsetX;
```

when i log `this.ng` to the console it's null...

studywolf:
seems to be fixed if you just drop the `this`

studywolf:
looks like it was changed waaay back in 02583f420f21b1691233d0e5c4b2145e7e46d34e by @drasmuss , from `screen_offset`.

tbekolay:
I think the proper fix would be `var self = this; ... self.ng.offsetX`.

But also, this is useful feedback in that since the sprint no one who's on master has tried to resize ensembles until now ;)

studywolf:
yeah i was kind of shocked that no one has resized ensembles in the last
month and a half haha

studywolf:
should all the other `ng. ...` references be changed to `self.ng` too?

tbekolay:
Looks like there are a mix of `ng` and `self.ng` in there yeah... probably either is fine, but best to be consistent. I'd go with `self.ng`, but either way!

studywolf:
alrighty, i'll update the PR

PR #704, opened 2016-03-19
Fix#703 ensemble resizing

studywolf:
Fix for #703. Just dropping the `this` from the call to `ng`.


tcstewar:
Just to check, what was your reasoning for the second commit here?  The first commit seems to be the one that fixes the problem....

tcstewar:
> Just to check, what was your reasoning for the second commit here? The first commit seems to be the one that fixes the problem....

Ah, I see, @tbekolay recommended it for consistency.  Cool!

tbekolay:
We discussed it in #703... it's just for consistency, as there was a mix of `ng` and `self.ng` in there.

tcstewar:
Looks good to me...  thanks for finding this....

PR #705, opened 2016-03-19
Resize only with ctrl

studywolf:
Just added an event handler for items that checks the key pressed, if it's ctrl then it sets 'resizable' to false, and then back to true when the key is let up. The idea is just to make it easier to move around items. I've often had it when an ensemble or whatnot is small and I can't click on it to move without zooming in, i'm only able to resize. So this addresses that!

### DIFF
--- a/nengo_gui/static/components/component.js
+++ b/nengo_gui/static/components/component.js
@@ -117,6 +117,16 @@ Nengo.Component = function(parent, args) {
             self.save_layout();
         });
 
+    $(document).bind('keydown', function(event) {
+        if (event.ctrlKey) {
+            interact(self.div).resizable(false);
+        }
+    });
+    $(document).bind('keyup', function(event) {
+        interact(self.div).resizable(true);
+    });
+
+
     /** Open a WebSocket to the server */
     this.uid = args.uid;
     if (this.uid != undefined) {
--- a/nengo_gui/static/components/netgraph_item.js
+++ b/nengo_gui/static/components/netgraph_item.js
@@ -170,7 +170,6 @@ Nengo.NetGraphItem = function(ng, info, minimap, mini_item) {
     if (!this.minimap) {
         // dragging an item to change its position
         var uid = this.uid;
-        var ng = ng;
         interact(g)
             .draggable({
                 onstart: function () {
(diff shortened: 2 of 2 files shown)


studywolf:
oops, apparently this also has the ng fix in there toooooo...best course of action @tbekolay ?

tbekolay:
You can either rebase this branch to master, or you can just note that this branch is based on #704 and should be merged first.

studywolf:
done! thanks!

tcstewar:
Hmm...  for me (on Chromium), once I touch Ctrl I can no longer resize any of the components.  It's like the keyup never happens at all.  Does that happen to you?

studywolf:
hmm nope it clicks back as soon as i let go, i'm using Google Chrome 49.0.2623.87 on ubuntu 14.04

nengo_gui/static/components/component.js
```diff
@@ -117,6 +117,16 @@ Nengo.Component = function(parent, args) {
             self.save_layout();
         });
 
+    $(document).bind('keydown', function(event) {
+        if (event.ctrlKey) {
+            interact(self.div).resizable(false);
+        }
+    });
+    $(document).bind('keyup', function(event) {
+        interact(self.div).resizable(true);
```
tcstewar: Should this be inside an ```if (event.ctrlKey)``` check?

studywolf: yes it should!

tcstewar:
> hmm nope it clicks back as soon as i let go, i'm using Google Chrome 49.0.2623.87 on ubuntu 14.04

Weird...   It's failing for me on Chrome 44.0.2403.107 as well... And firefox 36.0.4...  (on Mint 17) those are both old versions, but still, everything else works on them....  I'll poke around on other machines...

studywolf:
hmm yeah i'm on my workstation now and getting some weird behaviour from this. i'll double check things on my laptop when i'm back home tonight...

studywolf:
looks like it should be checking the `keycode` on keyup instead of `ctrlKey`, can you try it now @tcstewar?

tcstewar:
> looks like it should be checking the keycode on keyup instead of ctrlKey, can you try it now @tcstewar?

That fixes it for me on Chrome and Firefox  :)

One small thing: could there be a comment added to indicate that keyCode===17 is the ctrl button?

studywolf:
comment added!

tcstewar:
Cool....  and one more thing -- it should probably be added to the hotkeys text.  :)

studywolf:
hmmmm how would you describe this? Temporarily disable resize? Easy move? ...??

tcstewar:
> hmmmm how would you describe this? Temporarily disable resize? Easy move? ...??

Hmm...  I think just "Disable resize             Ctrl"  should be okay....

studywolf:
k!

studywolf:
added!

jgosmann:
Added a commit to use self-documenting code instead of comments and to be consistent with the use of key codes (instead of the `ctrlKey` flag).
Apart from that this LGTM. :cake:

mgserafi:
Tested and confirmed it still works in Chrome. 
LGTM! (just learned what that means)

PR #706, opened 2016-03-21
Remove console hiding

tcstewar:
As an alternate approach for #626 here I've just completely removed console hiding.  

The console is always there, but you can make it smaller (and it doesn't change its size on its own based on message length).  Also, the size of the console is now persistent, so you can shrink it away when you don't want to worry about it, which is almost the same as hiding it (but doesn't require memorizing a hotkey or having some visual indication as to how to get it back).

I'm mostly putting this PR out there for feedback, in comparison to @studywolf 's #698, which provides a hot key to toggle auto-hiding.

I think this PR is simpler and cleaner, but it'd be good to get some other feedback.

### DIFF
--- a/nengo_gui/static/ace.css
+++ b/nengo_gui/static/ace.css
@@ -27,6 +27,7 @@
     -moz-user-select: text;
     -ms-user-select: text;
     user-select: text;
+    display: block;
 }
 
 #console pre {
--- a/nengo_gui/static/ace.js
+++ b/nengo_gui/static/ace.js
@@ -40,14 +40,14 @@ Nengo.Ace = function (uid, args) {
     this.console = document.createElement('div');
     this.console.id = 'console';
     $('#rightpane').append(this.console);
-    this.console_height = 100;
-    this.console_hidden = true;
+    this.console_height = Nengo.config.console_height;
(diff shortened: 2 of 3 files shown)


studywolf:
aw why you gotta throw shade like that hahah

this is great! having the persistent height stay on reload / restart is awesome, i can just max the console height and use my external editor. ++++++++++

tcstewar:
> aw why you gotta throw shade like that hahah

Sorry, didn't mean to say it that way....  yours was addressing the problem with a minimal change to the current system (i.e. normal users wouldn't see any change from before), whereas this one is a somewhat significant user-facing change....  but I think worth it in terms of a) simplifying the code, and b) simplifying the user experience.

tcstewar:
> having the persistent height stay on reload / restart is awesome, i can just max the console height and use my external editor. ++++++++++

Ooo...  hadn't thought of that use case....  nice!

tcstewar:
Okay, after the meeting I think people are in favour of this, but it'd still be good for someone to look at the code...  It's mostly just deleting things, though...  :)

studywolf:
yesssssssssss

PR #707, opened 2016-03-21
Enable sliders on nodes with process outputs

tbekolay:
@celiasmith noticed that you couldn't make sliders when a node had a `Process` as the node output. With this commit, sliders now work properly. I think a few slider bugs might also be fixed with this as I reduced a bit of code duplication, but that's just a guess.

This PR also changes how `node.output` is overridden for all other node output types. The changed version uses processes, so this change breaks compatibility with Nengo 2.0.x, so that should be kept in mind when determining when to merge this commit (and is also why TravisCI fails on this branch).

The new `OverriddenOutput` should be slightly faster in some cases because the (former) `data` list is now a NumPy array where `np.nan` indicates that the value should not be overridden. This array is only modified, not copied, so we save several instances of making a new list of Nones, which may make a marginal difference in some cases.

If I read the old code correctly, data was only ever sent to the client when `callable(node.output)` evaluates to `True`, and it was sent on every timestep. In this PR, I change it so that data is only sent to the client when a value has changed from the previous timestep, and data is now sent to the client for all situations, not just callables. Hopefully I read the old code correctly and intuited why it was only sent then :)

### DIFF
--- a/nengo_gui/components/netgraph.py
+++ b/nengo_gui/components/netgraph.py
@@ -10,7 +10,7 @@ import json
 
 from nengo_gui.components.component import Component
 from nengo_gui.components.value import Value
-from nengo_gui.components.slider import Slider
+from nengo_gui.components.slider import OverriddenOutput
 from nengo_gui.modal_js import infomodal
 import nengo_gui.user_action
 import nengo_gui.layout
@@ -219,7 +219,7 @@ class NetGraph(Component):
                         removed_items.append(item)
                     elif not isinstance(new_obj, old_obj.__class__):
                         rebuilt_objects.append(item)
-                    elif (self.get_extra_info(new_obj) != 
+                    elif (self.get_extra_info(new_obj) !=
                           self.get_extra_info(old_obj)):
                         rebuilt_objects.append(item)
 
@@ -527,7 +527,9 @@ class NetGraph(Component):
         '''
         info = {}
         if isinstance(obj, nengo.Node):
-            if obj.output is None or obj.output is Slider.passthrough_fcn:
+            if obj.output is None or (
+                    isinstance(obj.output, OverriddenOutput)
+                    and obj.output.base_output is None):
                 info['passthrough'] = True
             if callable(obj.output) and hasattr(obj.output, '_nengo_html_'):
                 info['html'] = True
--- a/nengo_gui/components/slider.py
+++ b/nengo_gui/components/slider.py
@@ -2,7 +2,13 @@ import numpy as np
 import struct
 import collections
 
-from nengo.processes import Process
+try:
+    from nengo.processes import Process
+except ImportError:
+
+    class Process(object):
+        pass
+
 from nengo.utils.compat import is_iterable
 
 from nengo_gui.components.component import Component
@@ -83,6 +89,9 @@ class Slider(Component):
         self.from_client = np.zeros(node.size_out, dtype=np.float64) * np.nan
         self.override_output = OverriddenOutput(
             self.base_output, self.to_client, self.from_client)
+        if Process.__module__ == "nengo_gui.components.slider":
+            self.override_output = self.override_output.make_step(
+                shape_in=None, shape_out=node.size_out, dt=None, rng=None)
         self.start_value = np.zeros(node.size_out, dtype=np.float64)
         if not (self.base_output is None or
                 callable(self.base_output) or
--- a/nengo_gui/components/slider.py
+++ b/nengo_gui/components/slider.py
@@ -28,10 +28,8 @@ class OverriddenOutput(Process):
             f = self.passthrough
         elif isinstance(self.base_output, Process):
             f = self.base_output.make_step(shape_in, shape_out, dt, rng)
-        elif callable(self.base_output):
-            f = self.base_output
         else:
-            f = self.static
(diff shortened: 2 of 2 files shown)


tbekolay:
Also, if you want to test this out, the following is a minimal example:

```python
import nengo
from nengo.processes import WhiteSignal

model = nengo.Network()
with model:
    stim = nengo.Node(output=WhiteSignal(10, high=5, rms=0.5), size_out=2)
    a = nengo.Ensemble(n_neurons=50, dimensions=2)
    nengo.Connection(stim, a)
```

tbekolay:
Also, as an FYI this passes the GUI test suite on my local machine (which is using Nengo's `synapse-api` branch).

tcstewar:
Nice, and I like the reorganization.

Would it be possible to include a fallback of the original Slider object if it's running an older version of nengo?  I think it would just mean that on older versions sliders on processes wouldn't work, but that seems okay to me....

tcstewar:
> Would it be possible to include a fallback of the original Slider object if it's running an older version of nengo? I think it would just mean that on older versions sliders on processes wouldn't work, but that seems okay to me....

Ooo, actually, here's an even easier way to be backwards compatible:

```python
try:
    from nengo.processes import Process
    obsolete_processes = False
except ImportError:
    class Process(object):
        pass
    obsolete_processes = True
```

```python
        self.override_output = OverriddenOutput(
            self.base_output, self.to_client, self.from_client)
        if obsolete_processes:
            self.override_output = self.override_output.make_step(
                shape_in=None, shape_out=node.size_out, dt=None, rng=None)
```

tbekolay:
Great idea, I totally didn't think of that! I think you can simplify slightly like so:

```python
try:
    from nengo.processes import Process
except ImportError:
    Process = object
```

```python
        self.override_output = OverriddenOutput(
            self.base_output, self.to_client, self.from_client)
        if Process is object:
            self.override_output = self.override_output.make_step(
                shape_in=None, shape_out=node.size_out, dt=None, rng=None)
```

tbekolay:
`Process = object` doesn't work as `ndarray` is a subclass of `object`. ebcbaaf960e7af7c94e6 has a fix that works for me with Nengo 2.0.3.

tcstewar:
> Process = object doesn't work as ndarray is a subclass of object. ebcbaaf has a fix that works for me with Nengo 2.0.3.

I went through exactly that same approach and ran into the same ndarray problem...  :)  

My feeling is that it's better in this case to be explicit, and so that's why I went with the ```obsolete_process``` flag.  Otherwise the line ```if Process.__module__ == "nengo_gui.components.slider":``` looks really surprising.

tbekolay:
I mean, you could do something like `Process.__module__ == Slider.__module__`, then it's just saying that `Process` is defined here. Personally `obsolete_process` just didn't mean anything to me, and in general i think it's best to keep the definition of something close to where you're actually using it. But yeah, that's personal preference, feel free to change it if you like.

tcstewar:
> Personally obsolete_process just didn't mean anything to me, and in general I think it's best to keep the definition of something close to where you're actually using it. But yeah, that's personal preference, feel free to change it if you like.

I'm actually totally happy with it how it is -- I went with the ```obsolete_process``` thing only because I thought other people would yell at me for being weird.  :)

tcstewar:
It looks like the change to only sending info when the value is different from its previous value is messing up the DataStore history system.  To see this, try this model:

```python
import nengo

model = nengo.Network()
with model:
    stim = nengo.Node(lambda t: 1 if 0.5<t%3<1.0 else 0)
    nengo.Connection(stim, a)
```

Run it for a few seconds (to get a few pulses), then pause the simulation and drag time backwards.  The sliders will be out of sync with the actual data.

I think just removing the ```if (self.to_client is not None and np.any(val[val_idx] != self.value[val_idx])):``` line solves this, but I want to take a closer look at this after the discussion in the dev meeting.

tbekolay:
If you look at [what happens now](https://github.com/nengo/nengo_gui/blob/master/nengo_gui/components/slider.py#L46), data is only ever sent `if callable(self.base_output)`, which happens if `node.output` is callable or `None` (as `None` gets replaced with a passthrough function).

So, to get the same behavior change
```python
np.any(val[val_idx] != self.value[val_idx])
```
to
```python
self.f is OverriddenOutput.static
```

tcstewar:
> ```self.f is OverriddenOutput.static```

Nice!

Although I assume you mean ```self.f is not OverriddenOutput.static```  :)

tcstewar:
One other small bug (that was confusing me while debugging this):

```python
val = np.atleast_1d(self.f(t, *args))
```

should be 

```python
val = np.atleast_1d(self.f(t, *args)).astype(float)
```

(because if ```val``` is all ints, as in the case I posted, it'll round your slider movements to the nearest int)

tcstewar:
Should I go ahead and make those two changes and then merge this?  Other than those the code is looking solid -- can't find any more ways to break it, anyway.  :)

tbekolay:
Just pushed a commit to change those things. I think the int / float issue gets fixed by this because `self.value` is always a `float64`, but worth testing.

tcstewar:
Excellent, that fixes it all for me!  Merging now...  Thank you!

PR #712, opened 2016-03-29
Stop sim.close() from happening during sim.run()

### LINKED ISSUE #699: self._n_steps = self.signals[self.model.step].copy()
When I run this script 
```python
import nengo

model = nengo.Network()
with model:
    def prt(t):
        print t
        return 0
    stim = nengo.Node(prt )
    nengo.Connection(stim, a)
```
i get the following 
```bash
  File "/home/tdewolf/Dropbox/code/nengo_gui/nengo_gui/page.py", line 466, in runner
    self.sim.step()
  File "/home/tdewolf/Dropbox/code/nengo/nengo/simulator.py", line 199, in step
    self._probe()
  File "/home/tdewolf/Dropbox/code/nengo/nengo/simulator.py", line 177, in _probe
    self._probe_step_time()
  File "/home/tdewolf/Dropbox/code/nengo/nengo/simulator.py", line 146, in _probe_step_time
    self._n_steps = self.signals[self.model.step].copy()
TypeError: 'NoneType' object has no attribute '__getitem__'```
when it stops

### PR DESCRIPTION
**Pull request by tcstewar**
This fixes #699 and possibly #711 

At the moment, nengo gets a bit confused if a separate thread ```close```s a Simulator while it is still in the middle of a ```run()``` call.  This just reorganizes nengo_gui slightly to make sure the close gets called after the run.

### DIFF
--- a/nengo_gui/page.py
+++ b/nengo_gui/page.py
@@ -50,6 +50,7 @@ class Page(object):
         self.finished = False  # should this Page be shut down
         self._sim = None       # the current nengo.Simulator
         self.rebuild = False   # should the model be rebuilt
+        self.sims_to_close = []  # list of sims that should be closed
 
         self.code = None       # the source code currently displayed
         self.error = None      # any execute or build error
@@ -109,7 +110,7 @@ class Page(object):
     @sim.setter
     def sim(self, value):
         if hasattr(self._sim, 'close'):
-            self._sim.close()
+            self.sims_to_close.append(self._sim)
         self._sim = value
 
     def get_component(self, component_class):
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: nengo_gui/page.py [hunk 1]
```diff
@@ -50,6 +50,7 @@ def __init__(self, gui, filename, reset_cfg=False):
         self.finished = False  # should this Page be shut down
         self._sim = None       # the current nengo.Simulator
         self.rebuild = False   # should the model be rebuilt
+        self.sims_to_close = []  # list of sims that should be closed
```
tbekolay: Perhaps overkill, but maybe this should be a set? So that if (for some odd reason) a simulator is added twice, it doesn't get closed twice. I think right now closing twice won't error, but that might not be true for all backends.

tcstewar: Even if it's made to be a set, that won't stop a Simulator from maybe being closed twice, if it gets added twice at different times (i.e. if gets added to ```sims_to_close```, then it gets closed, and then it gets added to ```sims_to_close``` again).  Indeed, I'm having trouble coming up with scenarios where it's even possible to get up to a length of 2 on this list... So my feeling is that doing this as a set is overkill.  :)

PR #714, opened 2016-04-04
Fixed error message reporting

tcstewar:
The changed to Nengo Exceptions meant that the logic for finding the line number of the error was going wrong.  The new code explicitly marks the executed code rather than using the default ```<string>```.


tcstewar:
This is a pretty vital fix and people are installing it right now, so I'm merging this now.

PR #718, opened 2016-04-05
Fix dt==0 problem on nengo_spinnaker

tcstewar:
I'm not sure why actual_model_dt can be zero on spinnaker,
but I've seen it happen once so we may as well check for it.

### DIFF
--- a/nengo_gui/components/sim_control.py
+++ b/nengo_gui/components/sim_control.py
@@ -79,8 +79,9 @@ class SimControl(Component):
                 self.rate += (1 - decay) * rate
                 self.skipped = 1
 
-                # compute current proportion of full speed
-                self.rate_proportion = 1.0 - ((self.rate * self.delay_time) /
+                if self.actual_model_dt > 0:
+                    # compute current proportion of full speed
+                    self.rate_proportion = 1.0 - ((self.rate * self.delay_time) /
                                               self.actual_model_dt)
 
         # if we have a desired proportion, use it to control delay_time


mundya:
Is there anything I should be checking up on in `nengo_spinnaker`?

tcstewar:
> Is there anything I should be checking up on in nengo_spinnaker?

I don't think so...  I'm not sure about the conditions where this happened, but it's a reasonable enough thing to check for that I'm not worried about it.  And in general nengo_gui and nengo_spinnaker seem to behaving well together (if I have both of them on up-to-date master).

tcstewar:
> Is there anything I should be checking up on in nengo_spinnaker?

Hmm, I may have spoken too soon...  could you check if nengo_gui input sliders work for you on nengo_spinnaker?  Just run the default.py with a slider and the nengo_spinnaker backend.  For me, it seems to be ignoring the slider input.  However, that could be something wrong with the computer I'm using, or the board -- I'm travelling at the moment (I'm in Groningen) so I can't check with another machine or board....  The SPA sequence example works fine, but it doesn't involve an external input...

mundya:
> could you check if nengo_gui input sliders work for you on nengo_spinnaker?

I can do, afraid I'm away from a board at the moment but I'll check when I get one later today. Are you on master of `nengo_spinnaker`? Might be worth trying commit project-rig/nengo_spinnaker@e8ebb40390f09c941e9fd134c3c42a85ae70aaa6 which changes the port used to communicate node-data with the board.

tcstewar:
I now know what the slider issue was that I was worried was related to this one -- it isn't related.  So I'm merging this quick fix and starting a new issue on the slider thing.

PR #723, opened 2016-04-19
Fixed sliders on nengo_spinnaker

tcstewar:
Current nengo_spinnaker doesn't handle Processes for input Nodes, so that breaks sliders from working.  For now, this fix just calls make_step on the Process and uses that resulting function if it's running on nengo_spinnaker.  In the future, there should be some way of asking a backend if it supports Processes (and we should get Processes working on nengo_spinnaker), but in the meantime this lets things work.


tcstewar:
Since this is breaking most nengo_spinnaker examples, and it's a small fix, I'm merging it now.

PR #725, opened 2016-04-20
Fixed IPythonViz on Python3

tcstewar:
And another tiny little bug fix that allows IPythonViz to run on Python3


tcstewar:
Again, this is a small fix that's blocking for someone, so I'm quickly merging it

PR #736, opened 2016-05-08
Fix #735

### LINKED ISSUE #735: Firing rates plot error
Getting the following error when I try to use the firing rates plot

```
File ".../nengo_gui/components/spike_grid.py", line 49, in gather_data
    y = np.zeros(self.n_pixels, dtype=np.uint8)
TypeError: 'numpy.float64' object cannot be interpreted as an index
```

studywolf:
Casted float to int.

### DIFF
--- a/nengo_gui/components/spike_grid.py
+++ b/nengo_gui/components/spike_grid.py
@@ -46,7 +46,7 @@ class SpikeGrid(Component):
         self.max_value = max(self.max_value, np.max(x))
         if len(x) > self.n_neurons:
             x = x[:self.n_neurons]
-        y = np.zeros(self.n_pixels, dtype=np.uint8)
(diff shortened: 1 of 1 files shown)


tcstewar:
Weird...  must be something that affects some versions of numpy but not others.  We do similar casting elsewhere for a similar reason (e.g. https://github.com/nengo/nengo_gui/blob/master/nengo_gui/components/value.py#L32 ), but I've never seen this particular problem (and it doesn't happen on my computer).

But this fix looks good to me, so I'll merge it (and I'll even see if I can do it correctly -- as in without using the big green "merge pull request" button.... )  :)

tcstewar:
Merged!

tbekolay:
This particular condition used to raise a warning saying that it would eventually raise an error, so it looks like they finally did it!

PR #739, opened 2016-05-11
Set number of neurons in raster plot

tcstewar:
This adds a right-click menu option to set the number of neurons shown in a raster plot.  This value is saved in the ```.cfg```.

Also, instead of only showing the first N neurons, the raster plot shows neurons evenly spaced across the ensemble.  So if you have an ensemble of 11 neurons and only show 3 of them, you will get neurons 0, 5, and 10.

### DIFF
--- a/nengo_gui/static/components/raster.js
+++ b/nengo_gui/static/components/raster.js
@@ -61,6 +61,14 @@ Nengo.Raster.prototype.on_message = function(event) {
     this.schedule_update();
 }
 
+Nengo.Raster.prototype.set_n_neurons = function(n_neurons) {
+    this.n_neurons = n_neurons;
+    this.axes2d.scale_y.domain([0, n_neurons]);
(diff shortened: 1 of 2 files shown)


nengo_gui/static/components/raster.js
```diff
@@ -119,3 +127,57 @@ Nengo.Raster.prototype.reset = function(event) {
...
+                }
+                return valid;
+            }
+        },
+    });
+
+    $('#singleInput').attr('data-error', 'Input should be a positive integer');
+
+    Nengo.modal.show();
+    $('#OK').on('click', function () {
```
Seanny123: No spaces after `function`. Have you run this through `jscs`? Do you want me to do it for you?

tcstewar: None of this code has been run through JSCS...  I've been holding off until the large things are merged.  But you're right I should at least be doing it on these smaller changes.

PR #742, opened 2016-05-11
Set maximum font size for pointer clouds

tcstewar:
This can be handy for stopping the text from getting too large.

### DIFF
--- a/nengo_gui/static/components/pointer.js
+++ b/nengo_gui/static/components/pointer.js
@@ -28,6 +28,8 @@ Nengo.Pointer = function(parent, sim, args) {
 
     this.show_pairs = args.show_pairs;
 
+    this.max_size = args.max_size;
+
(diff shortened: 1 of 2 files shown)


Seanny123:
Code looks good to me. Will try to manually test it later.

Seanny123:
Tried this out and it worked beautifully. It's ready to merge.

nengo_gui/static/components/pointer.js
```diff
@@ -84,6 +86,7 @@ Nengo.Pointer.prototype.generate_menu = function() {
     } else {
         items.push(['Show pairs', function() {self.set_show_pairs(true);}]);
     }
+    items.push(['Set maximum font size..', function() {self.set_max_size_dialog();}]);
```
tbekolay: Should probably be 3 periods at the end of the label

PR #745, opened 2016-05-30
Fixed typo causing autosync to clear

tcstewar:
A variable with the wrong name was causing this option to be cleared whenever the config menu "Cancel" button was clicked.

I believe this is the source of the bug where every now and then someone would find that their network diagram suddenly stopped being updated.


Seanny123:
LGTM.

PR #746, opened 2016-05-30
Side navigation bar

tcstewar:
This reworks the top toolbar, adding a full side navigation bar.

The file open menu is now done with the sidebar, rather than in its own window (and thus no longer disappears if your mouse cursor goes outside it).  Many of the options in the top toolbar have been moved to the side bar, clearing up the top menu a bit.

Eventually, this will also provide the basis for drag-and-drop support, and we may want to merge the config menu into this nav bar (rather than having it as a separate modal popup).

PR #747, opened 2016-06-02
Side navigation bar (cleaned history)

tcstewar:
The commit history in the old branch and PR #746 was broken in a way that I didn't know how to fix.  However, much of the commit history there wasn't all that useful, so I just generated a new branch with the endpoint of that old PR.

### DIFF
--- a/nengo_gui/components/sim_control.py
+++ b/nengo_gui/components/sim_control.py
@@ -181,7 +181,8 @@ class SimControl(Component):
         fn = json.dumps(self.page.filename)
         js = self.javascript_config(info)
         return ('sim = new Nengo.SimControl(control, %s);\n'
-                'toolbar = new Nengo.Toolbar(%s); ' % (js, fn))
+                'toolbar = new Nengo.Toolbar(%s); '
+                'Nengo.sidemenu = new Nengo.SideMenu();' % (js, fn))
 
     def message(self, msg):
         if msg == 'pause':
--- a/nengo_gui/static/components/netgraph.js
+++ b/nengo_gui/static/components/netgraph.js
@@ -16,6 +16,11 @@ Nengo.NetGraph = function(parent, args) {
     this.offsetX = 0;          // global x,y pan offset
     this.offsetY = 0;
 
+    // override the random position chosen by the server for a new item.
+    // This is used by the drag-and-drop system to ensure that the newly
+    // created items appear at the position they were dropped at.
+    this.override_positions = {};
+
     var scale = 1.0;
     Object.defineProperty(this, 'scale', {
         // global scaling factor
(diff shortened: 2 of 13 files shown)


Seanny123:
LGTM. Merge into master tomorrow?

tcstewar:
Yay!  I'll do it tonight -- I need practice with doing the merge correctly.  :)

And it doesn't touch any of the same code as the server refactoring, so hopefully there's no conflicts with that!

tbekolay:
A few notes about the history...

- The history is there so that when you look through the commit log a month or a year down the road, you have some idea what the commit is doing without having to look at the diff -- you are not likely to care about meta stuff about the commit history of a PR that didn't get merged, the fact that this commit is indeed a single commit, etc. I think the commit message for the first commit should likely be the contents of the first message in #746 ("This reworks the top toolbar, adding a full side navigation bar....")
- Even when the history is completely reworked, it's best to keep it in the same pull request because that's where all the conversation about the PR happened. If you were to add a changelog entry about this, would you reference #747 or #746? Both? Really all you need to reference is #746, but then it's confusing that it's closed rather than merged. I will often make a new branch to rework the history... but when I do a diff between that branch and the original branch and see that they're identical, I'll get rid of the old branch and replace it with my new branch (`git branch -m name` renames a branch).
- Aside from the commit messages, the most important part of the commit history is the author information. Rees did a lot of the commits in #746, so it should be reflected in this new branch what portion of the work was done by who. In general, whoever did the most work (which is a judgment call) should be the author of the commit (you can change the author of a commit with `git commit --amend --author="New author <dude@gmail.com>"`). If (through fixup commits or commits that were squashed for other reasons) another author has done significant work on what is now one commit, then in nengo/nengo we've used the convention of adding a "Co-authored by:" note in the commit message (see e.g. https://github.com/nengo/nengo/pull/443/commits/08534384c021243da816f31e45499c8745138a12). If the fixup commits are just minor style fixes or whatever, then it doesn't really matter. You can definitely feel when the short fixup commit you're making becomes something relatively significant; that's usually when you want to use the co-authored by note.

These are the general guidelines that we follow in nengo/nengo and there have been some exceptions, but figured I would put them here so that nengo_gui can follow the same guidelines.

tcstewar:
Good points.  I should definitely put in that full comment from the PR ("This reworks the top toolbar...") and put Rees on as the primary original author.  I'll amend this before doing so.

For the branch name, does that mean I should locally delete side-nav, then locally rename my side-nav2 to side-nav, and then do a force push up to github?  That way it'll all be under #746, and with all the original comments?

tbekolay:
> does that mean I should locally delete side-nav, then locally rename my side-nav2 to side-nav, and then do a force push up to github?

Yep, that's right. Might want to `git diff side-nav side-nav2` before deleting / renaming to make sure it all went smoothly.

tcstewar:
Okay, I followed @tbekolay 's instructions and now this is all back in the old PR #746

PR #751, opened 2016-06-04
Adjust synaptic filter on value plots

tcstewar:
The value plots used to have a fixed filter of 0.01s.  This adds an option to the right-click menu to change the synaptic filter time constant.


tcstewar:
@Seanny123 , if you happen to get a chance to look at this today, that'd be great -- One of the in-class examples Chris wants to do on monday requires this.  It's pretty similar to the #739 PR that you already took a look at.

PR #776, opened 2016-06-14
Added static option to HTML component

arvoelke:
This is useful if you are serving something like an iframe containing javascript, where you don't want it to update every simulation timestep (i.e. served once every time the components are created). Tested manually with the following code:

```
def function(t): return None
function._nengo_html_ = '<iframe src="http://localhost:%d/" />' % self.port
function._nengo_static_ = True
nengo.Node(function, size_in=0, label="")
```

### DIFF
--- a/nengo_gui/components/htmlview.py
+++ b/nengo_gui/components/htmlview.py
@@ -6,13 +6,22 @@ from nengo_gui.components.component import Component
 class HTMLView(Component):
     """Arbitrary HTML display taking input from a Node
 
-    See nengo_gui/examples/basics/html.py for example usage"""
+    See nengo_gui/examples/basics/html.py for example usage.
+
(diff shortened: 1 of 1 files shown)


Seanny123:
Practically, do you have an example of how you would use this usefully? Just might be useful to point people to later.

arvoelke:
I may have something nice to point to in the very near future, but for now just imagine `localhost:port` is some web application that is communicating with the model through another channel (via some thread inside the model).

Seanny123:
LGTM.

tcstewar:
Nice.  I'm fine with ```_html_static_``` for now, but I also wonder whether another option like letting people optionally return a dictionary would work:

 ```python
_html_nengo_ = dict(html='<h1>title</h1>', static=True)
```

Seanny123:
@tcstewar make an issue after the merge?

PR #805, opened 2016-07-11
Refactor frontend

tbekolay:
This PR is similar to #673 in that it refactors a relatively large part of the codebase, but this time it's all frontend stuff. Frontend meaning anything that the browser touches (mostly JavaScript, but also CSS and HTML). This is a work in progress -- I haven't even implemented anything yet, but since this PR will involve introducing some new tools / concepts, and frontend development moves fast and none of us are on the [cutting edge](https://youtu.be/FPFBeyIHsPw?t=11), I'm making the PR now so I can document the process in addition to doing the actual refactoring.

I'll be posting updates frequently throughout the next week or two or however long this takes. If this sounds annoying to you, keep in mind that you can mute individual issues / PRs with this little button on right side of this page:

unsub

That's not the actual button, but that's what it looks like. It's over on the right. Just saying.


tbekolay:
Webpack
-------

Webpack is the first tool that I'm looking into. Everyone that uses it seems to love it, but the documentation is really rough. After looking into this in some detail today, webpack seems well suited for our current situation as we have a ton of assets that get loaded in `page.html`. With webpack, we will have two -- one `css` file and one `js` file. While loading multiple assets is not a huge deal when loading them locally, this will make a hug difference in page load times when accessing a remote server.

Link dump for future me, and for those looking to learn more about webpack:

- [5 minute introduction to webpack](https://egghead.io/lessons/javascript-intro-to-webpack): This is the most clear, practical introduction that I've seen; worth taking a look if you only have 5 minutes to spare.
- [30 minute talk + 10 minutes of questions on how instagram.com works](https://www.youtube.com/watch?v=VkTCL6Nqm6Y): (spoiler: it uses webpack) This is a pretty decent talk that motivates why something like webpack is needed; not many details on how to use it in the talk itself, but it's worth watching for the motivation alone.
- [webpack-howto](https://github.com/petehunt/webpack-howto): A practical how-to companion piece to the instagram.com talk. Definitely one of the better tutorials I came across.
- [Comprehensive webpack intro](http://www.theodo.fr/blog/2016/07/a-comprehensive-introduction-to-webpack-the-module-bundler/): Another intro tutorial that looks good. Haven't read this fully yet.
- [webpack docs](http://webpack.github.io/docs/): For completeness' sake, and so you can see for yourself how confusing the docs are. My favorite part: there's a disqus comment thread on each page, and it's filled to the brim with people complaining about the docs.

Modules
----

The point of webpack is to manage modules, where modules might refer to JS files, CSS, or something that transpiles to JS, CSS, etc. Unfortunately there is not one but three standards for how to define modules: CommonJS, AMD, and ES6 (a.k.a ECMA Script 6, a.k.a. ECMA Script 2015) modules.

The consensus I got early on was that CommonJS is the one to use, even though ES6 will be incorporated in browsers eventually. Unfortunately, TypeScript (which, at the moment, I think we should use) uses ES6 modules, and webpack doesn't support ES6 module natively yet (though it will in webpack 2, [whenever that gets released](https://github.com/webpack/webpack/issues/1545)).

So, at the moment, I'm thinking I'll have to know both CommonJS and ES6 syntax. Hopefully this will be a temporary thing and not the permanent state of affairs, but neither syntax is too complicated, so whatever!

- [ES6 module syntax](https://www.typescriptlang.org/docs/handbook/modules.html): Clear guide; also includes best practices, which is helpful. From the TypeScript documentation.
- [CommonJS module syntax](http://know.cujojs.com/tutorials/modules/authoring-cjs-modules): Another clear guide; includes some performance discussion also.
- [Using ES6 modules with TypeScript and webpack](http://www.jbrantly.com/es6-modules-with-typescript-and-webpack/): Some information about using ES6 modules with webpack.

---

That's all for today; I think I'm getting webpack enough to start actually implementing stuff tomorrow. The biggest decisions that I will have to mull over will be: Where should `webpack.config.js` live? Where should the JS / CSS modules live, should they be moved to simplify the directory structure?

Bonus link for those who feel like raw JS is fine: [ES6 syntax cheatsheet](http://espadrine.github.io/New-In-A-Spec/es6/), which we would get access to through TypeScript. Default function arguments!

jgosmann:
[On the dangers of relying on npm.](http://www.theregister.co.uk/2016/03/23/npm_left_pad_chaos/) ;)

But in all seriousness, it probably makes sense to use it.

tbekolay:
Okay, so the update for this afternoon is that I've resolved all of our dependencies through npm. They all exist in the `node_modules` directory in a form that we can use them. I had hoped that this would be enough to get the GUI in a workable state by symlinking the right files, but the npm versions include `require` statements that work in node.js but not in a browser. On the bright side, now I know what [browserify](http://browserify.org/) is for, and why it's named "browserify"! Of course, webpack is a more general and more powerful tool than browserify, but still an interesting aha! moment.

Aside from that fact that I'll need to do the webpack stuff before I can get the GUI in a workable state again (about half done at this point), there was one other issue that I ran into today: some of the versions of JS packages that we're using aren't installable with npm, though later versions are.

For now, I'm depending on the updated versions that are installable with npm to see if they work; if they don't, then we can put the old versions back in this repo.

Specifically, here are the forced upgrades I've done so far:

- `ace` editor: I can't tell what version we were using (doesn't seem to be in the minified source at all), but now we're using 1.2.2
- `jqueryfiletree`: 1.0.1 -> 2.1.4
- `jquery-ui`: 1.9.2 -> 1.10.4

Note that unlike all the other packages, `jqueryfiletree` is not available through npmjs.com, so I'm using an npm feature that installs packages directly from Github. By default, npm tries to do this through SSH (with a `git://` URL), which doesn't work for me as I haven't uploaded SSH keys to Github. If that's the case for you as well, setting the following git configuration options fixed the issue for me, and forces npm to clone through https instead:

```bash
git config --global url."https://".insteadOf git://
git config --global url."https://github.com/".insteadOf git@github.com:
```

Finally, for interest, here's the `package.json` that I've got going so far. I don't think this will change much (unlike my half-finished `webpack.config.json`):

```json
{
  "name": "nengo",
  "version": "0.2.1",
  "description": "Frontend components for Nengo GUI",
  "author": "Nengo developers",
  "license": "Free for non-commercial use",
  "main": "nengo_gui/www/nengo.js",
  "repository": {
    "type": "git",
    "url": "https://github.com/nengo/nengo_gui.git"
  },
  "bugs": {
    "url": "https://github.com/nengo/nengo_gui/issues"
  },
  "private": true,
  "scripts": {
    "build": "webpack"
  },
  "devDependencies": {
    "ace-builds": "^1.2.2",
    "bootstrap": "^3.3.4",
    "bootstrap-validator": "^0.8.1",
    "d3": "^3.5.3",
    "interact.js": "^1.2.4",
    "jquery": "^2.1.3",
    "jquery-ui": "^1.10.4",
    "jqueryfiletree": "jqueryfiletree/jqueryfiletree#2.1.4",
    "webpack": "^1.13.1"
  }
}
```

Issue #807, opened 2016-07-14
Unable to run Nengo GUI on Mac


Seanny123:
On OS X, Nengo GUI refuses to open a browser window to start the GUI. In Chrome, I get the following error:
<img width="1280" alt="screen shot 2016-07-14 at 12 26 10 pm" src="https://cloud.githubusercontent.com/assets/1918825/16853015/6d09f408-49bf-11e6-8f88-1f725e9e9c8e.png">

In Safari, the error is:
![screen shot 2016-07-14 at 12 26 10 pm 2](https://cloud.githubusercontent.com/assets/1918825/16852979/4222c94a-49bf-11e6-810e-a2dcce82baa3.png)

When I try to change `localhost` to `127.0.0.1`, I get the following error, which may or may not be relevant:
```
Starting nengo server at http://localhost:8080
WARNING:nengo_gui.server:Error response (403): Forbidden
Traceback (most recent call last):
  File "/Users/rgao/Documents/fluffypony/nengo_gui-master/nengo_gui/server.py", line 233, in do_GET
    self.handle_upgrade()
  File "/Users/rgao/Documents/fluffypony/nengo_gui-master/nengo_gui/server.py", line 260, in handle_upgrade
    self.upgrade_to_ws()
  File "/Users/rgao/Documents/fluffypony/nengo_gui-master/nengo_gui/server.py", line 282, in upgrade_to_ws
    raise Forbidden()
Forbidden: Forbidden
```

parenthetical-e:
On my computer, I get a different error that might be related.
```
ERROR:nengo_gui.guibackend:Error during websocket communication.                                                                                               │
Traceback (most recent call last):                                                                                                                             │
  File "/Users/type/anaconda/lib/python2.7/site-packages/nengo_gui/guibackend.py", line 223, in ws_default                                                     │
    msg = self.ws.read_frame()                                                                                                                                 │
  File "/Users/type/anaconda/lib/python2.7/site-packages/nengo_gui/server.py", line 363, in read_frame                                                         │
    self._read()                                                                                                                                               │
  File "/Users/type/anaconda/lib/python2.7/site-packages/nengo_gui/server.py", line 344, in _read                                                              │
    self._buf = self._buf + bytearray(self.socket.recv(512))                                                                                                   │
error: [Errno 54] Connection reset by peer                                                                                                                     │
INFO:root:No connections remaining to the nengo_gui server.                                                                                                    │
[...]

PR #808, opened 2016-07-15
Fix a number of server compatibility issues

jgosmann:
- This allows IP addresses like `127.0.0.1` as websocket origin for servers listening on localhost. Actually the whole `127.0.0.0/8` is allowed because all of these addresses are reserved for loopback. This is meant to allow direct use of the IP if `localhost` does not resolve correctly.
- Adds IPv6 support and allows `[::1]` (and equivalent addresses) accordingly as websocket origin. This is meant to allow to address the server with `localhost` on computers where this resolves to the IPv6 address (though with a sane configuration the IPv4 address should be tried when the IPv6 address fails).
- Safari complains when you send `LF` instead of `CRLF`, so we're sending `CRLF` now.

This is all meant to address issues reported in #807. However, one issue remains: Firefox does not support using IPv4 addresses instead of `localhost` because it complains that accessing `localStorage` is not a secure operation. I won't fix this for now because it is not related to the server refactoring and I don't understand why exactly Firefox complains (Chrome and Safari are fine with it.)

### DIFF
--- a/nengo_gui/guibackend.py
+++ b/nengo_gui/guibackend.py
@@ -102,17 +102,39 @@ class GuiRequestHandler(server.HttpWsRequestHandler):
         '/favicon.ico': 'serve_favicon',
     }
 
-    def get_expected_origins(self):
-        session = self.get_session()
+    def _is_loopback(self, host):
+        if host == 'localhost':
+            return True
+        if host.startswith('127.0.0.'):
+            _, least_significant = host.rsplit('.', 1)
+            try:
+                return int(least_significant) < 256
+            except ValueError:
+                return False
+        return False
+
+    def _is_expected_port(self, port):
+        return port == self.server.server_port
+
+    def _is_expected_host(self, host):
         has_password = self.server.settings.password_hash is not None
-        origins = []
-        if not has_password:
-            origins.append('localhost:' + str(self.server.server_port))
-            if self.server.server_port in [80, 443]:
-                origins.append('localhost')
-        elif session.login_host is not None:
-            return [session.login_host]
-        return origins
+        if has_password:
+            session = self.get_session()
+            return host == session.login_host
+        else:
+            return self._is_loopback(host)
+
+    def is_expected_origin(self, origin):
+        if ':' in origin:
+            host, port = origin.split(':', 1)
+            try:
+                port = int(port)
+            except ValueError:
+                return False
+        else:
+            host, port = origin, self.server.server_port
+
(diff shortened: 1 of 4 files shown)


Seanny123:
This fixes the problem in Chrome, but not Safari, where everything is only half-loaded:
screen shot 2016-07-14 at 5 55 14 pm

jgosmann:
What URL are you using?

On July 14, 2016 8:56:00 PM EDT, Sean Aubin <notifications@github.com> wrote:
>This fixes the problem in Chrome, but not Safari, where everything is
>only half-loaded:
>screen shot 2016-07-14 at 5 55 14 pmsrc="https://cloud.githubusercontent.com/assets/1918825/16860352/35a146e2-49ec-11e6-83ab-c49f7d59f37b.png">
>
>
>---
>You are receiving this because you were mentioned.
>https://github.com/nengo/nengo_gui/pull/808#issuecomment-232833634

-- 

Seanny123:
The `localhost` one

jgosmann:
Safari gives me some CSS errors, but seems to work for me.
Are there any suspicious messages in the developer console?

Seanny123:
I lost access to the computer that was causing the problem, so I can't
follow up on this. Let's ignore that problem for now.

nengo_gui/guibackend.py
```diff
@@ -102,17 +103,47 @@ class GuiRequestHandler(server.HttpWsRequestHandler):
         '/favicon.ico': 'serve_favicon',
     }
 
-    def get_expected_origins(self):
-        session = self.get_session()
+    def _is_loopback(self, host):
```
Seanny123: Could you add a quick docstring about what this function is doing? From what I can parse this is used to verify if the expected host was reached? This function only seems to get called in `assert` statements.

jgosmann: It would say something like "Checks whether `host` refers to the loopback interface." Not sure if that really adds any value because to me it is what the function name already says in more words.

Seanny123: You're right. I wrote that comment when I was sleepy. I shouldn't review code when I'm sleepy.

Seanny123:
LGTM

jgosmann:
This PR has become obsolete.

* The problem of IP addresses as websocket origin has been solved in #915 as it requires a login for any sort of connection even for localhost. Thus, no special case code for that is needed.
* IPv6 support has been added in a better way in #864.
* A fix for the Safari problem is also already merged into master and released as 0.3.1.

PR #809, opened 2016-07-15
Plugin system for nengo_gui

jgosmann:
This adds a plugin system to nengo_gui to allow other Python packages to provide additional functionality. One of the main motivations is to allow to exclude the SPA system from core Nengo while keeping support for the special plots in the GUI.

## Frontend refactoring required before merge

This is based on @tbekolay frontend refactoring which of course is far from completed, but I want to get this out here to receive feedback and be able to confidently continue development on the next generation SPA module. Also, I'm not sure how far @tbekolay refactoring will go, but before we make the plugin system part of the official API, we should ensure that we have well structured frontend JavaScript. After officially integrating the plugin system refactoring of the frontend may get more complicated because we might need to keep backwards compatibility for plugins.

In that refactoring the following points seem of special importance:
* Clear namespacing to have a evident structure where things in the front-end live and can be accessed by plugins, but also to have best-practices how plugins declare and name their namespaces.
* In `NetGraphItem` the menu entries are figured out through a bunch of if-statements. It would be better to use polymorphy. This will change how plugins can define additional menu items.
* Firefox gives me this warning: "mutating the [[Prototype]] of an object will cause your code to run very slowly; instead create the object with the correct initial [[Prototype]] value using Object.create" We should probably fix that warning.

## How does it work?

See [nengo/nengo_gui_plugin_template](https://github.com/nengo/nengo_gui_plugin_template) for a minimal example of a plugin.

Plugins use the "entry points" feature of `setuptools`. A Python package can define entry points in the `setup.py` and another package can discover these with `pkg_resources`. So a nengo_gui plugin defines such entry points. There are two types of entry points:

* `nengo_gui.plugins` provide classes deriving from `nengo_gui.plugin.Plugin` that can provide additional resources (which may be dynamic) accessible via HTTP. In the template it just provides the required JavaScript file.
* `nengo_gui.components` provides additional components/graph.

Overall the required boilerplate should be reduced a bit more I think, but to do so we need the refactoring of the frontend first.

## Feedback

What I want to know is basically: Does this fulfill our requirements? Anything people can think of that should be possible, but is not with this system? Can I base my work for the new SPA module on this (taking into account that some changes will be necessary as the frontend gets refactored)? The SPA module would probably also be a good test bed to see where features are missing or things are not working as intended.


jgosmann:
Things I forgot about so far:
- [ ] backends
- [ ] additional examples

It should be possible to implement both easily with entry points (especially backends).

Issue #813, opened 2016-08-05
Nengo_GUI freezes periodically


arnav-gudibande:
<img width="756" alt="screen shot 2016-08-05 at 11 20 26 am" src="https://cloud.githubusercontent.com/assets/15108659/17446554/b817fc06-5afe-11e6-9e38-eb5b9b1f7ecb.png">

While running my project on macOS 10.12, nengo complains about an unavailable resource and freezes intermittently during runtime. However, when running on Ubuntu 14.04 LTS, nengo runs the project without any issues.

Is there any way I can get nengo to stop complaining about this temporarily unavailable resource?

PS. I have also had this problem on OS X 10.11, so the beta macOS version is not the culprit

[Source Code for Project](https://github.com/arnav-gudibande/nengopacman)

jgosmann:
Thank you for the bug report.

We have rewritten a major part of the underlying server architecture of nengo_gui. This hasn't been part of an official release yet, but you can download the current development version from GitHub. Try to install it and see if that problem persists. If you're lucky it might already be fixed. If not, it would be very helpful to get the error output that you get with that version.

Also, as macOS 10.12 is still in beta, we haven't tested nengo or nengo_gui on it yet.

tcstewar:
I think I've found the problem, and it looks like it's in both the old and the new server architecture.  

I've got what I think is a fix for this, but it'd be great if you could see whether this works for your case.

There are two ways to grab this fix.  One would be to use ```git``` to grab the latest version of nengo_gui from github (```git clone http://github.com/nengo/nengo```, ```cd nengo```, ```python setup.py develop --user```) and then switch to the branch where I've added a potential fix (```git checkout fix-resource-unavail```).

The other option is to directly edit the version of nengo you currently have installed.  To do this, edit the file ```/anaconda/lib/python2.7/site-packages/nengo_gui/swi.py``` and replace line 722

```
            bytes_sent += self.socket.send(data[bytes_sent:])
```

with

```
            try:
                bytes_sent += self.socket.send(data[bytes_sent:])
            except:
                pass
```

Do either of those work for you?

arnav-gudibande:
Solution #2 works, thanks @tcstewar

PR #815, opened 2016-08-10
Fix resource unavailable bug

tcstewar:
At the moment, we don't handle a full socket gracefully.  This mostly comes up if you have a large complex htmlview, which ends up breaking horribly once everything fills up.

I still need to put together a nice test, but here's a fix (this was reported in #813 )

### DIFF
--- a/nengo_gui/server.py
+++ b/nengo_gui/server.py
@@ -585,6 +585,7 @@ class WebSocket(object):
 def _sendall(socket, data):
     bytes_sent = 0
     while bytes_sent < len(data):
+        select.select((), (socket,), ())  # ensure socket is not full
         bytes_sent += socket.send(data[bytes_sent:])
 
 


jgosmann:
I think the code could potentially be simplified by doing always a `select` call first and not catching any exceptions.

As for a good test ... we could open up a websocket connection, fill it up and check that no exception is raised. It's a little bit complicated to implement because to threads have to be coordinated ... not sure if it's worth the hassle.

tcstewar:
> I think the code could potentially be simplified by doing always a select call first and not catching any exceptions.

Oooo...  also a very good idea.  Much much cleaner, and seems to work well.  :)

tcstewar:
Since @jgosmann suggested the entire contents of this one-line fix, I'm going to act as the reviewer for it, even though I made the PR.  Thank you @jgosmann , it looks good to me (and I just checked that it till fixes things when the HtmlNode has lots of content)

PR #829, opened 2016-10-03
Visual Component Interface

rsimmons1:
This PR includes the new interface for drawing components, as well as a basic configure menu for ensembles and the ability to delete components. 

The design is loosely based off google draw in terms of the drag to draw interaction.

Several design components are a temporary implementation including.

- The connection toggle (uses two clicks to create a connection instead of a drag to draw)
- The component configure menu (its super basic, could have a better implementation)
- The delete component option (similar issues to the configure menu)
- The green edge highlight (highlights the screen when a component is toggled, may be confusing to users)
-  The word pass appearing in the generated code (makes code look messy)

However, these are all fully functional.

The rest seems good, really just looking for feedback on the look and feel of the new interface, please mention anything that feels awkward or is missing. Thanks for reviewing!

### DIFF
--- a/nengo_gui/static/ace.js
+++ b/nengo_gui/static/ace.js
@@ -32,7 +32,8 @@ Nengo.Ace = function (uid, args) {
     var code_div = document.createElement('div')
     code_div.id = 'editor'
     $('#rightpane').append(code_div);
-    this.editor = ace.edit('editor')
+    this.editor = ace.edit('editor');
+    this.editor.$blockScrolling = Infinity;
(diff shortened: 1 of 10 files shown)


tcstewar:
Nice!

One bug just from initial playing around: It I make a new object, it will place it inside a sub-Network if the *initial* click is inside that Network, but if the release location is outside the Network then the object will be drawn such that it looks to be outside the Network, when it's really inside it.

rsimmons1:
hmm interesting, I am pretty sure i know what happening there and should be able to fix it.

Issue #763, opened 2016-06-08
GUI stops updating


jgosmann:
with the following exception on the command line:
```
Starting nengo server at http://localhost:8080
ERROR:root:Error processing: u'{"act":"pos","uid":"gate","x":0.09051165512586382,"y":0.8777296829823005}'
Traceback (most recent call last):
  File "/Volumes/Home/blubb/Library/Python/2.7/lib/python/site-packages/nengo_gui-0.2.1.dev0-py2.7.egg/nengo_gui/guibackend.py", line 252, in _handle_ws_msg
    component.message(msg.data)
  File "/Volumes/Home/blubb/Library/Python/2.7/lib/python/site-packages/nengo_gui-0.2.1.dev0-py2.7.egg/nengo_gui/components/netgraph.py", line 413, in message
    act = nengo_gui.user_action.create_action(action, self, **info)
  File "/Volumes/Home/blubb/Library/Python/2.7/lib/python/site-packages/nengo_gui-0.2.1.dev0-py2.7.egg/nengo_gui/user_action.py", line 16, in create_action
    return Pos(net_graph, **kwargs)
  File "/Volumes/Home/blubb/Library/Python/2.7/lib/python/site-packages/nengo_gui-0.2.1.dev0-py2.7.egg/nengo_gui/user_action.py", line 215, in __init__
    self.x, self.y = self.obj_config.pos
TypeError: 'NoneType' object is not iterable
```

(note that this is on the `server-refactor` branch, #673, but it doesn't seem to be related to those changes to me)

jgosmann:
Let me know if anyone manages to reproduce this reliably ... it happened to me quite often trying to build a model, but I'm not managing to reproduce it on purpose.

PLindes:
Futher console output examples

ERROR:root:Error processing: '{"act":"pos","uid":"model.b","x":0.6997759832742082,"y":0.2873048265533181}'
Traceback (most recent call last):
  File "c:\users\plind\documents\github\nengo_gui\nengo_gui\guibackend.py", line 252, in _handle_ws_msg
    component.message(msg.data)
  File "c:\users\plind\documents\github\nengo_gui\nengo_gui\components\netgraph.py", line 413, in message
    act = nengo_gui.user_action.create_action(action, self, **info)
  File "c:\users\plind\documents\github\nengo_gui\nengo_gui\user_action.py", line 16, in create_action
    return Pos(net_graph, **kwargs)
[...]

PR #842, opened 2016-12-27
Handle exceptions during message processing

tcstewar:
Without this patch, any exception caused by a message
will break that component.  If this happens to netgraph,
then the entire system stops working, causing Nengo to
stop updating anything.

This is probably a good bullet-proofing thing to have in nengo gui, although any exception thrown during processing is also an indication of some other bug (so the exceptions still get printed to the screen).  However, for now, this means that things like #763 should stop happening or be less disastrous when they do happen.

### DIFF
--- a/nengo_gui/components/netgraph.py
+++ b/nengo_gui/components/netgraph.py
@@ -410,9 +410,14 @@ class NetGraph(Component):
                 # These should not use the undo stack
                 getattr(self, 'act_' + action)(**info)
             else:
-                act = nengo_gui.user_action.create_action(action, self, **info)
-                self.page.undo_stack.append([act])
-                del self.page.redo_stack[:]
+                try:
+                    act = nengo_gui.user_action.create_action(action,
+                                                              self, **info)
+                    self.page.undo_stack.append([act])
+                    del self.page.redo_stack[:]
+                except:
+                    print('error processing message', repr(msg))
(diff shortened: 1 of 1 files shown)


jgosmann:
Looks good to me, but haven't tested whether it actually helps.

jgosmann:
Not sure if this works. I just had a case where the GUI stopped updating and no exception printed on the console. Saving also did not work.

tcstewar:
> Not sure if this works. I just had a case where the GUI stopped updating and no exception printed on the console. Saving also did not work.

Grrr...   it does help for the cases listed in #763 but I hadn't run into any where there's no exception printed.  I'll search through and see if there are any places where we're swallowing an exception without showing anything....

jgosmann:
Is #763 even caused by a message or happens during the handling of a message?
Though, it's weird that this PR prevents the exception from being printed ...

jgosmann:
Ok, I tested this again, now it seems to work ... not sure what the problem was earlier (maybe I accidentally ran a different nengo_gui version or there is still another problem).

tcstewar:
I've been using this branch for a while now, and it seems to help and doesn't interfere with actually fixing whatever the real problem is.  So I'll merge it, as it's a good thing to have in regardless of whether the other bug gets found.

Issue #785, opened 2016-06-20
Make sure the user knows when Nengo GUI is NOT saving the file properly


ricardodeazambuja:
Several times during the summer school I realized Nengo GUI did not save a particular file after I had clicked the save icon, but I only noticed when I tried to reload the file and the changes were not there.

Normally this occurred when Nengo had an ugly error / warning message on the terminal window.

tcstewar:
Heck yes.  We detect a full shutdown of the server, but it looks like there's still lots of situations where the GUI can be non-responsive without telling anyone.

arvoelke:
I think this also falls under the scope of the issue: it would be nice if a warning prompt ("_You have unsaved changed. Do you really want to close this window?_") is shown if you try to close the tab or navigate away with unsaved changes. Similar to what FB does if you start typing a message / post and then try to close / navigate.

A few people (myself included) made this mistake during the summer school.

Seanny123:
Is this a duplicate of #782? If not, how are these two issues different?

jgosmann:
Also bumping this issue ... just lost a bunch of work again.
I'm also testing out #842 right now and I did not even get anything printed on the console, so with that there is absolutely no indication of malfunction whatsoever.

jgosmann:
@Seanny123 Regarding your question, this issue is more specific than #782. There are other situations than saving were things can stop functioning, e.g. #763. I think it would be good to address these specific issues and then have a general fallback for cases that we missed.

PR #843, opened 2016-12-28
Add an indicator for failed save.

jgosmann:
Addresses #785.

Colors the save icon red until the server confirmed the save. If the save fails the icon will stay red as indicator of the failure. This should improve the current state of affairs, though it's not a perfect solution:

* It would be best if the save would not fail.
* If it fails a descriptive error message would be better because with the red background it might not be clear what it means.
* The button will give a short red flash on a successful save.

Addressing these points is not trivial due to the asynchronous implementation. If we'd rely on the server to notify the client of the failure that notification might never be received. So we have to treat the save as failed until confirmed otherwise.


jgosmann:
This would probably be nice to have in the next release.

PR #853, opened 2017-01-23
Stablise auto-layout over multiple runs

### LINKED ISSUE #150: Inconsistent netgraph item location on runs
For example: on different runs of the spa.py model, the cortical netgraph item will sometimes appear on top of the other items and somtimes below.

astoeckel:
Previously, the auto-layout algorithm would place objects in
different locations with each run. This was caused by edges and
vertices being extracted from normal, unordered Python dictionaries.
Replacing these with OrderedDicts solves this porblem.

Fixes #150


tcstewar:
This looks great to me!  Nice clean fix.  And I love that it means that EnsembleArrays now display with their internal ensembles in creation order!

tcstewar:
This has now been merged.  Thank you!  3d3b4d638322d72aa2ebd263f21e079a9cb833eb

PR #856, opened 2017-01-26
Generate seeds before adding graph components

### LINKED ISSUE #855: Adding plots can change seeds and thus model behaviour
When adding a plot in Nengo GUI (or a slider?), additional objects like nodes can be inserted into the model. That might modify seeds in the model (if only a top-level seed is set) because seeds for the additional nodes might be generated inbetween other seeds for other objects, thus changing all seeds generated afterwards. This is especially a problem when trying to debug the behaviour of a single seed because adding and removing plots to investigate the behaviour can change the behaviour itself.

tcstewar:
This means that if you do set the seed for the top-level
Network in your model, adding or removing graphs will not
affect the seed for the Ensembles in your model.

This should fix #855

### DIFF
--- a/nengo_gui/page.py
+++ b/nengo_gui/page.py
@@ -12,6 +12,7 @@ import nengo
 import nengo_gui
 import nengo_gui.user_action
 import nengo_gui.config
+import nengo_gui.seed_generation
 
 
 class PageSettings(object):
@@ -434,6 +435,12 @@ class Page(object):
         with self.lock:
             self.building = True
 
+            # set all the seeds so that creating components doesn't affect
+            #  the neural model itself
+            seeds = nengo_gui.seed_generation.define_all_seeds(self.model)
+            for obj, s in seeds.items():
+                obj.seed = s
+
             # modify the model for the various Components
             for c in self.components:
                 c.add_nengo_objects(self)
@@ -455,6 +462,10 @@ class Page(object):
             except:
                 line = nengo_gui.exec_env.determine_line_number()
                 self.error = dict(trace=traceback.format_exc(), line=line)
+
+            # set the defined seeds back to None
+            for obj in seeds:
+                obj.seed = None
 
             self.stdout += exec_env.stdout.getvalue()
 
--- a/nengo_gui/seed_generation.py
+++ b/nengo_gui/seed_generation.py
@@ -1,0 +1,32 @@
+import nengo
+import nengo.utils.numpy as npext
+import numpy as np
+
+
+def define_all_seeds(net, seeds=None):
+    if seeds is None:
+        seeds = {}
+
+    if net.seed is None:
+        if net not in seeds:
+            # this only happens at the very top level
+            seeds[net] = np.random.randint(npext.maxint)
+        rng = np.random.RandomState(seed=seeds[net])
+    else:
+        rng = np.random.RandomState(seed=net.seed)
+
+    # let's use the same algorithm as the builder, just to be consistent
+    sorted_types = sorted(net.objects, key=lambda t: t.__name__)
+    for obj_type in sorted_types:
+        for obj in net.objects[obj_type]:
+            # generate a seed for each item, so that manually setting a seed
+            #  for a particular item doesn't change the generated seed for
+            #  other items
+            generated_seed = rng.randint(npext.maxint)
+            if obj.seed is None:
+                seeds[obj] = generated_seed
+
+    for subnet in net.networks:
+        define_all_seeds(subnet, seeds)
+
+    return seeds


tcstewar:
One difficulty with this PR: If you don't set a seed, you now only get a new model whenever the code gets re-executed, rather than whenever you re-build the model.  This is almost the same behaviour as before, except in the case where you:

- don't have a seed on the top-level Network
- press play to run the model
- press reset
- press play to run the model again

Before, this process would give you a new seed each time.  Now, this process gives you the exact same model on the first run as on the second run.

We could fix this by having nengo_gui keep track of all the seeds it has set, and then go through and clear them back to None.  But I'm actually not sure whether or not the new behaviour might be preferred.  It does match better with the idea that "reset" just resets the model, and might perhaps reduce confusion.

jgosmann:
Is this ready for review?

tcstewar:
> Is this ready for review?

Sure!  :)

nengo_gui/page.py
```diff
@@ -238,6 +239,10 @@ def execute(self, code):
                 self.error = dict(trace='must declare a nengo.Network '
                                   'called "model"', line=line)
             model = None
+        else:
+            # set all the seeds so that creating components doesn't affect
+            #  the neural model itself
+            nengo_gui.seed_generation.define_all_seeds(model)
```
jgosmann: I think this is not the right place to do the seed generation:
* It does not take effect for all possible ways to instantiate the GUI. For example, try `IPythonViz` with a top level seed and a single ensemble with a single neuron. For me it produces two different tuning curves depending on whether I insert a value plot or not.
* It happens for every change in the editor which might be expensive.

I think it might be better to do this just before the build happens.

jgosmann:
Let me know if you prefer me to make the requested changes. I just would like to get an opinion on the copy thing before.

nengo_gui/page.py
tcstewar: Ah, very good points.  Yes, it should get moved back to the build, which is where I had it first...  That'll just be removing commit a452c32

tcstewar:
> Since Nengo 2.3.0 we can copy Nengo models, so maybe we should do that before setting the seeds. But maybe we also do not want to bump the Nengo version requirement to 2.3.0? We could also check for the copy support and to the copy only if it is possible.

Interesting...  hadn't thought of doing the copy.  How expensive is it?  The other option would be to go through and set the seeds, but then immediately set them back to None after the build is complete.  That's probably faster than the copy, but more unclear...

And yes, we don't want to bump the nengo version requirement, so we'd have to check whether we can copy.

tcstewar:
> Interesting... hadn't thought of doing the copy. How expensive is it?

Hmm, looks like the copy is broken for lots of nengo model, including anything with SPA.  Example:

```python
import nengo.spa as spa

model = spa.SPA()
with model:
    model.a = spa.Buffer(16)
m2 = model.copy()
```

```
AttributeError                            Traceback (most recent call last)
<ipython-input-8-4c0bfc568332> in <module>()
      4 with model:
      5     model.a = spa.Buffer(16)
----> 6 m2 = model.copy()

/home/tcstewar/github/nengo/nengo/network.pyc in copy(self, add_to_container)
    227             # However, we deal with it here, so we ignore the warning.
    228             warnings.simplefilter('ignore', category=NotAddedToNetworkWarning)
--> 229             c = deepcopy(self)
    230         if add_to_container is None:
    231             add_to_container = len(Network.context) > 0

/usr/lib/python2.7/copy.pyc in deepcopy(x, memo, _nil)
    188                             raise Error(
    189                                 "un(deep)copyable object of type %s" % cls)
--> 190                 y = _reconstruct(x, rv, 1, memo)
    191 
    192     memo[d] = y

/usr/lib/python2.7/copy.pyc in _reconstruct(x, info, deep, memo)
    334             state = deepcopy(state, memo)
    335         if hasattr(y, '__setstate__'):
--> 336             y.__setstate__(state)
    337         else:
    338             if isinstance(state, tuple) and len(state) == 2:

/home/tcstewar/github/nengo/nengo/network.pyc in __setstate__(self, state)
    206     def __setstate__(self, state):
    207         for k, v in iteritems(state):
--> 208             setattr(self, k, v)
    209         if len(Network.context) > 0:
    210             warnings.warn(NotAddedToNetworkWarning(self))

/home/tcstewar/github/nengo/nengo/spa/spa.pyc in __setattr__(self, key, value)
    108             if value.label is None:
    109                 value.label = key
--> 110             self._modules[key] = value
    111             for k, (obj, v) in iteritems(value.inputs):
    112                 if type(v) == int:

AttributeError: 'SPA' object has no attribute '_modules'
```

I'll post an issue in Nengo for that, but in the meantime, I think resetting the set seeds back to None is the way to go.

jgosmann:
> hadn't thought of doing the copy. How expensive is it?

Should be cheap in most cases as there is not a lot of data that needs to be duplicated as long as no parameters are set to huge Numpy arrays.

> but then immediately set them back to None after the build is complete

That will require us to store which seeds where set by the user and we have to iterate over the whole model twice. It might be faster then copying, but I would not be surprised if it is slower in many cases either. The main advantage might be that it works for all Nengo versions?

tcstewar:
> The main advantage might be that it works for all Nengo versions?

That, and that copy seems to be broken for some nengo models right now.  I assume that'll get fixed, but in the meantime....  :)

tcstewar:
> Setting the seeds also modifies the model. This is probably fine for the in-GUI editing, but maybe not within an IPython notebook where the user can further modify the model after it has been displayed.

I've changed this so that the seeds get set back to None after building the model (and I also followed @jgosmann 's suggestion to do this in build, rather than where I was doing it).  I think it's ready for a re-review.  :)

jgosmann:
Right now the seed modification is happening in the build thread which leaves me with two questions:

1. Could that thread be interrupted and thus leave the the seeds in the modified state? I suppose with the normal GUI we run the code each time in a fresh environment, but this is not the case in Jupyter notebooks. But I'm not sure if a `with` or `try/finally` block construct would help because if the thread was killed, the `finally` condition might not execute either.
2. Could other threads access the model while the `seeds` are in a modified state? Again for the normal GUI this is probably not a problem, but in Jupyter notebook one could potentially execute further code during the build. Note that it would not help to move the code to the main thread because the server main thread is also just another non-main thread in the Jupyter notebook.

So maybe we have to do the copy after all to handle this correctly in all cases? nengo/nengo#1271 fixes copying for SPA models btw.

tcstewar:
> Could that thread be interrupted and thus leave the the seeds in the modified state? I suppose with the normal GUI we run the code each time in a fresh environment, but this is not the case in Jupyter notebooks. But I'm not sure if a with or try/finally block construct would help because if the thread was killed, the finally condition might not execute either.

It shouldn't be possible to interrupt it without doing something really weird.  I could see putting it (and the Connection/Node removal) in a finally block.

> Could other threads access the model while the seeds are in a modified state? Again for the normal GUI this is probably not a problem, but in Jupyter notebook one could potentially execute further code during the build. Note that it would not help to move the code to the main thread because the server main thread is also just another non-main thread in the Jupyter notebook.

That is possible, but it's also already a problem since all the Nodes and Connections get added to the Network during that time as well.  So I think dealing with that is a separate PR.

PR #860, opened 2017-02-07
Logging instead of print

### LINKED ISSUE #290: Use `logging` instead of print statements
We have more control over logging statements than print statements. Also fewer Python 3 issues!

astoeckel:
This pull request fixes #290. All modules fetch a module specific ``logger`` object which is then used to log messages previously printed with ``print``, or ``logging.`` without a ``logger`` object.

This patchset should not significantly change the output of nengo_gui visible to the user. A future improvement would be a nice formater, which may print the log severity and other information.  Note that all messages are now written to ``stderr`` instead of ``stdout``.

Note that this patch set touches various places all over the code, it should thus be reviewed thoroughly.

Things which need to be checked before merging this pull request:
- [ ] Were the log severities chosen correctly?
- [ ] Does lowering the default log level have any unwanted side-effects?
- [ ] Can the calls to the ``warnings``-module in ``ipython.py`` be safely replaced by calls to ``logging``? My guess was no, so I didn't change anything here.

### DIFF
--- a/nengo_gui/gui.py
+++ b/nengo_gui/gui.py
@@ -2,8 +2,6 @@
 backend."""
 
 import logging
-from __future__ import print_function
-
 import select
 import signal
 import sys
--- a/nengo_gui/gui.py
+++ b/nengo_gui/gui.py
@@ -1,6 +1,7 @@
 """Classes to instantiate and manage the life cycle of the nengo_gui
 backend."""
 
+import logging
 from __future__ import print_function
 
 import select
@@ -11,6 +12,9 @@ import webbrowser
 
 import nengo_gui
 from nengo_gui.guibackend import GuiServer
+
+
+logger = logging.getLogger(__name__)
 
 
 class ServerShutdown(Exception):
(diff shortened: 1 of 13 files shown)


nengo_gui/gui.py
```diff
@@ -100,9 +105,9 @@ def _confirm_shutdown(self, signum, frame):
             if line[0].lower() == 'y':
                 raise ServerShutdown()
             else:
-                print("Resuming...")
+                logger.info("Resuming...")
         else:
-            print("No confirmation received. Resuming...")
```
jgosmann: These last two `print` statements should either stay `print`s or should be converted to `sys.stdout.write` because this output is part of direct user interaction and should never be redirected to a log file for example.

jgosmann: Actually, I might argue that the other `print`s in this function should stay `print`s too.  Though it is not as clear cut for those.

nengo_gui/gui.py
```diff
@@ -70,25 +74,26 @@ class InteractiveGUI(BaseGUI):
...
         finally:
-            print("Shutting down server...")
+            logger.info("Shutting down server...")
 
             self.server.wait_for_shutdown(0.05)
             n_zombie = sum(thread.is_alive()
                            for thread, _ in self.server.requests)
             if n_zombie > 0:
-                print("%d zombie threads will close abruptly" % n_zombie)
+                logger.warning("%d zombie threads will close abruptly",
```
jgosmann: I'd probably make this a `debug` message. It isn't really relevant to the user.

nengo_gui/main.py
```diff
@@ -10,9 +10,12 @@
 from nengo_gui.password import gensalt, hashpw, prompt_pw
 
 
+logger = logging.getLogger(__name__)
+
+
 def old_main():
-    print("'nengo_gui' has been renamed to 'nengo'.")
-    print("Please run 'nengo' in the future to avoid this message!\n")
```
jgosmann: This should always be visible to the user and not end up in some log file. Thus, it should stay a `print` or even better go to `sys.stderr`.

nengo_gui/password.py
```diff
@@ -27,4 +31,4 @@ def prompt_pw():
         p1 = getpass("Enter password: ")
         if p0 == p1:
             return p0
-        print("Passwords do not match. Please try again.")
```
jgosmann: This is also code with direct user interaction via stdin and stdout, thus it should stay a `print`.

nengo_gui/guibackend.py
```diff
@@ -295,7 +295,7 @@ def get_session(self):
         return session
 
     def log_message(self, format, *args):
-        logger.info(format, *args)
+        logger.debug(format, *args)
```
jgosmann: What are “these” log messages? It seems that this function isn't called in normal operation. Maybe in some specific error cases? But if it is error cases a higher log level might be better?

astoeckel: This function is called from the web server base class and is called once for each incoming request.

jgosmann: You're right, for some reason those messages did not appear on my Macbook (potentially I was running the wrong version of Nengo GUI). I agree that the `debug` log level makes more sense.

jgosmann:
Added a commit to change things back to prints where appropriate. LGTM now. 🍰

PR #863, opened 2017-02-07
Fix password protection in Python 3.

jgosmann:


tcstewar:
Hmm, on Windows (Python 3.4 and 3.5), I get this error:

```
Starting nengo server at https://localhost:8080
  File "c:\users\terry\documents\github\nengo_gui\nengo_gui\server.py", line 235, in do_GET
    self.http_GET()
  File "c:\users\terry\documents\github\nengo_gui\nengo_gui\server.py", line 252, in http_GET
    response.send(self)
  File "c:\users\terry\documents\github\nengo_gui\nengo_gui\server.py", line 94, in send
    request.wfile.write(self.data)
  File "C:\Python35-64\lib\socket.py", line 593, in write
    return self._sock.send(b)
  File "C:\Python35-64\lib\ssl.py", line 861, in send
    return self._sslobj.write(data)
  File "C:\Python35-64\lib\ssl.py", line 586, in write
    return self._sslobj.write(data)
ssl.SSLEOFError: EOF occurred in violation of protocol (_ssl.c:1844)
```

Any ideas?  I'll keep poking at it and see if anything makes sense to me...

jgosmann:
weird, is there a way for you to print a hexadecimal representation of `self.data` in line 94 of `server.py`?

tcstewar:
> weird, is there a way for you to print a hexadecimal representation of self.data in line 94 of server.py?

....  okay, you're not going to believe this, but here's the problematic one:

```python
b''
```

Apparently SSL doesn't support sending empty messages?

However, if I try this approach to avoid it, I get a new problem:

```python
 if len(self.data) > 0:
     request.wfile.write(self.data)
```

Now the netgraph doesn't appear and the GUI is in a broken state....  until I try opening the Inspect console in the browser, at which point everything magically fixes itself and it all works fine.  ?!????!???

jgosmann:
Maybe it's not the empty message, but whatever has been sent before that?

tcstewar:
> Now the netgraph doesn't appear and the GUI is in a broken state.... until I try opening the Inspect console in the browser, at which point everything magically fixes itself and it all works fine. ?!????!???

Never mind, this is a separate bug that also exists in Python27 on master on my computer (but only when running under SSL).  I'll see if this happens on other machines, or is specific to my (fairly old) version of Chrome (it doesn't happen under Firefox).

tcstewar:
For this PR, I think the ```if len(self.data) > 0:``` fix is all we need.  Although I do notice that it's never an empty string if I'm not in HTTPS mode, so I'm not sure what's happening differently here.

jgosmann:
Can you figure out with what sort of requests that happens and which headers are being sent before the empty data block?

jgosmann:
Hey, can we get this PR merged?
The problems you are reporting, seem to be not caused by the changes in this PR and probably exist without too. But I'd prefer to look at those separately and this PR is at least resolving some problems.

jgosmann:
And it would be nice if PRs for Nengo GUI did not lie around for forever. I just noticed that this is another problem that I have now fixed twice apparently (other fix 08c07cb8312560c782764199e583219e17d4d75d and b36f31dcf0e1cd9e6bb442be3613fa0fd03dcdc5 in #915).

tcstewar:
> And it would be nice if PRs for Nengo GUI did not lie around for forever

Yes -- and that's my fault completely.  I'm going at this right now and I'm setting up a scheduled time for me to work on nengo GUI PRs every week.

tcstewar:
THis works nicely on Linux, and it does mean things run on Windows, but it looks like the SSL/password stuff has been broken on windows (for py2 and py3) for completely separate reasons for some time....  I'll merge this PR and set up a separate issue to deal with the SSL/password stuff on windows.

PR #865, opened 2017-02-07
Fix port selection.

jgosmann:
The `nengo` command ignored the `--port` argument. This is fixed here. Also, if no port is provided and 8080 is not available, it will increase the port number until an available port is found. This is consistent with the Jupyter notebook behaviour and usually pretty convenient. It makes it also less important to be able to pass arguments to nengo_gui through pytry as it allows to start multiple GUI instances with pytry.

### DIFF
--- a/nengo_gui/main.py
+++ b/nengo_gui/main.py
@@ -1,4 +1,5 @@
 import argparse
+import errno
 import logging
 import os.path
 import threading
@@ -29,7 +30,7 @@ def main():
         '--key', nargs=1, default=[None], type=str, help="SSL key file")
     parser.add_argument(
         '-P', '--port', dest='port', metavar='PORT',
-        default=8080, type=int, help='port to run server on')
+        type=int, help='port to run server on')
     parser.add_argument(
         'filename', nargs='?', type=str, help='initial file to load')
     parser.add_argument(
@@ -39,7 +40,8 @@ def main():
         default='nengo', type=str, help='default backend to use')
     parser.add_argument('--browser', dest='browser', action='store_true')
     parser.add_argument('--no-browser', dest='browser', action='store_false')
-    parser.add_argument('--auto-shutdown', nargs=1, type=float,
+    parser.add_argument(
+        '--auto-shutdown', nargs=1, type=float,
         help="Time limit before automatic shutdown. Set to 0 to deactivate.",
         default=[2])
     parser.set_defaults(browser=True)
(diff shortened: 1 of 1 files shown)


tcstewar:
This looks great to me!  

One thing that that I noted while testing is that on Windows, starting up a new server on a port that's already in use doesn't throw an exception, so it never does the bit of code to increase the port number.  I'll raise that as a separate issue.

jgosmann:
> One thing that that I noted while testing is that on Windows, starting up a new server on a port that's already in use doesn't throw an exception, so it never does the bit of code to increase the port number. I'll raise that as a separate issue.

Windows has to break everything once again ... but does it really allow you to open two servers listening on the same port? Which server gets the requests then? There should be some sort of error. 😖

tcstewar:
> Windows has to break everything once again ... but does it really allow you to open two servers listening on the same port? Which server gets the requests then? There should be some sort of error. 😖

The new one just silently sits there, and all the connections go to the old one.  :(  I'm not seeing any place where we hide the error....

jgosmann:
Can you check what happens if you start multiple Jupyter notebook instances?

tcstewar:
> Can you check what happens if you start multiple Jupyter notebook instances?

That seems to work fine.  The first one starts on 8888 and then the second one gives the ```The port 8888 is already in use, trying another random port.``` message and starts on 8889.

jgosmann:
The Jupyter code is not extremily different:
```python
        for port in random_ports(self.port, self.port_retries+1):
            try:
                self.http_server.listen(port, self.ip)
            except socket.error as e:
                if e.errno == errno.EADDRINUSE:
                    self.log.info('The port %i is already in use, trying another port.' % port)
                    continue
                elif e.errno in (errno.EACCES, getattr(errno, 'WSAEACCES', errno.EACCES)):
                    self.log.warning("Permission to listen on port %i denied" % port)
                    continue
                else:
                    raise
            else:
                self.port = port
                success = True
                break
```

tbekolay:
I think the difference is the error being caught; [socket.error](https://docs.python.org/2/library/socket.html#socket.error) is an IOError, which isn't a subclass of OSError. The base class is [EnvironmentError](https://docs.python.org/2/library/exceptions.html#exceptions.EnvironmentError) if you want to catch things other than socket errors.

jgosmann:
But other exceptions are not caught at all, so Nengo should exit with an exception if we're catching the wrong exception, but apparently it is not.

tbekolay:
Perhaps `self.http_server.listen` raises the exception then?

tcstewar:
> I think the difference is the error being caught; socket.error is an IOError, which isn't a subclass of OSError. The base class is EnvironmentError if you want to catch things other than socket errors.

This turns out to be needed on Linux, actually...  I get a socket.error instead of an OSError.  So I need this as the logic:

```python
    except EnvironmentError as err:
        if args.port is None and err.errno == errno.EADDRINUSE:
            ...
```

I think that's safe, as EnvironmentError has an errno attribute, so it should always be there....

However, that doesn't affect things on windows.  :(

tcstewar:
I'll switch the OSError to an EnvironmentError and merge this, and then start up a separate issue for tracking down why windows is being odd....

jgosmann:
Yeah, this PR isn't a regression in functionality on Windows anyways, is it?

tcstewar:
> Yeah, this PR isn't a regression in functionality on Windows anyways, is it?

Exactly -- this has always been broken on windows -- I just never noticed.  :)

jgosmann:
Though one thing after looking at the Jupyter code: They chose five incremental ports and then five random ports (around the initial point) and fail then, we keep incrementing. I wonder if that might be a problem in some situations ...

tcstewar:
> Though one thing after looking at the Jupyter code: They chose five incremental ports and then five random ports (around the initial point) and fail then, we keep incrementing. I wonder if that might be a problem in some situations ...

Hmm...  I'm fine with the keeping on incrementing approach... although I could maybe see cutting it off sooner.  I'm happy to leave it for a separate PR, though (if it does end up being a problem).

PR #866, opened 2017-02-08
Added link to forum

tcstewar:
I think a link in the top toolbar to the forum would be helpful.  I thought about also adding an API link, but I think getting people to the forum is more directly useful.

### DIFF
--- a/nengo_gui/templates/page.html
+++ b/nengo_gui/templates/page.html
@@ -94,6 +94,9 @@
                 <li id='Redo_last_button' role="presentation">
                     <a title='Redo last' class='glyphicon glyphicon-share-alt'> </a>
                 </li>
+                <li id='Forum_link' role="presentation" onclick="window.open('https://forum.nengo.ai');">
+                    Help
+                </li>
                 <li id='Help_button' role="presentation">
                     <a title='List of hotkeys' class='glyphicon glyphicon-question-sign'> </a>
                 </li>
--- a/nengo_gui/static/top_toolbar.css
+++ b/nengo_gui/static/top_toolbar.css
@@ -51,12 +51,17 @@
     cursor: default !important;
 }
 
-#filename:hover {
+#Forum_link {
+    padding: 0.5em;
+    cursor: hand !important;
+}
+
+#filename:hover, #Forum_link:hover {
     color: #eee;
     background: #666;
 }
 
-#Help_button{
+#Help_button, #Forum_link{
     float: right;
 }
 


tcstewar:
If anyone could take a quick glance at the look of this forum link I added to the top toolbar, that'd be handy.  I'd like to squeeze it into the next release (which should be soon!)

nengo_gui/templates/page.html
```diff
@@ -94,6 +94,9 @@
                 <li id='Redo_last_button' role="presentation">
                     <a title='Redo last' class='glyphicon glyphicon-share-alt'> </a>
                 </li>
+                <li id='Forum_link' role="presentation" onclick="window.open('http://forum.nengo.ai');">
```
tbekolay: The link should be https not http (it'll redirect automatically, but might as well put in the right URL)

nengo_gui/templates/page.html
```diff
@@ -94,6 +94,9 @@
...
+                    forum
```
tbekolay: It feels weird to have this lowercase plaintext right beside a bunch of icons... (I'll post a screenshot shortly). Also, the word "forum" isn't necessarily meaningful to our users... why would they go to the forum? Feels like we're giving them an answer to a question they don't have yet.

A few alternatives... not sure what my preference is:

1. Change the hotkey icon to [something else](http://getbootstrap.com/components/) (mabye glyphicon-info-sign? there's unfortunately no keyboard icon, though we could use a different iconset) and make the forum like the question mark icon.
2. Add the forum link to the Hotkeys modal instead, and maybe change the title to something like "Help" since that's what the icon suggests.
3. Change the text for "Need help?" or "Get help" or just "Help"

nengo_gui/templates/page.html
tcstewar: Oops, good point...  that's now fixed.

tcstewar:
> Change the text for "Need help?" or "Get help" or just "Help"

Ooo...  I like that.  Much clearer.  I'd prefer to have it as text, and not buried inside the hotkey menu, just to make it as blatant as possible.  But you're definitely right that "forum" is way too obscure.  What do you think of just "Help"?

image

tbekolay:
> What do you think of just "Help"?

Sounds good to me!

jgosmann:
The question mark icon looks to me also like a “help” button. Could that be confusing? Do we have a keyboard icon (as it shows the hot keys)?

tcstewar:
> The question mark icon looks to me also like a “help” button. Could that be confusing? Do we have a keyboard icon (as it shows the hot keys)?

I haven't found another icon that makes sense for the hot-key link, but the more I think about it, the more I like having both the question mark and the "Help" text there.  It means that there are two different things to draw people's attention to that top-right corner, and if one of them doesn't get them what they want, then they'll maybe click on the other one.

So I'm happy with this for now, but it's definitely up for changing in the future.  :)

Issue #731, opened 2016-05-07
Improve gui experience for large models (like Spaun)


tcstewar:
Right now, it's pretty awkward to work with large models.  This issue is just meant to brainstorm some possibilities of what we can do to improve things.  Some initial ideas:

1) Multiple views (being able to save a GUI layout and quickly swap between layouts)

2) Adding graphs without rebuilding the model (this would require some support from nengo to add/remove things to a live model)

3) Speed improvements with large models (the interface can get slow and jerky)

4) Running just a subnetwork of the model with fixed inputs (or with inputs copied from a previous run)

Other ideas?

jgosmann:
> 2) Adding graphs without rebuilding the model (this would require some support from nengo to add/remove things to a live model)

That could be related to having an incremental build process (nengo/nengo#869).

5) Panning the whole view when it's completely covered with a network (instead of moving that network). Maybe with a hot key, that would also make moving around in other circumstances easier as you don't need to find a visible spot of background to click on.

Issue #285, opened 2015-06-03
Ability to pan while zoomed into a subnetwork


bjkomer:
Currently if you zoom into a subnetwork in such a way that it fills up your whole screen, there is no way to pan to look around in that subnetwork. If you click and drag, you are actually moving the subnetwork around rather than moving your view. This could potentially mess up the network arrangement when you really just want to pan.

One option could be to allow the arrow keys to be used to pan, as well as clicking and dragging on the mini-map.

studywolf:
ooh or holding alt and click-hold to drag the background, i think that's a standard thing...

celiasmith:
do we want to worry about a touch screen solution?

Travis DeWolf wrote:

bjkomer:
Hmm, it does get more complicated when you throw touchscreens into the mix.

One option could be to have a 'draggable' section of the component. This could be a corner, or an area at the top (similar to how you can drag browser windows around by clicking the top, but can't drag them around clicking anywhere else). When the network is collapsed, the whole thing becomes draggable, but when it is expanded this draggable region needs to be somewhere else (like the top). I have a feeling that this could get confusing to some users..
Another option could be to have some other touch combination allow it to pan all the time. Something like putting one finger stationary on the screen, and then dragging with the other.

As for on a desktop, I like the alt-click idea. If you alt-click on a browser window right now, you can drag the window around (didn't learn that until just now!), so this would override that for people. I have a feeling not many people would miss it though.

tcstewar:
> do we want to worry about a touch screen solution?

At the moment, touch screen are horribly broken anyway, so I wouldn't worry.

jgosmann:
On touch devices you could still use the minimap.

krisfrenette:
This was fixed as part of #881.

PR #881, opened 2017-04-03
Always pan view when dragging with CTRL or MMB pressed

astoeckel:
Always pan the entire view when dragging the mouse while holding the control key or whenever the middle mouse button is pressed.

**Rationale:**
When zoomed into several layers of nested networks, it is currently impossible to navigate the visualization without zooming out and then zooming in again at another point, since dragging moves the component/subnetwork the mouse is over, and not the viewport.

Other applications solve this by panning the view if a modifier key is pressed. For example, in Inkscape this is solved by letting users pan the view when the middle mouse button (MMB) is pressed. In Blender (a prime example for usability), one can pan the view when pressing CTRL. I propose allowing both, since Apple stuff does not feature a middle mouse button to my knowledge.

**Issues with this PR:**
The entire patch is rather hackish. It actually intercepts the ondragmove event of the components and redirects the event to the Netgraph component if either CTRL or MMB are pressed (``event.button`` is useless in the MouseMoveEvent.) Furthermore, a small patch to intercept.js is required. We may have to file an upstream bug report.

**Todo:**
- [x] Test with Firefox
- [x] Test with Chromium
- [ ] Test with Edge
- [ ] Test with Safari

### DIFF
--- a/nengo_gui/static/components/netgraph.js
+++ b/nengo_gui/static/components/netgraph.js
@@ -683,3 +683,19 @@ Nengo.NetGraph.prototype.scaleMiniMapViewBox = function () {
     this.view.setAttribute('width', w / this.scale);
     this.view.setAttribute('height', h / this.scale);
 }
+
+/** Called from the individual component ondragmove handlers in order to pan
+ * the entire view whenever CTRL or the middle mouse button are pressed.
+ * Returns true if the event has been processed by the NetGraph component and
+ * no further event processing must be performed by the calling code.
+ * Returns false otherwise. */
+Nengo.NetGraph.prototype.capture_move_event = function (event) {
+    // For more constants see https://www.w3.org/TR/uievents/#mouseevent
+    var MOUSE_BUTTON_MIDDLE_MASK = 4
+
+    if (event.ctrlKey || (event.buttons & MOUSE_BUTTON_MIDDLE_MASK)) {
+        interact(this.svg).ondragmove(event);
+        return true;
+    }
+    return false;
+}
--- a/nengo_gui/static/components/component.js
+++ b/nengo_gui/static/components/component.js
@@ -73,6 +73,10 @@ Nengo.Component = function(parent, args) {
                 self.menu.hide_any();
             },
             onmove: function (event) {
+                if (Nengo.netgraph.capture_move_event(event)) {
+                    return;
+                }
+
                 var target = event.target;
 
                 self.x = self.x + event.dx / (self.viewport.w * self.viewport.scale);
--- a/nengo_gui/static/components/netgraph_item.js
+++ b/nengo_gui/static/components/netgraph_item.js
@@ -177,6 +177,10 @@ Nengo.NetGraphItem = function(ng, info, minimap, mini_item) {
                     self.move_to_front();
                 },
                 onmove: function(event) {
+                    if (Nengo.netgraph.capture_move_event(event)) {
+                        return;
+                    }
+
                     var w = self.ng.get_scaled_width();
                     var h = self.ng.get_scaled_height();
                     var item = self.ng.svg_objects[uid];


jgosmann:
This addresses #285 and is a point also mentioned in #731.

jgosmann:
Also, this PR might interact with #806.

astoeckel:
With respect to possible interference with #806 : Despite being a hack, this PR touches very little code and still achieves a significant usability improvement. Since #806 is quite an endeavour with a lot of changes, I would propose to merge this PR (#881) as soon as possible to get better usability with the current master branch. I'd rather implement this behaviour again for the refactored frontend than forcing people to live without the option to pan the viewport for any period of time.

tbekolay:
Yeah, IMO don't worry about #806. It'll be trivial for me to look at the diff and resolve conflicts.

nengo_gui/static/lib/js/interact-1.2.4.js
```diff
@@ -3354,6 +3354,7 @@
         this.shiftKey  = event.shiftKey;
         this.metaKey   = event.metaKey;
         this.button    = event.button;
+        this.buttons   = event.buttons;
```
jgosmann: How do `event.button` and `event.buttons` differ?

nengo_gui/static/components/netgraph.js
```diff
@@ -683,3 +683,16 @@ Nengo.NetGraph.prototype.scaleMiniMapViewBox = function () {
...
     this.view.setAttribute('height', h / this.scale);
 }
+
+/** Called from the individual component ondragmove handlers in order to pan
+ * the entire view whenever CTRL or the middle mouse button are pressed.
+ * Returns true if the event has been processed by the NetGraph component and
+ * no further event processing must be performed by the calling code.
+ * Returns false otherwise. */
+Nengo.NetGraph.prototype.capture_move_event = function (event) {
+    if (event.ctrlKey || (event.buttons & 4)) {
```
jgosmann: I assume `4` is the code for the middle mouse button?

nengo_gui/static/lib/js/interact-1.2.4.js
jgosmann: Also, [it seems that this is fixed in more recent versions of interact.js](https://github.com/taye/interact.js/commit/d384fcc47cbb0f2f47c65dd09e7855befd646522). I would prefer to update interact.js instead of rolling our own patch.

astoeckel: ``event.button`` is the mouse button that triggered the event. Since the ``ondragmove`` event is generated by ``interact.js`` as a response to the DOM ``MouseMoveEvent``, ``button`` does not have any meaning here. ``buttons`` on the other hand contains all mouse buttons that were pressed when the event was generated. This is what we want.

Regarding the more recent version of ``interact.js``: jap, let's use that instead. Should I upgrade ``interact.js`` in this PR or create another one?

nengo_gui/static/components/netgraph.js
astoeckel: Yes, as part of the MouseEvent specification. See https://www.w3.org/TR/uievents/#mouseevent

If you think this makes any sense, I can surely add constants for the mouse buttons instead of relving to rely on magic numbers.

jgosmann: Yes, some sort of named constant would be nice.

nengo_gui/static/lib/js/interact-1.2.4.js
jgosmann: Probably fine in this PR? @tcstewar?

astoeckel: I thought about this and decided to open a separate PR. In case there will be any problems with the new interactjs version this can be tested independent from the changes in this PR.

tcstewar:
This got merged as part of #958    Thank you all!

PR #882, opened 2017-04-06
Update interactjs to version 1.2.6

astoeckel:
PR #881 implements view panning with the middle mouse
button. This required a tiny patch to interactjs.
Fortunately, the corresponding change is already part
of the current interactjs version.

This PR has been filed independently of #881 in order
to facilitate separate testing of the new interactjs
version and #881.


tcstewar:
Thank you!

PR #889, opened 2017-05-05
More informative message when making a Simulator in the GUI

tcstewar:
A number of new users have asked questions like this one:

https://forum.nengo.ai/t/how-to-see-the-plots/282

where they are taking Nengo examples and pasting them into the GUI (rather than running them in a Python interpreter). 

This PR gives a more informative error message when this happens, and directs them to a web page that gives more detail.

There are 2 commits here -- the first one gives the error message, and the second one makes the link work.  I've separated them like that because it's a bit annoying how I'm getting a HTML from the server to the client -- we used to assume that this was always text, not HTML.  If someone has a better way to do this, that'd be good too.

### DIFF
--- a/nengo_gui/exec_env.py
+++ b/nengo_gui/exec_env.py
@@ -117,6 +117,11 @@ class ExecutionEnvironment(object):
 
         sys.stdout = sys.__stdout__
 
+        # ensure what has been printed is safe to show in html
+        s = self.stdout.getvalue()
+        s = s.replace('<', '&lt;').replace('>', '&gt;')
+        self.stdout = StringIO(s)
+
         if not self.allow_sim:
             if self.added_directory is not None:
                 sys.path.remove(self.added_directory)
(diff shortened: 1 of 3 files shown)


Seanny123:
LGTM.

nengo_gui/exec_env.py
```diff
@@ -117,6 +117,11 @@ def __exit__(self, exc_type, exc_value, traceback):
 
         sys.stdout = sys.__stdout__
 
+        # ensure what has been printed is safe to show in html
+        s = self.stdout.getvalue()
+        s = s.replace('<', '&lt;').replace('>', '&gt;')
```
tbekolay: Doesn't this part make the link not work? Like, you want to keep the `<` in what you pass to `.html` so that it renders as an actual link. Or is that not how `.html` works?

tcstewar: This happens before the link is added.  This is to make sure that anything that gets printed with normal print statements doesn't get turned into an HTML tag.  This used to happen automatically because we used `.text` on the javascript side, but we can't do that now because that would change *everything* and make the link not work.

So the process here is to let all the print statements happen, convert them to be HTML-safe, and then after that `exec_env.stdout.write` in `page.py` will happen.  Very confusing.  :(

tbekolay: Confusing but sounds correct ;) :+1:

tcstewar:
Thank you!

PR #891, opened 2017-05-15
Basic support for nengo_spa.

jgosmann:
This adds basic support for `nengo_spa`, but breaks support for `nengo.spa` (note the dot instead of underscore). There might still be things that do not work properly and before a merge the complete SPA related code needs to be gone through and checked.

I think there are slightly to many differences to conveniently support both SPA versions. With some refactoring this might be more realistic, but that also requires more effort and it migth be better to wait for the refactored frontend #806.

Ultimately, I expect these classes to move to `nengo_spa`, but this requires the plugin structure to be in place which I probably won't tackle before the frontend refactoring is done either.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,6 +22,7 @@ Release History
 0.3.2 (unreleased)
 ==================
 
+- Support for nengo_spa
 - Added --browser option
 - Added --unsecure option
 - Fixed backspace not working on sliders, search box
--- a/nengo_gui/components/pointer.py
+++ b/nengo_gui/components/pointer.py
@@ -2,7 +2,8 @@ import copy
 import itertools
 
 import nengo
-import nengo.spa as spa
+import nengo_spa as spa
+from nengo_spa.examine import pairs
 import numpy as np
 
 from nengo_gui.components.component import Component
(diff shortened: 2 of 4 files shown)


tcstewar:
Just before things go too far in this direction, I'm definitely going to make it a requirement that we will have to support `nengo_spa` and `nengo.spa` at the same time.  It does make sense to have a branch like this for working with `nengo_spa`, but we'll have to support both before it gets merged.

tcstewar:
(and, I think you're right that the eventual path to that involves the refactoring and then the plugin system, and then this ends up in `nengo_spa` rather than in `nengo_gui`)

jgosmann:
> Just before things go too far in this direction, I'm definitely going to make it a requirement that we will have to support nengo_spa and nengo.spa at the same time.

Yes, I'm on board with that. I just wanted to have an initial possibility to get the plots for `nengo_spa`. This is not intended to be merged in the current form (but the PR might make it easier to find if anyone needs it).

If `nengo_spa` provides the plots via a plugin system, the problem of supporting both should be solved because nengo_gui can still provide the code for legacy SPA without any changes and nengo_spa would provide the code that it needs itself.

PR #864, opened 2017-02-07
Add dual stack IPv6 support.

jgosmann:
This implements support for IPv6 with a dual stack solution supporting IPv4 and IPv6 at the same time. In theory this also allows to listen on multiple ports or network interfaces at the same time. But this is not exposed within `main.py`, so the `nengo` command only allows to listen on all `localhost` addresses (no password protection) or all interfaces (with password protection). I assume this to be fine in almost all cases and do not intend to change this as long as there are no such requests because it would complicate the interface without need.

Due to the current implementation of the WebSocket origin checking the GUI does not support access by IP address (e.g. `127.0.0.1` or `[::]`), except when started with a password.

Addresses #21.

PR #898, opened 2017-06-05
Fix WebSocket compatibility with Safari.

jgosmann:
Safari complains if the line break is not CRLF, but only LF.

This is cherry-picked from #808 (which has been marked as reviewed more than 9 months ago, but got never merged 😒 Anyways, the newer #864 should be merged before #808 now, but people ran into problems with Safari, so that really should get fixed.)

### DIFF
--- a/nengo_gui/server.py
+++ b/nengo_gui/server.py
@@ -265,12 +265,11 @@ class HttpWsRequestHandler(server.BaseHTTPRequestHandler):
         raise NotImplementedError()
 
     def upgrade_to_ws(self):
-        response = '''HTTP/1.1 101 Switching Protocols\r\n\
-Upgrade: websocket\r\n\
-Connection: Upgrade\r\n\
-Sec-WebSocket-Accept: {sec}\r\n\
-\r\n\
-'''
+        response = (
+            'HTTP/1.1 101 Switching Protocols\r\n'
+            'Upgrade: websocket\r\n'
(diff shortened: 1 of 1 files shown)


jgosmann:
You're right, it's the same string. The issue is actually already fixed in master, but not in the released 0.3.0 ...

tcstewar:
Ah, good!  I'd thought this was in master already....  So I people can grab the github master version to fix this. 

Should we do a 0.3.1 release right now?  It looks like that'll do this fix, the tweak to the error messages, and the fix for slicing on learning rules.  So not a lot, but perhaps well worth it for this bug...

jgosmann:
I'm in favor of a release. If it doesn't have a lot of changes, it is also less likely to break other things.

On June 5, 2017 9:55:10 PM EDT, tcstewar <notifications@github.com> wrote:
>Ah, good!  I'd thought this was in master already....  So I people can
>grab the github master version to fix this. 
>
>Should we do a 0.3.1 release right now?  It looks like that'll do this
>fix, the tweak to the error messages, and the fix for slicing on
>learning rules.  So not a lot, but perhaps well worth it for this
>bug...
>
>-- 
>You are receiving this because you authored the thread.
>https://github.com/nengo/nengo_gui/pull/898#issuecomment-306359274

-- 

jgosmann:
This has been resolved in the 0.3.1 release.

PR #908, opened 2017-06-30
Enable backspace in searchbox.

### LINKED ISSUE #821: Can't use backspace when finding in text editor
When searching for text in the pop-up find dialogue the backspace doesn't work.

ikajic:
Added a few lines of code that allow using backspace in the search box. Should resolve #821.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,6 +22,7 @@ Release History
 0.3.2 (unreleased)
 ==================
 
+- Fixed backspace not working on sliders, search box
 - Added autocomplete to text editor
 - Added visual depiction of modulatory and inhibitory connections
 - Token-based authentication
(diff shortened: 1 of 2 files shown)


ikajic:
> I'm a bit worried that this might be a slippery slope of adding exceptions when the backspace key should not be deactivated. To me it seems that it should never be deactivated for any type of input field. That should be detectable and might be a more general solution to the problem.

I agree with you -- I am sure there is a cleaner and better way to implement this. On the other hand, I find it incredibly annoying not to be able to use backspace when searching, which is something I do often (I don't know for others who don't use GUI).

jgosmann:
Definitely an annoying bug for those who use these features. I added a commit (to be squashed with the other ones) that instead of checking for specific CSS classes looks for HTML tags. It is still two tags that need to be checked, but I believe there aren't any more than the two and it applies to all those tags, not just those with specific CSS classes.

tcstewar:
Thank you Ivana and Jan!  I've combined it all together to one commit, and it looks quite happy to me.

Issue #910, opened 2017-07-19
Configure Settings not working in jupyter notebook


s72sue:
Unable to configure settings when the gui is embedded in a jupyter notebook

Steps to reproduce:
create a nengo model
import the visualizer:  IPythonViz(model)
Click on Utilities -> Configure Settings

You will see that Configure options pane doesn't show up.

Seanny123:
Resolved for me on the master branch, so I'm going to go ahead and close this issue.

PR #911, opened 2017-08-08
Allow configure dialog to open without editor.

jgosmann:
Addresses #910.

The solution is a bit of a hack. The offending checkbox should probably
be completely removed or disabled if there is no editor (at the moment
it can be toggled, but won't have any effect). Also the if-statements
introduced in this commit are a bit of a hack. There is probably some
better way to address this without the repeated checking whether
Nengo.ace is defined. However, it might make sense to wait for the
completion of the frontend refactoring before addressing this.

### DIFF
--- a/nengo_gui/static/modal.css
+++ b/nengo_gui/static/modal.css
@@ -136,3 +136,7 @@ div.modal {
 .modal-dialog.modal-sm td {
     padding: 4px;
 }
+
+#myModalForm .disabled {
+    color: darkgrey;
+}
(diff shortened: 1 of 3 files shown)


tcstewar:
> The solution is a bit of a hack. The offending checkbox should probably
> be completely removed or disabled if there is no editor (at the moment
> it can be toggled, but won't have any effect). Also the if-statements
> introduced in this commit are a bit of a hack. There is probably some
> better way to address this without the repeated checking whether
> Nengo.ace is defined. However, it might make sense to wait for the
> completion of the frontend refactoring before addressing this.

I added quick change to make it visually disabled.  It is a bit of a hack in the code, but I think you're right that this is fine for now, and it'll just be on the list of things that should be made nicer in the refactoring.

Issue #914, opened 2017-09-29
Sessions can consume memory indefinitely


jgosmann:
[Sessions are only ever deleted when they are accessed.](https://github.com/nengo/nengo_gui/blob/master/nengo_gui/guibackend.py#L51) This can be a problem if someone issues a large number of requests to server as the memory consumed by the sessions will never be cleared.

Even if the memory is cleared once a session expires, it a large number of created sessions in a short time frame can still cause a problem (denial of service). The solution to this is probably to limit the number of unauthorized sessions.

jgosmann:
> The solution to this is probably to limit the number of unauthorized sessions.

A better solution is probably to not create a session at all until a successful login.

Issue #817, opened 2016-08-21
IPythonViz does not work in Jupyter Notebook that is not on localhost or accessed via HTTPS


jgosmann:
There is probably a number of problems here, but the first is that browser block non-HTTPS content on HTTPS sites (and they should). That means we should start Nengo GUI in SSL mode, but there a few problems with that:

1. How do we detect if we need to start in SSL mode? The Python code is executed in the IPyKernel that has no knowledge of the frontend accessing it (it could be even multiple frontends). Potentially it is possible to provide both SSL and non-SSL services with a single GUI server (especially given PR #864). However usually this requires two different ports and then we still do not know which of the two ports to connect to. On the frontend site `//some-address` should automatically select the right protocol and either port 80 or 443 as used by normal webservers. However, we cannot expect to be able to run Nengo GUI on those ports.
2. Where do we get the SSL certificates from? Ideally we would reuse the ones used for the Jupyter server, but I don't think we can obtain those paths automatically. We could require the user to provide them, but that's inconvenient. It might be possible to auto-generate them, but this requires the user to click through that scary warning about self-signed certificates (and without the user being aware that certificates were generated, they might think of this as suspicious?).

jgosmann:
I started investigating this issue and updated the description with some information. Comments welcome!

tbekolay:
You can also get free SSL certs from https://letsencrypt.org/ (and in Python through https://github.com/certbot/certbot). Probably this won't work with localhost, but maybe there's a way around that.

jgosmann:
Good idea, but it requires the web server to be accessible over the Internet which will rarely be the case.

On February 8, 2017 5:40:45 PM EST, Trevor Bekolay <notifications@github.com> wrote:
>You can also get free SSL certs from https://letsencrypt.org/ (and in
>Python through https://github.com/certbot/certbot). Probably this won't
>work with localhost, but maybe there's a way around that.
>
>-- 
>You are receiving this because you were assigned.
>Reply to this email directly or view it on GitHub:
>https://github.com/nengo/nengo_gui/issues/817#issuecomment-278485997

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

jgosmann:
[...]

Issue #203, opened 2015-05-27
Add configuration option for default browser


jgosmann:
At the moment the visualizer open in the default browser. It would be nice to overwrite this to use a different installed browser (for example if someone is using an incompatible default browser and does not want to switch completely). IPython notebook has a similar feature.

hunse:
At the very least we should document how to use a different browser with the current setup, since it is possible just by going to the URL.

tcstewar:
It looks like this would be easy to add as a command-line option.  It's just ```webbrowser.get(name).open``` rather than ```webbrowser.open```.

tcstewar:
I just posted a PR that attempts to do this: #934

PR #915, opened 2017-09-29
Token based authentication

### LINKED ISSUE #867: Implement token-based authentication
The Nengo GUI should always require authenticated access because it can run arbitrary code with the permissions of the user it runs under. Even for a GUI server just accepting connections from localhost only this should be done because on a multiuser machine other people could connect to the server and run code with another user's permissions.

[Jupyter solved this with tokens.](http://blog.jupyter.org/2016/12/21/jupyter-notebook-4-3-1/) It shouldn't be too complicated to implement something similar for Nengo GUI.

jgosmann:
This resolves some security problems on multi-user systems. Prior to these changes any user on the local system could have connected to the Nengo GUI server to run arbitrary code under the permissions of a different user. With the token based authentication such an evil user has to obtain the token first.

This is similar to how Jupyter Notebook handles things.
    
Fixes #867. Also fixes some Python 3 issues with the password protection.

Another benefit is that this allows to automatically login into a password protected server. Though maybe that might give the impression that the password protection does not work?

At the moment the only way to have a Nengo GUI server that listens for outside connections is to spawn a password protected server. I wonder whether we should add the possibility to allow outside connections even without a password as the token based authentication will still be required? If so, should this be restricted to SSL connections? (I think Jupyter Notebook only allows to enable outside connections only together with SSL and it probably makes sense as sniffing the token gives full access.) Of course one could set a password (hopefully secure) dummy password and still use the token ...

### DIFF
--- a/nengo_gui/gui.py
+++ b/nengo_gui/gui.py
@@ -70,8 +70,9 @@ class InteractiveGUI(BaseGUI):
 
     def start(self):
         protocol = 'https:' if self.server.settings.use_ssl else 'http:'
-        print("Starting nengo server at %s//%s:%d" %
-              (protocol, 'localhost', self.server.server_port))
+        print("Starting nengo server accessible at:\n  %s//%s:%d/?token=%s" % (
+            protocol, 'localhost', self.server.server_port,
+            self.server.settings.auth_token))
 
         if not sys.platform.startswith('win'):
             signal.signal(signal.SIGINT, self._confirm_shutdown)
@@ -141,7 +142,9 @@ class GUI(InteractiveGUI):
     def start(self):
         t = threading.Thread(
             target=webbrowser.open,
-            args=('http://localhost:%d' % self.server.server_port,))
+            args=('http://localhost:%d/?token=%s' % (
+                self.server.server_port,
+                self.server.settings.auth_token,)))
         t.start()
 
         super(GUI, self).start()
--- a/nengo_gui/gui.py
+++ b/nengo_gui/gui.py
@@ -72,7 +72,7 @@ class InteractiveGUI(BaseGUI):
         protocol = 'https:' if self.server.settings.use_ssl else 'http:'
         print("Starting nengo server accessible at:\n  %s//%s:%d/?token=%s" % (
             protocol, 'localhost', self.server.server_port,
-            self.server.settings.auth_token))
+            self.server.auth_token))
 
         if not sys.platform.startswith('win'):
             signal.signal(signal.SIGINT, self._confirm_shutdown)
--- a/nengo_gui/guibackend.py
+++ b/nengo_gui/guibackend.py
@@ -20,7 +20,7 @@ import nengo_gui
 import nengo_gui.exec_env
 import nengo_gui.page
 from nengo_gui import server
-from nengo_gui.password import checkpw
+from nengo_gui.password import checkpw, gensalt
 
 
 logger = logging.getLogger(__name__)
@@ -86,10 +86,14 @@ class RequireAuthentication(object):
     def __call__(self, fn):
         def auth_checked(inst):
             session = inst.get_session()
-            has_password = inst.server.settings.password_hash is not None
-            if has_password and not session.authenticated:
-                return server.HttpRedirect(self.login_page)
-            return fn(inst)
+            if session.authenticated:
+                return fn(inst)
+            elif ('token' in inst.db and
+                    inst.db['token'] == inst.server.settings.auth_token):
+                session.authenticated = True
+                session.login_host = inst.headers.get('host', None)
+                return fn(inst)
+            return server.HttpRedirect(self.login_page)
         return auth_checked
 
 
@@ -118,18 +122,25 @@ class GuiRequestHandler(server.HttpWsRequestHandler):
         session = self.get_session()
         content = b''
 
+        if session.authenticated:
+            return server.HttpRedirect('/')
+
+        if self.server.settings.password_hash is None:
+            return server.HtmlResponse(content + b'''
+                <p>Password authentication not enabled, please use the token
+                link.</p>
+            ''')
+
         if 'pw' in self.db:
             if checkpw(self.db['pw'], self.server.settings.password_hash):
                 session.authenticated = True
                 session.login_host = self.headers.get('host', None)
+                return server.HttpRedirect('/')
             else:
                 content += b'<p><strong>Invalid password. Try again.'
                 content += b'</strong></p>'
         else:
             content += b'<p>Please enter the password:</p>'
-
-        if session.authenticated:
-            return server.HttpRedirect('/')
 
         return server.HtmlResponse(content + b'''
             <form method="POST"><p>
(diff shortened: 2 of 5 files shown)


jgosmann:
The token based authentication is also needed to properly implement the `IPythonViz` functionality in Jupyter Notebooks accessed via SSL, see #817.

jgosmann:
I added a commit to fix #914.

jgosmann:
I added a commit to allow to provide the token in the `Authorization` HTTP header or via the password field. [Jupyter is allowing the same.](https://jupyter-notebook.readthedocs.io/en/stable/security.html) In addition, Jupyter generates an additional one-time token for the browser link and I add the same in another commit. This is required because opening the browser might expose the token to other users via the system's process list.

jgosmann:
I added a commit to simplify the websocket origin checking. As the user logs in even on pure localhost servers, the login host can always be used for the origin. This supersedes the logic used in #808 and addresses some of the issues reported in #807.

tcstewar:
I just realized one complication with this system:  it is still common for people to have nengo starting up the wrong browser, and then they open Chrome and connect to localhost:8080.  This token system specifically stops that approach.

My thought is that we can address this by adding a parameter to nengo_gui to control which browser pops up, which seems like a good idea anyway and is supported by Python fairly well (https://docs.python.org/2/library/webbrowser.html).  I'll make a separate PR for that, and that should eliminate that use case of needed another token.

That said, are there any other situations where we might need to generate another token?  Is there an interface for doing so?

jgosmann:
Note that the token link printed on the command line is **not** a one-time token. So you can use that link to open another browser. This is also how Jupyter does it, though, Jupyter also provides a command to list all running servers with their instances which we don't.

> My thought is that we can address this by adding a parameter to nengo_gui to control which browser pops up

Yes! I wanted that for more than two years: #203

> That said, are there any other situations where we might need to generate another token? Is there an interface for doing so?

Whenever the token gets exposed through an insecure channel. The `GuiServer.gen_one_time_token()` provides an interface to obtain such a token, but of course ideally one would avoid exposing the token.

jgosmann:
Also, if you don't use the token URL, you wil be redirected to the login page where you can enter the token into the password field (Jupyter allows the same thing). Or you can set a password and use that to login.

tcstewar:
> Note that the token link printed on the command line is not a one-time token. So you can use that link to open another browser. This is also how Jupyter does it,

!!!  Cool!  For some reason I had it in my mind that that was a one-time token....   weird.....

tcstewar:
I very much like this.  A few small (hopefully) things come to mind:

- the message `Starting nengo server at http://localhost:8080/?token=7cd68a03ad39249e8e3aac0d4abefd706450876a0d62d9b3`  should probably be changed to something like `Starting nengo server.  Access it at http://localhost:8080/?token=7cd68a03ad39249e8e3aac0d4abefd706450876a0d62d9b3`.  A carriage return in there would be nice to simplify copy-and-paste for people with 80-character-wide text windows....

- when run without an SSL password, when I go to http://localhost:8080 I get a login page with no valid password (and on firefox I don't even get the login page).  Would it be possible to have an error message instead telling people to go find the link with the token where they started nengo from?

jgosmann:
> I had it in my mind that that was a one-time token.... weird.....

The token that is used to open the browser automatically is a one-time token because opening the browser with a link will expose it shortly in the system's process list. But this is not the case when the user copies the token from the command line and pastes it into a browser.

Start up message can be changed to anything you like.



> when run without an SSL password, when I go to http://localhost:8080

Did you mean to type http**s**:// or did you indead mean http://?

> I get a login page with no valid password (and on firefox I don't even get the login page).

Sounds like a bug to me ... but it might make sense to add some more explanation to the login page (the Jupyter notebook login page has a somewhat lengthy explanation on it).

tcstewar:
> Did you mean to type https:// or did you indead mean http://?

`http`, since that's running without a password.

jgosmann:
Oh, I somehow misread. I get the login page on both Firefox and Chrome; and the token works fine as password.
What exactly do you get in Firefox? Can you check the network requests that are actually made (and their responses) in the developer console?

tcstewar:
> I get the login page on both Firefox and Chrome; and the token works fine as password.

Ah, looks like I copied-and-pasted wrong.  It works fine for me in Chrome

> What exactly do you get in Firefox?

Here's all I get:

image

> Can you check the network requests that are actually made (and their responses) in the developer console?

and here's the developer console (I'm not quite sure this is what you were looking for:)

image

tcstewar:
(the Response payload is empty, and this is on Firefox 56.0.2)

jgosmann:
That bug is fixed by some commit in some PR ... I'll look for it.

jgosmann:
PR #919 should solve the problem, more specifically commit fcc975df2f7ed6a62ccf268a364b8506e856dc1d.

tcstewar:
> PR #919 should solve the problem, more specifically commit fcc975d.

Sweet!  I'll get that merged in right now (no need to bring it into this PR).  

With that fixed, the only thing I think this PR needs is to modify the text a bit in two places:  1) the text printed to the console, and 2) the text on the login screen (letting people know they can use the token in the console as a password or just use the url from the console).  Shall I try making those changes?

jgosmann:
Sure, go ahead. ✍️

jgosmann:
The login page could probably also use some CSS styling ... 💅

tcstewar:
I just added some quick text fixups for those two spots.  They're not wonderful, but at least now I don't feel like someone who was used to the old way will be completely lost when presented with the new system.

I'm going to go check if the Firefox problems are solved via that other PR, and if so I'm quite happy to merge this one.  :)

jgosmann:
Is the space before the URL on purpose?

tcstewar:
> Is the space before the URL on purpose?

Yes, I just think it looks slightly nicer.  Could also have added a blank line above and below it...  but mostly I wanted it on its own line so that it's easier to copy-and-paste it in some consoles that don't handle word-wrapping well (i.e. Windows)

tcstewar:
Excellent.  Firefox now works great for me.  :)   

There are a lot of commits here, but I think I'll keep them anyway (other than the fixups, of course), as they all seem pretty atomic.  :)

jgosmann:
> Yes, I just think it looks slightly nicer. Could also have added a blank line above and below it... but mostly I wanted it on its own line so that it's easier to copy-and-paste it in some consoles that don't handle word-wrapping well (i.e. Windows)

To me it looks like a slight accident. If it were two space it would appear to be more on purpose and give a slightly better visual indentation. Blank lines would be fine too. (Or I could try to be not to OCD about this.)

> There are a lot of commits here, but I think I'll keep them anyway (other than the fixups, of course), as they all seem pretty atomic. :)

👍

tcstewar:
> To me it looks like a slight accident. If it were two space

Two spaces sounds good to me.  :)  And it still (just barely) fits within an 80-column console.  :)

tcstewar:
Merged!  Thanks for this -- Got to have this security in place, and I'm very excited about it leading to being able to run nengo_gui remotely within jupyter notebook!

tcstewar:
(sorry about github claiming there are "unmerged commits"....  that's because I did a fixup to add a line to CHANGES.rst to say what happened, but I only did that on master and not on this branch as well....)

PR #919, opened 2017-10-29
Fixes to the HTTP response headers

jgosmann:
Add the content length field to the header as otherwise certain clients might not read all data and prevent an empty line (effectively ending the headers) when no cookies are set.


tcstewar:
Thanks!  :)

PR #921, opened 2017-10-30
Implement tunneling of the GUI through Jupyter notebook

### LINKED ISSUE #817: IPythonViz does not work in Jupyter Notebook that is not on localhost or accessed via HTTPS

jgosmann:
This PR implement tunneling of connections to the Nengo GUI through the Jupyter notebook server. This allows to use `IPythonViz` in notebooks accessed within an SSL protected Jupyter environment or in notebooks that are not accessed from localhost.

This PR is based on #915 and has the commits from #919 cherry-picked. Furthermore, this PR contains a number of other fixes that could be extracted into separate PRs as those should be merged independent of the tunneling functionality (but are mostly required for the tunneling functionality too):

* 6fbb4c81d1d920f1dce86d0853f196b0b533fc9f: Block JS access to session cookie
* 06b445df17f3409d9f77e52610713e6c72a51556: Use HTTP/1.1 for HTTP responses
* 0bc46f13991f1b73de5151aa481d9bd0f89d2166: Use cookies library supporting newer RFC

Closes #817.

To try it out the Jupyter server extension has to be enabled with:

```bash
jupyter serverextension enable nengo_gui.jupyter
```

Still todo:

- [x] Vendorize the cookies library.
- [x] Error handling in the tunneling request handlers.
- [x] [Favicon should be specified in `link` tag](https://en.wikipedia.org/wiki/Favicon#How_to_use) for full prefix support.
- [x] Kernel shutdown takes a long time due to the Nengo GUI server. Can we fix that?
- [x] Some cleanup of the tunneling code
- [x] Installation instructions.


jgosmann:
This is now finished. If the server extension is not activated and one tries to use `IPythonViz`, it will show instructions on how to activate the extension.

Given the major changes in the PR, should we consider updating the name of the module itself and/or `IPythonViz`?

nengo_gui/static/nengo.js
```diff
@@ -39,7 +39,9 @@ Nengo.create_websocket = function(uid) {
     } else{
         var ws_proto = 'ws:';
     }
-    var ws_url = ws_proto + '//' + parser.host + '/viz_component?uid=' + uid;
+    var ws_url = ws_proto + '//' + parser.host + window.location.pathname +
+        'viz_component?uid=' + uid;
+    console.log(ws_url)
```
tcstewar: I assume this console.log should be removed?  :)

jgosmann: yeah

jgosmann:
This should be merged with #923 because of f07406b4b634ee73b65b1c851c642e4451020e7c.

jgosmann:
It would be nice to have this merged.

nengo_gui/ipython.py
```diff
@@ -33,7 +33,7 @@ class ConfigReuseWarning(UserWarning):
     pass
 
 
-class IPythonViz(object):
+class InlineGUI(object):
```
tcstewar: I am very much on-board with renaming this, but we'll definitely want a deprecation system here, for backwards compatibility...

jgosmann: Added deprecation warnings.

jgosmann:
I hope that this is ready for review now. Seems to work in Notebook, Lab, and Binder. Turned out that the Jupyter instance might use a base URL that needs to be added to Nengo GUI's prefix, but we can't determine the base URL for `InlineGUI`. Thus I implemented something that allows setting the base URL from the first incoming request. A little bit of a hack, but seems to be the least hacky way to get things working properly.

jgosmann:
@tcstewar What's the roadmap for merging this PR?

tcstewar:
> @tcstewar What's the roadmap for merging this PR?

It looks like there's no unmerged PRs left this depends on, so this PR seems like something that should be up next.  I guess step 1 would be to rebase it to master....  that said, I'm not going to have a chance to look at it closely this week.....  but that seems like a good way to start.

jgosmann:
Rebased.

jgosmann:
@tcstewar please review

PR #923, opened 2017-10-30
Do not allow access to resources outside of static.

jgosmann:
This is another security issue. By providing paths like `/static/../__init__.py` it was possible to access arbitrary files on the file system. In `master` this requires at least authentication, but in #921 I'm removing the constraint because the favicon should be available without login and we might want to use some of the style sheets on the login page at some point.

### DIFF
--- a/nengo_gui/guibackend.py
+++ b/nengo_gui/guibackend.py
@@ -150,7 +150,11 @@ class GuiRequestHandler(server.HttpWsRequestHandler):
     @RequireAuthentication('/login')
     def serve_static(self):
         """Handles http://host:port/static/* by returning pkg data"""
-        fn = os.path.join('static', self.resource)
+        assert self.resource[0] == '/'
+        static_dir = 'static' + os.sep
+        fn = os.path.normpath(self.resource[1:])
+        if os.path.commonprefix((static_dir, fn)) != static_dir:
+            raise server.Forbidden()
         mimetype, encoding = mimetypes.guess_type(fn)
         data = pkgutil.get_data('nengo_gui', fn)
         return server.HttpResponse(data, mimetype)


tcstewar:
On Windows, this fails because the string passed to `pkgutil.get_data` is of the form `\static\....`.  Windows then decides to treat this as `c:\static\....` which it can't find.

Adding this line seems to fix it in windows: `fn = os.path.relpath(fn, '/')`.  Alternatively, `fn = fn[1:]` could also work.  I'm not sure if either of these are the correct solution, though....

jgosmann:
I have to think about this.
In the meantime, I'd like to complain that drive letters are stupid idea (imo). ;)

tcstewar:
> In the meantime, I'd like to complain that drive letters are stupid idea (imo). ;)

<grins>  Agreed.  :)

jgosmann:
I made a commit that should make this work on Windows too, but I haven't tested it on either OS.

jgosmann:
Tested the change on Linux and works. Because browsers collapse `..` in the URL, one has to use a tool like `netcat`. For example:
```console
$ netcat localhost 8080
GET /static/../main.py?token=[TOKEN] HTTP/1.1
Host: localhost

```
(you need two newline at the end)

With this PR it will give a 403 response
```http
HTTP/1.0 403 Forbidden
Server: BaseHTTP/0.6 Python/3.6.1
Date: Tue, 12 Dec 2017 20:54:41 GMT
Content-type: text/html
Set-Cookie: _session_id=1e2134ff11a84d8775629e6f2e108c91cfc349ca

```
whereas master will provide `main.py`:
```http
HTTP/1.0 200 OK
Server: BaseHTTP/0.6 Python/3.6.1
Date: Tue, 12 Dec 2017 20:54:01 GMT
Content-Type: text/x-python
Content-Length: 4129
Set-Cookie: _session_id=489cd5aa032ae1a5b1c625659685cc0d24970a63

import argparse
import errno
import logging
import os.path
import threading
import webbrowser
...
```

tcstewar:
> Because browsers collapse .. in the URL, one has to use a tool like netcat

Nice!  That's incredibly useful -- thank you...    It works on Windows, btw...  :)

tcstewar:
This should have been merged ages ago....  thank you, @jgosmann , it's merged now!

PR #924, opened 2017-11-03
Fix crash with voltage graph

tbekolay:
The voltage plot is broken at the moment, as [reported on the forum](https://forum.nengo.ai/t/on-and-off-error-signal-neurons-of-ensemble-a-voltage-slice-none-5-none-shape-5/417/1). I'm not sure when that started, but it seems to be an issue with voltage re-using the Value plot infrastructure in JavaScript, but not Python. This should be done better, and it is in the refactoring, but for the time being this PR fixes the voltage plot crash.

### DIFF
--- a/nengo_gui/components/voltage.py
+++ b/nengo_gui/components/voltage.py
@@ -10,8 +10,9 @@ from nengo_gui.components.component import Component
 class Voltage(Component):
     """Represents neuron voltage over time."""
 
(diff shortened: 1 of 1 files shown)


tcstewar:
Looks good to me!  This is nice and small and quick, so I'm just going to merge it now....  Thank you!

Issue #926, opened 2017-11-03
Implement a statusbar for global tooltips


astoeckel:
Depending on the context (e.g. the mouse being over a certain element), the GUI may allow certain  modifier keys to change the behaviour of actions, such as mouse clicks/dragging (see #881). However, these modifiers are currently not discoverable by the user. To enhance discoverability, many applications such as MS Office, Gimp, or Inkscape possess a status bar (see [1] for evidence). 

[1] https://msdn.microsoft.com/en-us/library/windows/desktop/bb760728(v=vs.85).aspx

astoeckel:
I'm going to implement a prototype of this concept which works in conjunction with #881, so this feature can be evaluated.

jgosmann:
This can be closed I assume?

PR #927, opened 2017-11-03
Implement basic status bar

astoeckel:
See #926. This adds a global status bar. It can be accessed using the following functions:

- Nengo.status_bar.set_caption()
- Nengo.status_bar.get_caption()
- Nengo.status_bar.hide()
- Nengo.status_bar.show()

Furthermore, this PR implements basic display of possible mouse/keyboard interactions with the object the mouse is currently over. Especially, it will tell you that you can
- Drag to pan if the mouse is over the netgraph
- Press CTRL to pan if over any object
- Drag to resize if over boundary of an object
- Drag to change slider value

Where the help string is adapted to the actual object type the mouse is over (graph, ensemble, node, network).

### DIFF
--- a/nengo_gui/static/sim_control.css
+++ b/nengo_gui/static/sim_control.css
@@ -3,7 +3,6 @@
 .sim_control {
     background-color: #eee;
     box-shadow: 0px -2px 4px 0px #999;
-    height: 80px;
     position: relative;
     overflow: hidden;
     flex: 0 0 auto;
--- a/nengo_gui/static/sim_control.css
+++ b/nengo_gui/static/sim_control.css
@@ -9,7 +9,11 @@
     flex: 0 0 auto;
     -webkit-flex: 0 0 auto;
     width: 100%;
-    z-index:99999999;
+    z-index:99999990; /* The SVG elements in the netgraph use z-index, so make
+                         sure they never cover the control panel. The z-index
+                         here is smaller than the 99999999 used by other UI divs,
+                         so other controls (e.g. the status bar) can cover the
+                         box shadow. */
 }
 
 .time_slider{
--- a/nengo_gui/static/sim_control.css
+++ b/nengo_gui/static/sim_control.css
@@ -20,6 +20,7 @@
     border: 1px solid #ccc;
     border-radius: 4px;
     overflow: hidden;
+    height: 60px;
 }
 
 .time_slider .shown_time {
--- a/nengo_gui/static/status_bar.js
+++ b/nengo_gui/static/status_bar.js
@@ -86,6 +86,6 @@ Nengo.StatusBar = (function () {
 (function () {
     "use strict";
 
-    Nengo.status_bar = new Nengo.StatusBar(document.querySelector("#main"));
+    Nengo.status_bar = new Nengo.StatusBar(document.querySelector("#control"));
     Nengo.status_bar.set_caption("Welcome to Nengo GUI.");
 })();
--- a/nengo_gui/static/status_bar.js
+++ b/nengo_gui/static/status_bar.js
@@ -1,0 +1,91 @@
+/**
+ * Simple JavaScript status bar which is used at the bottom of the net graph
+ * window.
+ *
+ * Creates a global instance Nengo.status_bar inside the netgraph which can be
+ * accessed using the following functions:
+ * - Nengo.status_bar.set_caption()
+ * - Nengo.status_bar.get_caption()
+ * - Nengo.status_bar.hide()
+ * - Nengo.status_bar.show()
+ */
+
+Nengo.StatusBar = (function () {
+    "use strict";
+
+    /**
+     * StatusBar constructor. Instantiates the DOM for the status bar and
+     * attaches it to the given parent DOM element.
+     *
+     * @param parent is the DOM node the StatusBar should be attached to.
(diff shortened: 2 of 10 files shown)


tcstewar:
> Integration with #881 will be in that PR.

I would much prefer if the #881 PR could be merged without making it dependent on this one....

astoeckel:
@tcstewar  Sure, I'll rebase this commit on #881 and implement the functionality here (because the status bar without any meaningful content probably doesn't make sense anyways).

nengo_gui/static/sim_control.css
```diff
@@ -9,7 +9,8 @@
     flex: 0 0 auto;
     -webkit-flex: 0 0 auto;
     width: 100%;
-    z-index:99999999;
+    z-index:99999998; /* The SVG elements in the netgraph use z-index, so make
+                         sure they never cover the control panel */
```
jgosmann: It is easy to miss that this number ends in 8. Maybe point at that in the comment (and say why)?

nengo_gui/static/status_bar.js
```diff
@@ -0,0 +1,90 @@
...
+ *
+ * Creates a global instance Nengo.status_bar inside the netgraph which can be
+ * accessed using the following functions:
+ * - Nengo.status_bar.set_caption()
+ * - Nengo.status_bar.get_caption()
+ * - Nengo.status_bar.hide()
+ * - Nengo.status_bar.show()
+ *
+ * @author Andreas Stöckel
+ * @date November 2017
```
jgosmann: We usually do not include authorship and date information in the source files. The VCS takes care of that information more accurately anyways.

nengo_gui/static/status_bar.js
```diff
@@ -0,0 +1,90 @@
...
+        "use strict";
+
+        // Create the component by manually assembling the DOM
+        this.status_bar_div = document.createElement('div');
+        this.status_bar_div.classList.add('status_bar');
+
+        // Timeout for updating the caption
+        this.caption_update_timeout = null;
+
+        // Append the component to the parent container
```
jgosmann: This comment doesn't really say anything, that the next line doesn't say already. I would remove it.

nengo_gui/static/status_bar.js
astoeckel: Thanks, changed that.

nengo_gui/static/status_bar.js
astoeckel: Removed last comment and reworded first comment

nengo_gui/static/sim_control.css
astoeckel: I updated the comment accordingly. Would be nice to get rid of such magic numbers.

jgosmann: Yeah, but how?

jgosmann: 🤔

astoeckel:
I've implemented a basic set of interaction displays in the status bar. As listed in the commit message, I'm displaying:

- Drag to pan if the mouse is over the netgraph
- Press CTRL to pan if over any object
- Drag to resize if over boundary of an object
- Drag to change slider value

This definitively improves the discoverability of the middle mouse button/CTRL pan action, and I also find it helpful that the object type, e.g. 'node', 'ensemble', 'graph', etc. is shown in the status bar. In the future, once it is actually possible to select objects, it might make sense to show information such as '3 nodes, 2 ensembles (100 neurons) selected.' (compare Inkscape).

As for the position of the status bar I'm not entirely sure whether it should stay where it is. In case it is too distractive where it is right now, it might also be moved to the very bottom of the window.

Tested with Firefox and Chromium on Linux.

Feedback welcome!

nengo_gui/static/sim_control.css
astoeckel: Just checked the DOM again. I don't see any need for the z-index at all (the comment is based on what Terry told me). The netgraph is contained within a DOM element with `overflow: hidden` and there are no elements with `position: fixed`, so these are two completely separate object stacks.

jgosmann:
> This definitively improves the discoverability of the middle mouse button/CTRL pan action,

👍 

> I also find it helpful that the object type, e.g. 'node', 'ensemble', 'graph', etc., is shown in the status bar

I don't find that information very salient because it is hidden in the sentence. Putting it at the front of the sentence (“Node: [you can do these things with it]”) or adding some styling (bold) to it might help.

> As for the position of the status bar I'm not entirely sure whether it should stay where it is. In case it is too distractive where it is right now, it might also be moved to the very bottom of the window.

I'd move it to the bottom of the window (though I have to check what that look like). It is also cut-off by the console at the right it the moment which looks super weird to me.

astoeckel:
Thanks for the feedback, I'll try to incorporate it into a next revision as soon as I find the time (which could take two weeks).

jgosmann:
@astoeckel Do you still intend to work on this? I was thinking about integrating the build progress bar with Nengo GUI and a status bar might be the logical place to put it.

jgosmann:
I added my proposed changes. I think this should be ready for review now?

tcstewar:
This has now been merged as part of #958   Thank you @astoeckel for doing this!  I really like how it turned out!

Issue #598, opened 2015-11-10
Display of Connection types


tcstewar:
It'd be nice to have a few different Connection types displayed in some way.  For example:

- normal NEF connections
- inhibitory connections
- excitatory connections (?)
- learning connections
- modulatory connections

These could be different end-markers on the lines, or maybe different colours.

astoeckel:
Since this is a feature I wanted to have myself for quite a long time here, I'm going to assign myself, as discussed with @tcstewar and @jgosmann today. Plan is to merge such "small" UI changes before the refactoring to try them out, being perfectly aware that they will be thrown away once the refactoring is done.

astoeckel:
I've implemented a first version of this (see pull request #XXX). So far I've got inhibitory, normal, and connections attached to a learning rule (I call those modulatory in the code, may be wrong). @tcstewar  can you elaborate on the difference between a learning connection and a modulatory connection?

tcstewar:
> can you elaborate on the difference between a learning connection and a modulatory connection?

The "learning connection" is the one whose weights are changing (i.e. the one with the learning rule), and the "modulatory connection" is the one providing the signal that changes the weights (i.e. the one connected to the learning rule).  :)

astoeckel:
Ah, then I've used the terms correctly, thanks for the clarification. Any idea how the "learning connection" may be displayed? I figure that in my current implementation it should be fairly obvious that a connection is a "learning connection" because it has the modulatory connection attached to it.

krisfrenette:
Is this completed then? If not can someone please add what is left to complete this story.

tcstewar:
I think we've got the important ones added.  There's more that could be done, but I don't think there's a strong consensus about what extra indications would be good without adding unnecessary clutter.

Issue #879, opened 2017-03-27
Adding indentation breaks nengo_gui


jgosmann:
(This might be related to #877 and/or #763.)

Open `nengo_gui` without any arguments to get the default network:

```python
import nengo

model = nengo.Network()
with model:
    stim = nengo.Node([0])
    a = nengo.Ensemble(n_neurons=50, dimensions=1)
    nengo.Connection(stim, a)
```

Then insert the following network line:
```python
import nengo

model = nengo.Network()
with model:
    with nengo.Network():
    stim = nengo.Node([0])
    a = nengo.Ensemble(n_neurons=50, dimensions=1)
    nengo.Connection(stim, a)
```

This will produce an error (so far so good and still expected), but now start indenting the lines starting at `stim = ...`:

```python
import nengo

model = nengo.Network()
with model:
    with nengo.Network():
        stim = nengo.Node([0])
        a = nengo.Ensemble(n_neurons=50, dimensions=1)
        nengo.Connection(stim, a)
```

The stim and a objects will stay outside of the network, but the connection disappears, until opening the network, than it will reappear, which I suppose is actually correct behaviour in a way. Opening the network will leave it empty. Close it and open it again and now the objects are drawn in the network, but the outside objects still exist. Clicking and dragging the outside objects will move the inside objects. Same for resizing. I played around with it a bit more (not sure what I did) and the netgraph stopped reacting at all.

![spectacle h29184](https://cloud.githubusercontent.com/assets/850157/24375679/326ba7d0-1307-11e7-96e0-c249432384f6.png)

Issue #877, opened 2017-03-21
nengo_gui assumes reference is to an object in the model


tcstewar:
Consider this code:

```python
import nengo
model = nengo.Network()
with model:
    a = nengo.Ensemble(100, 1)    
    c = nengo.Connection(a, a)    
#model.connections.remove(c)
```

This works fine.  But try uncommenting the last line.  nengo_gui doesn't realize that the object that it used to refer to as `c` is gone from the model, since it still has access to something called `c`.  So it doesn't delete the recurrent connection.  Note that this is not a problem in this case:

```python
import nengo
model = nengo.Network()
with model:
    a = nengo.Ensemble(100, 1)    
    nengo.Connection(a, a)
model.connections.pop(0)
```

The part of the code that's causing this is here:

https://github.com/nengo/nengo_gui/blob/master/nengo_gui/components/netgraph.py#L126

Basically, `uid` is the python text that refers to the item we're currently considering (`'c'` in this case), and old_item is the object that that text used to refer to (the previous time we evaluated the code), and new_item is the new object.  The code then compares the two to see if there's anything we need to do to update the gui.

When I wrote that, I assumed that anything that we can still access still exists in the model, but that's not necessarily the case, especially as I've gotten more and more used to doing helper functions that modify the nengo network.

I'm not sure what to do about this yet.  The only correct solution I can think of is to also check that every object is inside the `model` graph, but that involves doing a scan through the *entire* model, since it might be nested ridiculously deep.  I can cache this scan so that we only have to do it once, but so far I'd been avoiding doing such a complete scan (right now, it only scans into Networks that are not collapsed in the gui).  

It's probably not too expensive in terms of time to do that scan, but I was really hoping to avoid it....

jgosmann:
I think that function needs some refactoring anyways: #767.
Also, there is a somewhat related bug that has been fixed in so far that it doesn't crash the GUI anymore, but the underlying code is still not correct: #763 [This comment might be most useful in explaining the problem.](https://github.com/nengo/nengo_gui/issues/763#issuecomment-224713123)
I remember that I was trying to fix it, but it wasn't easy because the complexity of `_reload` (and maybe a proper qix would have needed some major refactoring? I don't recall).

Maybe we can solve all those problems at once?

jgosmann:
> The only correct solution I can think of is to also check that every object is inside the model graph, but that involves doing a scan through the entire model, since it might be nested ridiculously deep.

[...]

PR #929, opened 2017-11-04
Connection types

astoeckel:
This is a first implementation of #598.

Example of new connection types

I've split the code into four small commits, I think it makes most sense to go through these those independently when reviewing.

To summarise:
* *Commit 1*: Is the entire server-side implementation. Adds the connection `kind` to the `info` structure. This should be fairly tidy, and I've added some unit-tests here.
* *Commit 2*: Is a essentially a two-line change to the Python code adding the `kind` as a `class` and some changes to the CSS for adapting stroke and colour according to the `kind`. Furthermore, I changed the marker symbols for non-normal connections (inspired by what I usually end up using when I manually draw these diagrams in Inkscape, see [1]; it's trivial to change that to something else), though I did not touch the marker symbol code for minimaps.
* *Commit 3*: Is a minimal change which just centres the marker on the post-object in case the connection is modulatory.
* *Commit 4*:  Series of dirty hacks on the client-side to allow connections to connection objects.

[1] https://rawgithub.com/astoeckel/syde_750_project_lifespan_inference/master/doc/media/diag_px.svg

### DIFF
--- a/nengo_gui/components/netgraph.py
+++ b/nengo_gui/components/netgraph.py
@@ -319,22 +319,10 @@ class NetGraph(Component):
                     changed = True
 
         elif isinstance(old_item, nengo.Connection):
-            old_pre = old_item.pre_obj
-            old_post = old_item.post_obj
-            new_pre = new_item.pre_obj
-            new_post = new_item.post_obj
-            if isinstance(old_pre, nengo.ensemble.Neurons):
-                old_pre = old_pre.ensemble
-            if isinstance(old_post, nengo.connection.LearningRule):
-                old_post = old_post.connection.post_obj
-            if isinstance(old_post, nengo.ensemble.Neurons):
-                old_post = old_post.ensemble
-            if isinstance(new_pre, nengo.ensemble.Neurons):
-                new_pre = new_pre.ensemble
-            if isinstance(new_post, nengo.connection.LearningRule):
-                new_post = new_post.connection.post_obj
-            if isinstance(new_post, nengo.ensemble.Neurons):
-                new_post = new_post.ensemble
+            old_pre = NetGraph.connection_pre_obj(old_item)
+            old_post = NetGraph.connection_post_obj(old_item)
+            new_pre = NetGraph.connection_pre_obj(new_item)
+            new_post = NetGraph.connection_post_obj(new_item)
 
             old_pre = self.page.get_uid(old_pre)
             old_post = self.page.get_uid(old_post)
@@ -345,12 +333,8 @@ class NetGraph(Component):
 
             if new_pre != old_pre or new_post != old_post:
                 # if the connection has changed, tell javascript
-                pres = self.get_parents(
-                    new_pre,
-                    default_labels=new_name_finder.known_name)[:-1]
-                posts = self.get_parents(
-                    new_post,
-                    default_labels=new_name_finder.known_name)[:-1]
+                pres, posts = self.get_connection_hierarchy(new_item,
+                    default_labels=new_name_finder.known_name)
                 self.to_be_sent.append(dict(
                     type='reconnect', uid=uid,
                     pres=pres, posts=posts))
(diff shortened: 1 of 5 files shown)


astoeckel:
Fixed some code in the last commit. There was shared functionality in `set_post`, `set_pre`, and `remove` that was extended in the first to methods, but not the last one. Made `remove` use `set_post` and `set_pre`
to avoid this redundancy.

nengo_gui/components/netgraph.py
```diff
@@ -580,24 +583,81 @@ def send_pan_and_zoom(self, client):
...
+    def connection_kind(conn):
+        """
+        Categorises the given connection into one of three kinds: modulatory,
+        inhibitory, and normal.
+        """
+        if isinstance(conn.post_obj, nengo.connection.LearningRule):
+            return "modulatory"
+        if isinstance(conn.post_obj, nengo.ensemble.Neurons):
+            trafo = conn.transform
+            if trafo.size > 0 and np.all(trafo < 0):
```
jgosmann: I wonder if this should rather be `trafo <= 0`?

nengo_gui/components/netgraph.py
```diff
@@ -580,24 +583,81 @@ def send_pan_and_zoom(self, client):
...
+
+    def create_connection(self, client, conn, parent):
+        """
+        Assembles the a JSON description of the given connection object and
+        sends it to the client.
+
+        client: websocket object to which the assembled JSON description is
+        sent.
+        conn: connection object for which the description should be generated.
+        parent: UID of the parent network the connection belongs to.
```
jgosmann: We're using [NumpyDoc style](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt) docstrings.

nengo_gui/static/components/netgraph_conn.js
```diff
@@ -167,25 +169,40 @@ Nengo.NetGraphConnection.prototype.set_pre = function(pre) {
...
-        this.post.conn_in.splice(index, 1);
     }
     this.post = post;
     if (this.post !== null) {
-        /** add myself to post's input connections list */
-        this.post.conn_in.push(this);
+        let elems = [this.post]
+        if (this.post instanceof Nengo.NetGraphConnection) {
+            elems = [this.post.pre, this.post.post]
+        }
```
jgosmann: The preceding four lines appear twice in this function, maybe extract them into a separate function (e.g., `get_elems`)?

nengo_gui/components/netgraph.py
astoeckel: Hm, I don't think so. I wouldn't consider a connection to be inhibitory if the connection has no effect whatsoever.

jgosmann: `np.all(trafo <= 0)` allows for a connection that has not effect, but it doesn't imply that the connection has no effect. I would consider a sparse inhibitory matrix still inhibitory.

astoeckel: Ok, I see, you are right. I'm going for

```python
np.all(trafo <= 0.0) and not np.all(np.isclose(trafo, 0.0))
```

and add some more test cases.

nengo_gui/components/netgraph.py
astoeckel: I've updated the docsting and sqaushed my changes into the first commit.

nengo_gui/components/netgraph.py
astoeckel: Ok, fixed this, as discussed above.

jgosmann: You can use `np.allclose` instead of `np.all(np.isclose(...))`.

nengo_gui/static/components/netgraph_conn.js
astoeckel: I've extracted a helper function and added some documentation explaining what this code is actually doing. I've squashed my changes into the last commit.

astoeckel:
I've squashed all of Jan's suggestions into the four commits described above -- I wanted to keep these four commits since they are independent logical units.

jgosmann:
Using this code
```python
import nengo
import numpy as np

model = nengo.Network()
with model:
    stim = nengo.Ensemble(50, 1)
    #nengo.Connection(stim, a)
    err = nengo.Node(0.)
    
    c = nengo.Connection(stim, a, learning_rule_type=[nengo.PES(), nengo.PES()])
    nengo.Connection(err, c.learning_rule[0])
    nengo.Connection(nengo.Node(1), c.learning_rule[1])
    nengo.Connection(err, a.neurons, transform=-np.ones((a.n_neurons, 1)))
```
and commenting line 12 makes the modulatory go to the wrong place:
gui1
Uncommenting makes it go back to the correct place, but the second modulatory connection does not reappear:
gui2
Not sure if that is a problem with the code introduced in this PR or if this is related to some existing issues. Live editing code easily breaks the GUI, see #877, [comment](https://github.com/nengo/nengo_gui/issues/763#issuecomment-224713123) in #763, #879). I would have tried to solve those problems by now, but the last time I tried I did not manage to fully understand the core part of the code involved with this issues.

astoeckel:
Thanks for pointing this out, I'll see whether I can fix the problem.

Nonetheless, I'd like to argue that we should merge the code even if small imperfections remain, since I believe the current change-set is as minimal as possible. As you pointed out, both the Python and JS code involved in generating the netgraph view are "rather ad-hoc, messy, and weirdly interconnected" (and not documented at all). So any non-surgical change is likely to break other stuff.

astoeckel:
Noticed and corrected two missing semicolons, and reflecting the fact that I'm no longer *monkeypatching* (which I did originally) the NetgraphItem interface into the NetgraphConnection in a comment.

jgosmann:
Shall we document the problem of connections going to the wrong places in an issue?

astoeckel:
Well, I'm currently looking into the issue, if I can't come up with a simple solution today I'm fine with merging this branch and documenting the issue separately.

astoeckel:
Ok, I fixed the issue I think.

There was some code-duplication in `netgraph.py` regarding the derivation of the post and pre object. I only changed the part related to the creation of new objects, but not the part of the code related to updating objects.

I've simply moved the common code to a function `get_connection_hierarchy` and the problem disappeared. For now I leave that patch in a separate commit, but it should actually be part of "Commit 1" above.

jgosmann:
> I've simply moved the common code to a function get_connection_hierarchy and the problem disappeared. For now I leave that patch in a separate commit, but it should actually be part of "Commit 1" above.

That's actually most helpful for reviewers and it makes it easier to determine what changed since the last time looking at it (and updating of local branches is easier if nothing has been force pushed, especially if someone made some local commits). For small and normal-sized PRs it's usually best to squash and cleanup the history after/review or close to the merge. Might even be done by the person merging in some cases.

You can mark commits as to be combined with a different commit with `--fixup` and `--squash` commits. I usually use `--squash` if I want to combine a commit later, but also add some explenation for this added commit (that might be removed in the final squashing/rebasing).

PR #930, opened 2017-11-04
Add autocompletion to editor.

jgosmann:
A feature that I was hoping for a long time finally implemented: Autocompletion in the editor. Was actually pretty straight-forward to do. This uses the Python jedi module that is also used by IPython (so there is a high chance that it is already installed). If it is not installed, no autocompletions are provided a warning, saying to install jedi for autocompletion, is given.

Currently I enabled the live autocompletion that happens while typing. This might be a potential performance pitfall. We might want to disable it or add a configuration option for it. If it is not enabled, the autocompletion can be triggered with Ctrl+Space.


tcstewar:
Okay, the autocomplete is pretty much awesome.  :)

My plan is to force it on everyone by merging it into master, and then if anyone doesn't like it we can add an option to turn it off (and/or have more of a discussion about it)

PR #932, opened 2017-11-05
Add syntax highlighting for nengo_spa.Actions.

jgosmann:
This adds a first attempt on syntax highlighting for `nengo_spa.Actions` to the Nengo GUI editor (as requested in nengo/nengo_spa#91). The syntax highlighting can probably still improved and there might be cases that are not properly covered.

* The highlighting is detected by the string `spa.Actions(` which should work for `nengo_spa.Actions` too, but in theory `spa` and `Actions` could have completely different names or someone could do `from nengo_spa import Actions`. These cases are not covered (and cannot be covered without executing the Python code which is a bad idea in a syntax highlighter).
* Around the `.` and the `(` one could have newlines, these are not supported (not sure if the JS regexes can be made to match newlines). Other whitespace is allowed.
* The actions use the normal Python highlighter with some rules for new keywords (`ifmax`, `elifmax`) etc added in. There might be some Python highlighting rules that should not apply inside the actions, but I'm not sure ... (too lazy to think through all the cases and remove those rules)
* Every identifier starting with a capital letter will be highlighted as Semantic Pointer even though it might be a variable or class name.
* The builtins `dot`, `reinterpret`, and `translate` are highlighted as such.
* Because the action rules use mostly normal highlighting they don't stand out very much. Not sure if there is a way to improve that. the surrounding `spa.Actions(...)` will be highlighted as keyword, not sure if it should be highlighted and as what (but it might help to visually delimit the actions block).

There is also an issue with loading this syntax highlighting rules. The ACE editor is supposed to be loaded with  `requirejs` which we are not doing (ACE is doing it itself for us, but that leaves some path configuration unconfigured). Thus, the highlighting cannot be dynamically loaded as it is supposed to be done and the `mode-python.js` and `mode-spa.js` need to be manually loaded in the HTML. I suppose that will be fixed in the refactoring and as fixing it likely requires some reorganization of our JavaScript files I have no intention in doing so.

And finally, a screenshot showing off the highlighting:
highlight

### DIFF
--- a/nengo_gui/static/ace.js
+++ b/nengo_gui/static/ace.js
@@ -33,7 +33,7 @@ Nengo.Ace = function (uid, args) {
     code_div.id = 'editor'
     $('#rightpane').append(code_div);
     this.editor = ace.edit('editor')
-    this.editor.getSession().setMode("ace/mode/python");
+    this.editor.getSession().setMode("ace/mode/spa");
     this.editor.gotoLine(1);
     this.marker = null;
 
--- a/nengo_gui/static/mode-spa.js
+++ b/nengo_gui/static/mode-spa.js
@@ -1,0 +1,99 @@
+define('ace/mode/spa', function(require, exports, module) {
+
+var oop = require("ace/lib/oop");
+var PythonMode = require("ace/mode/python").Mode;
+var SpaHighlightRules = require("ace/mode/spa_highlight_rules").SpaHighlightRules;
+
+var Mode = function() {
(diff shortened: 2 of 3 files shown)


jgosmann:
One more point: At some point the syntax highlighting mode should be provided by `nengo_spa` itself and be advertised via an entry point to Nengo GUI. But that will have to wait until we have the infrastructure to the same thing for components and that again won't happen before the frontend refactoring.

jgosmann:
Made obsolete by nengo/nengo_spa#114.

PR #931, opened 2017-11-04
Fix tests

jgosmann:
This makes the tests run and pass again. It includes a minor fix the the dual-stack IPv6 support with automatically selected ports.

There might be some more changes necessary to get things working on Travis-CI again.

PR #933, opened 2017-11-06
Fix tests (local)

jgosmann:
This makes the tests run and pass again. It includes a minor fix for the dual-stack IPv6 support with automatically selected ports and a fix for shutting down requests. Furthermore, this disables the travis build as I am unable to fix it at the moment. See #931 for my attempts.


jgosmann:
(And I suppose it is useful in testing all the other PRs?)

PR #934, opened 2017-11-07
Control which browser is used

tcstewar:
Here's an attempt at #203.  It should let you select a browser when starting nengo, so you can do something like  `nengo --browser firefox`.  It uses whatever browsers are registered with python (see https://docs.python.org/2/library/webbrowser.html for a list).

However, it doesn't seem to work on Windows for anything other than `windows-default`, and I haven't seen a good way to figure out what browsers are available (to give a useful error message).  But, I figured I would post the PR and see if it works for other people.

### DIFF
--- a/nengo_gui/main.py
+++ b/nengo_gui/main.py
@@ -18,6 +18,13 @@ def old_main():
 
 
 def main():
+    try:
+        browser_help = ('browser to use (options: "%s")' % 
+                        '", "'.join(webbrowser._tryorder))
+    except (AttributeError, TypeError):
+        # just in case the undocumented webbrowser._tryorder changes
+        browser_help='browser to use (e.g. chrome, firefox)'
+
     parser = argparse.ArgumentParser()
     parser.add_argument(
         '-p', '--password', dest='password', metavar='PASS',
@@ -46,7 +53,7 @@ def main():
         default='nengo', type=str, help='default backend to use')
     parser.add_argument('--browser', dest='browser', type=str,
         metavar='BROWSER', default=True,
-        help='browser to use (e.g. chrome, firefox)')
+        help=browser_help)
     parser.add_argument('--no-browser', dest='browser', action='store_false')
     parser.add_argument(
         '--auto-shutdown', nargs=1, type=float,
@@ -115,7 +122,17 @@ def main():
             if args.browser is True:
                 wb = webbrowser.get()
             else:
-                wb = webbrowser.get(args.browser)
+                try:
+                    wb = webbrowser.get(args.browser)
+                except webbrowser.Error:
+                    try:
+                        print('Known browsers: \n  %s' % 
+                              '\n  '.join(webbrowser._tryorder))
+                    except (AttributeError, TypeError):
+                        # just in case the undocumented webbrowser._tryorder
+                        #  changes
+                        print('Could not determine the list of known browsers.')
+                    raise
             t = threading.Thread(
                 target=wb.open,
                 args=('%s//%s:%d/?token=%s' % (
--- a/nengo_gui/main.py
+++ b/nengo_gui/main.py
@@ -44,7 +44,9 @@ def main():
     parser.add_argument(
         '-b', '--backend', metavar='BACKEND',
         default='nengo', type=str, help='default backend to use')
-    parser.add_argument('--browser', dest='browser', action='store_true')
+    parser.add_argument('--browser', dest='browser', type=str,
+        metavar='BROWSER', default=True,
+        help='browser to use (e.g. chrome, firefox)')
     parser.add_argument('--no-browser', dest='browser', action='store_false')
     parser.add_argument(
         '--auto-shutdown', nargs=1, type=float,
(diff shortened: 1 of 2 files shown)


nengo_gui/main.py
```diff
@@ -38,7 +38,8 @@ def main():
     parser.add_argument(
         '-b', '--backend', metavar='BACKEND',
         default='nengo', type=str, help='default backend to use')
-    parser.add_argument('--browser', dest='browser', action='store_true')
+    parser.add_argument('--browser', dest='browser', type=str,
+        metavar='BROWSER', default=True, help='browser to use')
```
astoeckel: Include some common options in the help string, "(e.g., firefox, google-chrome)"

tcstewar:
I've added a commit that I'm not sure about -- it displays a list of known browser options.  But I'm not happy with how I'm getting that list since it's an undocumented python thing.

astoeckel:
Displaying the list of known browsers works for me. Note that (at least on my system) chrome is called `google-chrome`:

```
andreas@andreas-pc2:~$ nengo --browser chrome
Known browsers: 
  xdg-open
  gvfs-open
  firefox
  epiphany
  google-chrome
  links
  File "/home/andreas/.local/bin/nengo", line 11, in <module>
    load_entry_point('nengo-gui', 'console_scripts', 'nengo')()
  File "/home/andreas/source/nengo_gui/nengo_gui/main.py", line 110, in main
    wb = webbrowser.get(args.browser)
  File "/usr/lib64/python3.6/webbrowser.py", line 51, in get
    raise Error("could not locate runnable browser")
webbrowser.Error: could not locate runnable browser
andreas@andreas-pc2:~$ nengo --browser google-chrome
Starting nengo server at http://localhost:8080
```

tcstewar:
> Note that (at least on my system) chrome is called google-chrome:

Yup, that's why I wanted the list added.  On OS X, it's `chrome`, and on linux it could be any of `"google-chrome", "chrome", "chromium", "chromium-browser"`.  <sigh>

jgosmann:
Can we print the list of browser with `nengo --help`? That is where I would expect that information. (Maybe by setting the `epilog` of the `ArgumentParser`?)

Also, I think it is fine to use the undocumented `_browsers` variable, but I would put it into a try/except and in the case of an exception print a statement that the list cannot be obtained. That ensures that future Python versions cannot break the `nengo` command by renaming `_browsers`.

Furthermore, it might make sense to sort the list either alphabetically or by `_try_order` to have entries in order of (likely) relevance to the user.

tcstewar:
> Also, I think it is fine to use the undocumented _browsers variable, but I would put it into a try/except and in the case of an exception print a statement that the list cannot be obtained. That ensures that future Python versions cannot break the nengo command by renaming _browsers.
> 
> Furthermore, it might make sense to sort the list either alphabetically or by _tryorder to have entries in order of (likely) relevance to the user.

Excellent ideas.  Those are both now implemented.  :)

tcstewar:
> Can we print the list of browser with nengo --help? That is where I would expect that information. (Maybe by setting the epilog of the ArgumentParser?)

Also a good idea.  I've added the options right into the help text for the --browser option.

nengo_gui/main.py
```diff
@@ -109,8 +109,11 @@ def main():
                 try:
                     wb = webbrowser.get(args.browser)
                 except webbrowser.Error:
-                    print('Known browsers: \n  %s' % 
-                          '\n  '.join(webbrowser._browsers.keys()))
+                    try:
+                        print('Known browsers: \n  %s' % 
+                              '\n  '.join(webbrowser._tryorder))
+                    except:
```
jgosmann: Bare excepts are considered to be bad practice (because they also catch things like keyboard interrupts). Usually the broadest I do is `except Exception`, but here an `except (AttributeError, TypeError)` should suffice. The `AttributeError` is raised if `_tryorder` does not exist, the `TypeError` if it is a type that does not support string joining (e.g. not a list or a list with non-strings).

jgosmann: Do we know that `set(webbrowser._browsers.keys()) == set(webbrowser._tryorder)`?

tcstewar: > but here an except (AttributeError, TypeError) should suffice

Hmm, good point....   those two should be enough...

> Do we know that set(webbrowser._browsers.keys()) == set(webbrowser._tryorder)?

Argh, we don't technically know that.  It's true for all the implementations I can see, but there is a provision in the webbrowser.py code that it is possible to register a webbrowser and not put it on the list (if someone calls `register` with `update_tryorder` set to something other than `1` or `-1`).  However, since that would seem to be something that someone else would have to call manually, I don't see it being possible in our case.   Some future Python version might do it, but I'm not sure how much to worry about that....

jgosmann: Yeah. We could also indicate that the list we print may be incomplete, but it's probably also fine as it is right now.

tcstewar:
Thank you everyone for the fixes....  Merging this in now!

PR #943, opened 2017-12-04
Fix rendering of recurrent inhibitory connections

tcstewar:
The addition of coloured lines for different connections missed the special case of recurrent inhibitory connections.  Here's what it looks like now:

image

This PR changes that to

image

### DIFF
--- a/nengo_gui/static/components/netgraph.css
+++ b/nengo_gui/static/components/netgraph.css
@@ -29,20 +29,20 @@ g.net rect{
     stroke-width: 2px;
 }
 
-.netgraph g.inhibitory > line {
+.netgraph g.inhibitory > .conn {
     stroke: #c00000;
 }
 
-.netgraph g.inhibitory > path {
+.netgraph g.inhibitory > .marker {
     fill: #c00000;
 }
 
-.netgraph g.modulatory > line {
+.netgraph g.modulatory > .conn {
     stroke: #8eb627;
     stroke-dasharray: 1em;
 }
 
-.netgraph g.modulatory > path {
+.netgraph g.modulatory > .marker {
     fill: #8eb627;
 }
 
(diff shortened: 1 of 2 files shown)


jgosmann:
The half-circle symbol in the screenshots posted is rotated in a weird way ...

astoeckel:
I'll look into this. I'm also not sure whether Terry's change is the best way to fix this, since optimally we shouldn't have to include this CSS rule for each connection type.

astoeckel:
I improved Terry’s patch a little by explicitly adding the "marker" attribute to the element maker element in the SVG. This allows to set its fill colour without accidentally affecting all the other SVG paths in the group.

I also fixed the weird rotation of the marker element in recurrent inhibitory connections.

tcstewar:
> I improved Terry’s patch a little by explicitly adding the "marker" attribute to the element maker element in the SVG. This allows to set its fill colour without accidentally affecting all the other SVG paths in the group.

Nice!  Much better than what I was doing...



> I also fixed the weird rotation of the marker element in recurrent inhibitory connections.

Cool!  Thanks for that....  I flipped the direction to be consistent with the normal recurrent arrow, but other than that it looks great to me!

tcstewar:
(when I merge this, I'll collapse it all down to one commit and indicate @astoeckel as the author, as that's the solution we went with...  :)  )

jgosmann:
Note: As we're not signing commits [you can manually set the author on a commit](https://stackoverflow.com/questions/3042437/change-commit-author-at-one-specific-commit) in case squashing sets the wrong author.

tcstewar:
> Note: As we're not signing commits you can manually set the author on a commit in case squashing sets the wrong author.

Sweet!  Thanks!

astoeckel:
I also just realised that the stroke colour of recurrent connections was not set correctly, so I fixed that by explicitly adding the "conn" class to the connection line and colouring that.

tcstewar:
Oh, and @astoeckel , sorry for doing the commit above!  I didn't check and see you still had yourself assigned to it.  I shouldn't have made commits on it without checking that.....

astoeckel:
No problem, sorry for the confusion ;-) I'm not 100% sure regarding the offset in the marker path that you've changed from `10.0` to `5.0`. I consciously set it to `10.0` to have the marker at the apex of the circle, such that the direction indicated by the marker is tangential to the circle.

Should we keep the marker in the current location but rotate it marker a little?

I really don't care that much, just wanted to give some rationale as for why I introduced that offset.

Issue #752, opened 2016-06-06
Crosshair grid for value plots doesn't work in Iceweasel


tcstewar:
Maybe doesn't work in firefox as well

astoeckel:
I can confirm that the crosshair doesn't work in a recent Firefox. Just installed Chromium and for the first time noticed the crosshair.

jgosmann:
I think it was never working in Firefox?

Seanny123:
This still doesn't work, even with the new FireFox.

PR #947, opened 2017-12-18
Spike sounds

celiasmith:
An update to the raster plot so you can listen to the spikes of neurons you hover over.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,6 +22,7 @@ Release History
 0.4.4 (unreleased)
 ==================
 
+- Improvement: Added audible spike sounds
 - Improvement (Experimental): Nodes have access to the GUI keyboard state
 - Improvement: support for nengo-bio Connections
 - Improvement: show multiple connections between objects
--- a/nengo_gui/static/components/raster.js
+++ b/nengo_gui/static/components/raster.js
@@ -23,6 +23,62 @@ Nengo.Raster = function(parent, sim, args) {
     this.axes2d = new Nengo.TimeAxes(this.div, args);
     this.axes2d.scale_y.domain([0, args.n_neurons]);
 
+    // Flag for whether or not update code should be changing the highlight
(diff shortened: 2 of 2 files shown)


tcstewar:
This really is just so cool.  :)

One UI quesion for people: should this be controlled by mouseover?  Or should you have to click on a row to set that one as the spiking one?  That can make it easier to select something and then, say, move the mouse over to a slider and start moving it up and down and listen to the sounds change.

Also, as it currently stands, it always gives clicks based on the spikes that are being generated *right now*, not what's shown on the right-hand side of the spike raster plot..  So if you rewind time, you'll be hearing spikes from the future.   I think I have a way around that, but I still need to poke at that a bit....  :)

celiasmith:
Good point about clicking on it to 'lock' the neuron being listened too... that would be better, I agree.

astoeckel:
Doesn't work in Firefox. Cannot select rows in the raster plot, which works in Chromium. Probably related to #752 .

Also, playing back audio should use the WebAudio API which is meant for this kind of real-time audio playback, whereas the "Audio" element is for the playback of longer audio files. E.g. the browser will only open one playback stream and mix concurrent events. Am going to have a look at this.

astoeckel:
Implemented the usage of WebAudio, now plays clicks with the actual spike frequency (was much slower on my computer before).

tcstewar:
I just rebased this to master and made some small tweaks to the UI (the width of the selection box, and not having clicks outside the spike area suddenly turn into clicks on mouseover)

tcstewar:
Rebased to master again.  :)

tcstewar:
I just had a sudden inspiration to add a menu option to turn this on and off, so that it's not as surprising when your computer starts making loud clicking sounds (and makes it clearer how to turn it off).  That's a small enough tweak that I'm happy to now merge this into master.  :)

PR #952, opened 2018-04-18
Added option to turn off autocompletion

tcstewar:
With very large models, I sometimes find the autocompletion somewhat slow and unresponsive, and it makes typing a bit annoying.  So, here's a quick option to let you turn the autocompletion on and off.


jgosmann:
How about just turning off the live autocompletion, but still allowing it on key press (the shortcut is something like Ctrl+Space I believe)?

tcstewar:
> How about just turning off the live autocompletion, but still allowing it on key press (the shortcut is something like Ctrl+Space I believe)?

Ooo, good idea -- and that's even simpler than this fix.  I've made that change now (and yes, it's Ctrl+Space)

tcstewar:
> Inserted the word ‘live’ into the checkbox label and while I was at it also added the autocompletion hotkey to the hotkey help window.

Nice!  Thank you!

PR #955, opened 2018-04-28
Use post method for autocompletion.

### LINKED ISSUE #954: Autocompletion should send code data via post
It seems (haven't verified this a 100%) that the autocompletion is sending the code via the GET method (including it in the URL), but it should use the POST method (because it can be more data than allowed in the URL).

It could also be a problem restricted to the tunneling in #921.

jgosmann:
Fixes #954.


tcstewar:
Looks good to me!  Thank you!

PR #958, opened 2018-05-01
Build progress bar

jgosmann:
This adds a build progress bar to the status bar (based on PR #927). This is useful as it gives an indication how long it might take to build the model.

Note that this requires that backends follow the API established by the reference simulator, i.e. they need to allow for a `progress_bar` argument to both the constructor and the `run_steps` method. The backend could ignore the argument (as long as it exists) which would not update the build progress bar in the GUI (but nothing should break). For more backwards compatibility we could try to inspect the function signatures, but I am not sure whether that is worth the effort and this might lead to other problems.

As this adds an additional element to the status bar, I also did some renaming. The pure text component is now names “info line” whereas status bar means both that text component and the progress bar.


tcstewar:
> Or is this for older Nengo versions?

Yup, we only added `nengo.utils.compat.escape` after the 2.7.0 release....

tcstewar:
The last fixup https://github.com/nengo/nengo-gui/pull/958/commits/16e629de19fdd6e97c744f0f72e1d6c630aca003 is just that the slider bar was at height=0....  oops.  :)

tbekolay:
FYI commits with messages starting with `squash!` or `fixup!` should be squashed or fixed up into the relevant commits. LMK if you want a series of steps on how to do that

tcstewar:
Argh, I forgot to change those log messages -- due to the history of this PR the things those two commits somehow got squashed but still had the squash! indicator in their message -- even though there's nothing for them to squash into.....   I was just going to edit the log message to fix that....

PR #959, opened 2018-05-14
Fix Python 3 issue in HBB tutorial

tbekolay:
As reported in [this forum post](https://forum.nengo.ai/t/problem-w-tutorial-file-4-spa-sequencerouted-cleanupall-py/543/2).

### DIFF
--- a/nengo_gui/examples/hbb_tutorials/chapter7/4-spa_sequencerouted_cleanupAll.py
+++ b/nengo_gui/examples/hbb_tutorials/chapter7/4-spa_sequencerouted_cleanupAll.py
@@ -69,8 +69,9 @@ with model:
         pd_new.append([vocab[pd[element]].v.tolist()]) 
         
     #cleanup = nengo.Ensemble(300, dimensions=vsize)
-    model.cleanup = spa.State(neurons_per_dimension=300/vsize, dimensions=vsize)
-    
+    model.cleanup = spa.State(neurons_per_dimension=int(300//vsize),
+                              dimensions=vsize)
+
     #Function that provides the model with an initial input semantic pointer.
     def start(t):
         if t < 0.4:


nengo_gui/examples/hbb_tutorials/chapter7/4-spa_sequencerouted_cleanupAll.py
```diff
@@ -69,8 +69,9 @@
         pd_new.append([vocab[pd[element]].v.tolist()]) 
         
     #cleanup = nengo.Ensemble(300, dimensions=vsize)
-    model.cleanup = spa.State(neurons_per_dimension=300/vsize, dimensions=vsize)
-    
+    model.cleanup = spa.State(neurons_per_dimension=int(300//vsize),
```
jgosmann: I haven't looked at this tutorial in detail, but it seems weird to me to divide the neuron number by the vocabulary size. Why isn't this a fixed number of neurons per dimension? Also note the commented line above ... (which should be removed?) It actually seems to be that an `EnsembleArray` would be more appropriate here then either `spa.State` or an `Ensemble`.

Apart from that the casting to `int` is technically not necessary (`vsize` will be an integer), but doesn't really hurt.

tbekolay: I just wanted to fix the bug here, not do a style pass on all the tutorials. Let's try to keep the scope of the PR to that purpose so we can get the bug fixed rather than bog it down in other improvements (which should definitely be done eventually).

> Apart from that the casting to int is technically not necessary (vsize will be an integer), but doesn't really hurt.

We got a bug report because of it, so obviously it is necessary. In Python 3, dividing two integers gives a float.

```
In [1]: 300 / 5
Out[1]: 60.0
```

jgosmann: Integer division with integers gives you an integer:
```ipython
In [1]: 300 // 5
Out[1]: 60
```

> I just wanted to fix the bug here, not do a style pass on all the tutorials.

I'm not requesting a style pass on all the tutorials, but to me the change seemed to address the symptom without fixing the underlying cause (so to say). It might be worth to take a little bit of extra time to just slightly improve these two lines of code (e.g., delete the commented line and change it to a constant number of neurons or an ensemble array) if they are being touched already. That being said, I'd rather have the current fix merged than no fix at all.

PR #960, opened 2018-05-23
Ipythonviz fixes

### LINKED ISSUE #846: jupyter not finding 'model'?
i'm getting this error: 

untitled

with the most recent versions of nengo and nengo_gui, any thoughts?

jgosmann:
Fixes the access to the authentication token that has changed and fixes #846 by not using the editor saved in the config (and no longer saving the editor to the config at all). Instead the editor from the page settings will always be used.

Relying on the config can cause problems. For instance, when the config was writting by the standard GUI with ACE editor, it will load the ACE editor also when reusing that config within IPythonViz. But the editor needs to be deactivated in IPythonViz to have it work properly (the editor sends empty code that gets compiled and is then used instead of the actual model passed to IPythonViz).

### DIFF
--- a/nengo_gui/config.py
+++ b/nengo_gui/config.py
@@ -58,6 +58,9 @@ class Config(nengo.Config):
                     lines.append('_viz_config[%s].has_layout=%s'
                                  % (uid, self[obj].has_layout))
 
+            elif isinstance(obj, nengo_gui.components.editor.Editor):
+                pass  # Do not persist editor
+
             elif isinstance(obj, nengo_gui.components.component.Component):
                 lines.append('%s = %s' % (uid, obj.code_python(uids)))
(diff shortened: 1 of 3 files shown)


nengo_gui/config.py
```diff
@@ -58,6 +58,9 @@ def dumps(self, uids):
                     lines.append('_viz_config[%s].has_layout=%s'
                                  % (uid, self[obj].has_layout))
 
+            elif isinstance(obj, nengo_gui.components.component.Editor):
```
tcstewar: Should this be `nengo_gui.components.editor.Editor`?

jgosmann: Yes, good catch.

tcstewar:
Thank you!  Great to have IPythonViz working again!

PR #963, opened 2018-05-30
Add dt option

tcstewar:
For some of the backends (especially nengo_brainstorm), it is quite useful to be able to adjust dt.  This adds a quick command-line option to do this.


tcstewar:
Oops, I based this on the wrong branch -- lemme redo this.

PR #964, opened 2018-05-30
Added dt parameter

tcstewar:
For some of the backends (especially nengo_brainstorm), it is quite useful to be able to adjust dt. This adds a quick command-line option to do this.


arvoelke:
Is there an easy way to expose this value to the Nengo model (script in the GUI)?

tcstewar:
> Is there an easy way to expose this value to the Nengo model (script in the GUI)? I'm thinking something like being able to use the variable nengo_gui.settings.dt (or similar) within the model.

Currently we don't have anything like that.  I could definitely see working towards that sort of direction, along with the `nengo_gui.hooks` discussion and also I think even how the keyboard state is kept in #942

arvoelke:
In the mean time, here's a *really* hacky way to access the page settings from within the model, just in case anyone is dying for a temporary solution:
```
import inspect
dt = inspect.getargvalues(inspect.stack()[1][0]).locals['self'].settings.dt
```
(Tested with Python 2.7 and the master branch of `nengo-gui`.)

You can also replace `dt` with `backend` to get the current backend. 

Another solution could be to extend the `code_locals` dictionary passed into `exec` in some way.

Issue #975, opened 2018-06-08
Potential race condition in Jedi autocompletion?


jgosmann:
It seems that on some machines the autocompletion will break by typing too fast. There might be a race condition that another autocompletion request gets processed before the previous one finished which would give to concurrent jedi processes that might access the same cache.

tcstewar:
This might possibly be related to #962 ?

jgosmann:
Yes, I believe that is the same class of errors.

PR #977, opened 2018-06-09
Acquire lock when using Jedi

jgosmann:
This should ensure thread safety, but might not give the best
performance. It might be possible to skip processing of some
requests if too many arrive in a short interval instead of processing
all of them with a delay once the lock becomes available for each
request.

Addresses #975.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,6 +22,7 @@ Release History
 0.4.3 (unreleased)
 ==================
 
+- Bugfix: thread-safety for jedi autocompletion
 - Bugfix: Handle authentication when connecting to multiple servers
 - Bugfix: Fail gracefully when binding server
 
--- a/nengo_gui/completion.py
+++ b/nengo_gui/completion.py
@@ -7,7 +7,7 @@ except ImportError:
         "Install the jedi module to get autocompletion in Nengo GUI.")
 
     class Script(object):
-        def __init__(*args, **kwargs):
+        def __init__(self, *args, **kwargs):
             pass
 
         def completions(self):
--- a/nengo_gui/completion.py
+++ b/nengo_gui/completion.py
@@ -1,3 +1,4 @@
+import threading
 import warnings
 
 try:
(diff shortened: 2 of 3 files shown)


jgosmann:
I've got the user who experienced the problem to test this branch. So far it seems to be working.

Seanny123:
Still getting the same error:
```
  File "c:\users\ctnuser\git\nengo_gui\nengo_gui\server.py", line 400, in do_GET
    self.http_GET()
  File "c:\users\ctnuser\git\nengo_gui\nengo_gui\server.py", line 416, in http_GET
    response = getattr(self, command)()
  File "c:\users\ctnuser\git\nengo_gui\nengo_gui\guibackend.py", line 83, in auth_checked
    return fn(inst)
  File "c:\users\ctnuser\git\nengo_gui\nengo_gui\guibackend.py", line 198, in complete
    self.db['filename'])
  File "c:\users\ctnuser\git\nengo_gui\nengo_gui\completion.py", line 23, in get_completions
    return script.completions()
  File "c:\python27\lib\site-packages\jedi\api\__init__.py", line 172, in completions
    self._pos, self.call_signatures
  File "c:\python27\lib\site-packages\jedi\api\completion.py", line 90, in __init__
    self._like_name = helpers.get_on_completion_name(self._module_node, code_lines, position)
  File "c:\python27\lib\site-packages\jedi\api\helpers.py", line 28, in get_on_completion_name
    leaf = module_node.get_leaf_for_position(position)
  File "c:\python27\lib\site-packages\parso\tree.py", line 309, in get_leaf_for_position
    raise ValueError('Please provide a position that exists within this node.')
ValueError: Please provide a position that exists within this node.
```

tcstewar:
Even though this doesn't seem to completely solve the problem, given that jedi isn't thread-safe (https://jedi.readthedocs.io/en/latest/docs/development.html#module-jedi.cache), we should definitely merge this in anyway.  Thanks for doing this!

PR #981, opened 2018-06-12
Fix shared session ID across ports.

### LINKED ISSUE #979: Session ID shared across tabs connected to different GUI servers

jgosmann:
Fixes #979.

Cookies do not differentiate between ports and are sent to all servers
on the same domain independent of the port. Thus, we append the port
to the cookie name to tell them apart.


tcstewar:
Thank you!  Nice and clean fix....  :)

Issue #1000, opened 2018-12-06
Give name from assignment higher precedence than label


jgosmann:
Currently, the label for a Nengo object will have higher precedence than the name determined from assignment. For example:
```python
with nengo.Network(label="this name is used") as not_this_one:
    also_not_this_one = nengo.Ensemble(10, 1, label="this name is used again")
```
However, many Networks in nengo_spa have a default label which is not very specific. In those, cases it would be beneficial to use the name from the assignment:
```python
use_this_name = spa.State(64)  # Actual name will be "State"
```

I currently cannot think of any situation where the proposed behaviour would be undesired. If a name from assignment can be found, it usually will be more specific (because it should be unique in the scope, or if not the name will be used for only one of the objects anyways). If no such name is found, the name can be set via the label as before.

tcstewar:
> I currently cannot think of any situation where the proposed behaviour would be undesired.

The one that comes to mind for me is that I've often wanted to set `label` in order to directly control what the GUI displays, for purposes of making demos or screenshots for papers.  I'd like to maintain that functionality, if possible...

But I definitely agree that the default labels in `nengo_spa` aren't very specific.  One other possibility would be to remove the labels from `nengo_spa`, if they aren't specific enough to be useful....

jgosmann:
That would lead to even less specific labels in other situations as many objects in expressions are not assigned to a discoverable variable.

tcstewar:
> That would lead to even less specific labels in other situations as many objects in expressions are not assigned to a discoverable variable.

Hmm..  so why not only set labels on things that aren't assigned to discoverable variables?

jgosmann:
Because there is no way (except maybe some fragile introspection) to know what gets assigned to a variable when the label is being set.

On December 7, 2018 1:31:37 PM GMT+01:00, tcstewar <notifications@github.com> wrote:
>> That would lead to even less specific labels in other situations as
>many objects in expressions are not assigned to a discoverable
>variable.
>
>Hmm..  so why not only set labels on things that aren't assigned to
>discoverable variables?  
>
>
>
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/nengo/nengo-gui/issues/1000#issuecomment-445219042

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

tcstewar:
[...]

PR #1002, opened 2019-01-12
Use class name of a network if it is more informative

jgosmann:
This implements my proposal from #1000. I think this is an improvement and would enable Nengo SPA to **not** set the label for networks and give better names to SPA objects when no explicit label is given.


jgosmann:
ping @tcstewar @krisfrenette

jgosmann:
Ping @tcstewar @krisfrenette

PR #1004, opened 2019-05-14
Fix for connection line markers in recent nengo versions

tcstewar:
When running with the master branch of Nengo, the GUI crashes when it tries to display connections that got to `post.neurons`.  For example:

```python
import nengo

model = nengo.Network()
with model:
    a = nengo.Ensemble(n_neurons=100, dimensions=1)
    b = nengo.Ensemble(n_neurons=100, dimensions=1)
    
    nengo.Connection(a.neurons, b.neurons)
```

This is because the gui is checking the Connection.transform object, which it expects to be either a scalar or a matrix, rather than the new Transform object.

Here is a quick fix for the common use case.  In general, we might want to also check for other types of Transforms and display them in different ways.  Note that I made this fix be backward compatible with older versions of nengo, but not in a particularly good way.

### DIFF
--- a/nengo_gui/components/netgraph.py
+++ b/nengo_gui/components/netgraph.py
@@ -610,9 +610,10 @@ class NetGraph(Component):
             trafo = conn.transform
             if hasattr(trafo, 'init'):   # check for DenseTransform
                 trafo = trafo.init
-            if trafo.size > 0 and (np.all(trafo <= 0.0) and
-                    not np.all(np.isclose(trafo, 0.0))):
-                return "inhibitory"
+            if hasattr(trafo, 'size'):
+                if trafo.size > 0 and (np.all(trafo <= 0.0) and
+                        not np.all(np.isclose(trafo, 0.0))):
+                    return "inhibitory"
         return "normal"
 
     def get_connection_hierarchy(self, conn, default_labels=None):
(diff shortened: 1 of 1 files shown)


arvoelke:
Note this interacts with https://github.com/nengo/nengo-gui/pull/1001, since after that PR the GUI will not support the master branch of Nengo (`3.0.0-dev`). This will be most helpful for when `nengo-gui` supports `nengo>=3.0`, although unsure where that stands on the road map, and whether that will need to support `nengo<3.0` in a backwards-compatible way (which might simplify your fix).

astoeckel:
Please use #1010 instead of this PR (I don't have the rights to commit nengo/nengo-gui, so I had to create a new PR). My PR is a little cleaner (uses proper detection for Nengo 3.0 and uses the correct function to access the matrix).

@arvolke Given transient state of the current GUI it would IMHO be great if we could have fewer philosophical discussions and just fix some annoyances before the summer school.

arvoelke:
> @arvolke Given transient state of the current GUI it would IMHO be great if we could have fewer philosophical discussions and just fix some annoyances before the summer school.

I'd love that but there isn't much else I can do. I hope these fixes get in.

astoeckel:
This is redundant since #1012 has been merged; I'm closing this. Thanks!

PR #1009, opened 2019-06-02
Reset pan and zoom on autolayout

astoeckel:
Previously, when applying an autolayout, the object coordinates were
rescaled such that they fit the current viewport taking pan and zoom
into account.

Unfortunately, this causes (a) numerical instability due to exponentially
exploding/imploding coordinates, (b) issues with scale-dependent objects
such as the multi-connection arcs I'm currently implementing.

This PR resets pan and zoom instead of rescaling the coordinates
in a new "feedforward_layout_done" message handler. Since this would
also change the coordinates of the non-netgraph components, the
transformation inverse to the pan/zoom reset is applied to these
components.

### DIFF
--- a/nengo_gui/static/components/netgraph.js
+++ b/nengo_gui/static/components/netgraph.js
@@ -442,6 +442,20 @@ Nengo.NetGraph.prototype.on_message = function(event) {
                 break;
             }
         }
+    } else if (data.type === 'feedforward_layout_done') {
+        // Preserve the location of the non-netgraph components by applying
+        // the inverse transformation of the below pan/zoom reset
+        const x = viewport.x, y = viewport.y, scale = viewport.scale;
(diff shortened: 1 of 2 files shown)


astoeckel:
Just to add some more context: one quite quickly ends up with config files like these:
```
_viz_config[_viz_0].x = 2.787297689367216e+26
_viz_config[_viz_0].y = 2.0661282308455743e+26
_viz_config[_viz_0].width = 1.0907455255117949e+26
_viz_config[_viz_0].height = 1.466644177044316e+26
```
At which point the `interact` library does weird stuff, e.g., resizing objects no longer works properly. This PR prevents this from happening.

tcstewar:
Great catch!  Thank you for finding this!  It's been merged.  :)

PR #1010, opened 2019-06-02
Fix `connection_kind` in Nengo 3.0

astoeckel:
Connection kind detection broke with Nengo 3.0 and the introduction of
multiple transformation kinds. `test_netgraph.py` exits with failures.

This patch fixes this issue in a backwards-compatible manner by checking
for the presence of Nengo 3.0 and accessing the transfomation in the
correct way.

You can reproduce this issue by running
```bash
pytest --noconftest nengo_gui/tests/test_netgraph.py
```

**Note:** this PR is similar to #1004 but is (I think) a little cleaner.

### DIFF
--- a/nengo_gui/components/netgraph.py
+++ b/nengo_gui/components/netgraph.py
@@ -608,9 +608,13 @@ class NetGraph(Component):
             return "modulatory"
         if isinstance(conn.post_obj, nengo.ensemble.Neurons):
             trafo = conn.transform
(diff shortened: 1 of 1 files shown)


astoeckel:
As noted in the corresponding PR, #1012 also includes this PR. However, #1010 is far more extensive, hence I'm leaving this open.

tcstewar:
This has now been merged.  Thank you!

PR #1015, opened 2019-06-07
Fix bounding box computation for connections

### LINKED ISSUE #1014: Modulatory connections are drawn to (0,0)
All modulatory connections seem to connect to the top-left corner.  For example:

```python
import nengo
model = nengo.Network()
with model:
    pre = nengo.Ensemble(n_neurons=100, dimensions=1)
    post = nengo.Ensemble(n_neurons=100, dimensions=1)
    learn_conn = nengo.Connection(pre, post,
                                  learning_rule_type=nengo.PES())
    error = nengo.Ensemble(n_neurons=100, dimensions=1)
    nengo.Connection(error, learn_conn.learning_rule)
```

image

astoeckel:
Since connections no longer use SVG the 'line' object but
a generic 'path' instead, the attributes 'x1', 'y1', 'x2',
'y2' no longer exist.

Instead of trying to parse the SVG attributes, just ask the
browser for the correct bounding box.

Fixes #1014

### DIFF
--- a/nengo_gui/static/components/netgraph_conn.js
+++ b/nengo_gui/static/components/netgraph_conn.js
@@ -496,16 +496,12 @@ Nengo.NetGraphConnection.prototype.intersect_length = function(theta, alpha, wid
  */
 
 Nengo.NetGraphConnection.prototype.get_bounding_box = function () {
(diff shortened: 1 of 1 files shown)


tcstewar:
Thank you for this quick fix!

PR #1019, opened 2019-09-26
Fix broken sliders given recent nengo.Process API change

tcstewar:
The current `master` version of nengo causes nengo-gui to break if you run a model and have a slider.

Here's the commit that broke things:

https://github.com/nengo/nengo/commit/e4ac75f8cc4b140b892e765226049e393b5e07bd

(it changes the Process API, and nengo-gui uses a Process to implement the sliders)

This fix is backwards-compatible and continues to work with older nengo versions.

### DIFF
--- a/nengo_gui/components/slider.py
+++ b/nengo_gui/components/slider.py
@@ -21,13 +21,20 @@ class OverriddenOutput(Process):
         self.to_client = to_client
         self.from_client = from_client
 
-    def make_step(self, shape_in, shape_out, dt, rng):
+    def make_step(self, shape_in, shape_out, dt, rng, state=None):
         size_out = shape_out[0] if is_iterable(shape_out) else shape_out
 
         if self.base_output is None:
             f = self.passthrough
         elif isinstance(self.base_output, Process):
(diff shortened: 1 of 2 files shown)


tcstewar:
@tbekolay if you could quickly double-check that I'm interpretting the new API correctly then I can quickly merge this  :)

nengo_gui/components/slider.py
```diff
@@ -21,7 +21,7 @@ def __init__(self, base_output, to_client=None, from_client=None):
         self.to_client = to_client
         self.from_client = from_client
 
-    def make_step(self, shape_in, shape_out, dt, rng):
+    def make_step(self, shape_in, shape_out, dt, rng, state=None):
```
tbekolay: Yep, this looks good. I think you also need to change line 30 below to pass in `state`; you could do it in a `try/except` as a quick fix.

tcstewar: Ah, good point...  Thank you!  I've added that via the try/except option (along with calling `make_state` to create the state in the first place)

PR #1021, opened 2019-10-06
Fix jupyter integration

jgosmann:
This fixes two problems:

* Newer Jupyter versions use Tornado 6 and the old code wasn't compatible to that Tornado version.
* Newer Chrome versions (and potentially other browser) check the content type of HTTP responses and the server extension would respond with two (potentially conflicting) Content-Type headers which might lead to resources not being loaded properly.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -23,6 +23,7 @@ Release History
 0.4.5 (unreleased)
 ==================
 
+- Bugfix: Fix for running with Tornado 6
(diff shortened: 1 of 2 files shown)


jgosmann:
@tcstewar ?

tcstewar:
Nice fixes here, @jgosmann !  Thank you for doing this!    This seems to work fine on the various computers I have around.

PR #1024, opened 2020-02-12
Fix bug opening browser from script

tbekolay:
In https://forum.nengo.ai/t/trouble-with-nengo-gui/1071 a user found a bug when using the `GUI` class. This fixes it.


tcstewar:
Thank you!  Good catch....

PR #1025, opened 2020-02-26
Update examples for Nengo 3.0

drasmuss:
Some of the examples didn't work with Nengo 3.0 (as noticed in https://github.com/nengo/nengo-gui/issues/1023), went through all of the examples and fixed any errors that popped up.


tcstewar:
That looks great to me.  This does mean those examples don't work with nengo<2.6, but I think that's okay.

PR #1029, opened 2020-07-21
Removed special characters

p3jawors:
- example would crash the gui due to some special characters in the
comment
- removed whitespace
- bug found through forum post https://forum.nengo.ai/t/bug-found-in-tutorial-file-that-crashes-gui/1257


arvoelke:
FWIW, I ran `LC_ALL=C grep --color='auto' -P -n "[\x80-\xFF]" . -R` to check for non-ascii characters elsewhere in the repo and only found this one occurrence (aside from some JS and vendor stuff outside our control).

PR #1030, opened 2020-08-24
Fix for the spike raster plot

studywolf:
The Nengo commit
`8cf5626d Combine "rates"/"spikes" state into "output"`
broke the current spike raster plot because it's looking for 'spikes'
attached to Ensembles for plotting. This commit fixes it so spike raster
plots work again by searching for 'output' instead.

Also I got rid of an end of line space, smh.


tcstewar:
Hmm, the intent with that if statement was to only show spike rasters for spiking neurons.  I think all neurons have an `output` property, so I think this change would mean we'd try to show spike rasters for non-spiking neurons, and that is going to be visually very confusing. 

Is there another way to determine if a neuron is spiking now?

drasmuss:
Every neuron has a `.spiking=True/False` attribute now https://github.com/nengo/nengo/blob/master/nengo/neurons.py#L67

tcstewar:
Nice!  Looks like a nicer fix that what I tried in 

https://github.com/nengo/nengo-gui/commit/304d92978709a62a4df69410b9c6ddd89ca83b35

If this is meant to get things working with nengo 3, though, then I'd also recommend sneaking in the other fix in that branch: 

https://github.com/nengo/nengo-gui/commit/95e1902dc9cf460454781d3f651c4284bebcd202

tbekolay:
Thanks Terry, I cherry-picked in 95e1902dc9cf460454781d3f651c4284bebcd202. Merging and doing a release :+1:

PR #1037, opened 2022-01-06
Fix collections for Python 3.8 and higher

Tobias-Fischer:
See https://stackoverflow.com/questions/53978542/how-to-use-collections-abc-from-both-python-3-8-and-python-2-7/53978543#53978543 - collections.Iterable is not available anymore for Python 3.8 and higher.

### DIFF
--- a/nengo_gui/compat.py
+++ b/nengo_gui/compat.py
@@ -5,6 +5,11 @@ import sys
 
 import numpy as np
 
+try:
+    collectionsAbc = collections.abc
+except AttributeError:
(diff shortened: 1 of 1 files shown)


jgosmann:
Is Nengo GUI still targeting Python versions <3.3? Otherwise, it should suffice to always import from `collections.abc`, keeping the code a bit simpler.

astoeckel:
I agree with @jgosmann. I've opened an alternative PR at #1040. Thank you for opening the PR nonetheless!

